PR #11, opened 2019-03-21
Added jest tests for tombstone and header.

katiemcgoff:
*Issue #:* bcgov/entity#231

*Description of changes:*
Added jest tests for tombstone and header. Also made typescript not strict, ie: looser validation on possible null values.


katiemcgoff:
> This might be because it was my first time setting up my environment for this project, but when I run the e2e tests one of them fails. I know John is in the middle of tuning those up so maybe it doesn't matter and if so everything looks good to me and I'll merge

@kialj876 Yep the e2e tests aren't up to date for the coops stuff. I'm unsure of whether we're doing in-code e2e tests. The hookups should work, ie: they should run, but they will fail because they're looking for some number of page elements - based on the template project (Hello World type thing) rather than the coops layout.

PR #52, opened 2019-05-01
473 ar ui development

kialj876:
*Issue #:bcgov/entity#473*

*Description of changes:*
- AR + entity info plumbing finished (payment integration + login from other team not covered)
        - AR + entity call mocks currently
        - entity values need to be looked at (hardcoded status + businessNum because they werent in the 
            mock)
        - payment calls a mock but no error checking except for the call succeeding
- jest tests added for ar + entity (axios reconfigured for stubs + better use going forward)
- vuelidate working and implemented around agm date


kialj876:
It checks the date format and that it's a valid date - I'll put in a couple comments to make it more clear

PR #80, opened 2019-05-22
560 multiple filing and files

thorwolpert:
*Issue #:* bcgov/entity#560

*Description of changes:*
Split JSONSchema into multiple files.
Added flask extension to support caching.
Altered filings to remove other changes from AR and allow for multiple filings in a submission.


schemas/src/registry_schemas/utils.py
```diff
@@ -43,32 +32,78 @@ def get_schema(filename: str) -> dict:
 
 def _load_json_schema(filename: str):
```
katiemcgoff: @thorwolpert Looks like we've lost these helper functions to convert python dates to the expected json-schema date and date-time formats . They are used in colin api, and I thought they'd be useful in legal api - which is why I put them here and not just in colin. If not, @kialj876 will have to add them to colin api utils in the work he's currently doing.

katiemcgoff: Diff is unclear here - it's the convert_to_json_date and convert_to_json_datetime functions

thorwolpert: Heavy merge conflicts for some reason, and no tests :) so I dropped out stuff that wasn't checked.

Any reason to not use isoformat and fromisoformat ? tis a bit easier and faster.

thorwolpert: eg. 
datetime.fromisoformat(cache_ts)
epoch_date.isoformat()

katiemcgoff: Ha! Nope no reason other than I didn't know/remember about those functions :) @kialj876 when dates start breaking here's the solution.

PR #134, opened 2019-07-08
UI and store refactoring

leksmall:
*Issue #:*

*Description of changes:* 
Organized the store
Cleaned up the Registered Address component and integrated it with the mock api
Cleaned up the App, EntityInfo, AnnualReport components
Moved setting the corpNum to create lifecycle hook to avoid watch on corpNum
Moved the loading of data to the mount method of the respective child components. 
Fixed a few tests


katiemcgoff:
Approving even though not all tests have been updated - there will be overlap with the next changes so holding off on doing that work.

PR #148, opened 2019-07-11
Changes to post filings to legal api

leksmall:
*Issue #:* /bcgov/entity#849

*Description of changes:*
Changes to post data to legal api filings end point on click of File and Pay button


coops-ui/src/views/AnnualReport.vue
```diff
@@ -161,19 +162,78 @@ export default {
...
+          email: 'no_one@never.get'
+        }
+      }
+
+      if (this.isDataChanged('OTCDR')) {
+        changeOfDirectors = {
+          changeOfDirectors: {
+            certifiedBy: 'Full Name',
+            email: 'no_one@never.get',
+            directors: this.$refs.directors.getAllDirectors()
```
katiemcgoff: $refs.directors gets a js error - there is no ref called "directors". I understand that you are referring to the Directors component above, and I tried out a fix with ref="directors" and it also doesn't work, probably because it's a conflict with the name of the component. If you add ref="directorsList" and use $refs.directorsList here, that will work.

PR #154, opened 2019-07-15
updated unit tests

severinbeauvais:
*Issue #:* https://github.com/bcgov/entity/issues/779

*Description of changes:*

- added App unit tests
- added StdHeader unit tests
- added Dashboard unit tests
- added TodoList unit tests
- added FilingHistoryList unit tests
- added AddressListSm unit tests
- added DirectorListSm unit tests
- added AGMDate unit tests
- added dashboard sub-component validity flags
- updated misc unit test stub data
- fixed misc unit tests to properly use nextTick(), etc
- director and addresses components now disable only when AGM date is invalid
- deleted obsolete FilingHistoryTable component
- deleted obsolete HelloWorld component
- updated router to use VueRouter correctly
- fixed lint errors


WalterMoar:
Good to see the tests being written in the style Vue Test Utils! It's too bad we have such a huge number of tests that need a substantial rewrite to do things the "Vue way". I think at some point we also need to come up with a strategy for how blackboxy we want the tests to be, but those are hard to do when the components are so tightly coupled.

coops-ui/package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "vue-cli-service lint",
     "lint:nofix": "vue-cli-service lint --no-fix",
     "test:e2e": "vue-cli-service test:e2e",
-    "test:unit": "vue-cli-service test:unit",
+    "test:unit": "vue-cli-service test:unit --testPathPattern=*",
```
katiemcgoff: what's the purpose of the new flag?

coops-ui/src/router.ts
```diff
@@ -1,10 +1,10 @@
 import Vue from 'vue'
-import Router from 'vue-router'
+import VueRouter from 'vue-router'
```
katiemcgoff: What is the benefit/difference between Router and VueRouter?

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -168,110 +137,89 @@ describe('Directors.vue', () => {
...
-      expect(
-        directorListUI[0].querySelector('.cease-btn').attributes.getNamedItem('disabled').value
-      ).toEqual('disabled')
-      expect(
-        childvm.$el.querySelector('.new-director-btn').attributes.getNamedItem('disabled').value
-      ).toEqual('disabled')
+      expect(directorListUI[0].querySelector('.cease-btn').disabled).toBe(true)
+      expect(vm.$el.querySelector('.new-director-btn').disabled).toBe(true)
+
+      done()
```
katiemcgoff: I see that done() was removed from afterEach() but added to the end of most of these tests. Do you know if that is necessary? If we can keep it in afterEach() that would be better for code re-use, rather than repetition.

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -168,110 +137,89 @@ describe('Directors.vue', () => {
...
-  it('buttons/actions are enabled when "No AGM" option set', () => {
-    var directorListUI = childvm.$el.querySelectorAll('.director-list .container')
-
-    // set "No AGM" in AGMDate component
-    childvm.$store.state.noAGM = true
-
-    setTimeout(() => {
-      // confirm that flag is set correctly
-      expect(childvm.agmEntered).toEqual(true)
+      const directorListUI = vm.$el.querySelectorAll('.director-list .container')
```
katiemcgoff: Is there a reason this was switched from var to const? If it serves the same purpose I'm resistant to making the change, as well as some of the other changes that may not functionally change anything (rewrite of click() function for example) because of the overlap with other team members.

WalterMoar: In TypeScript we should be using `const` where possible, and `let` if it's a basic type that can change its value (that is, not for objects unless you need to change the reference). The JavaScript `var` and its weird scoping should be avoided, unless the weird scoping is really needed - and I think that's a code smell.

coops-ui/package.json
severinbeauvais: It allows a developer to run the tests in a single file.
I kept having to look up the parameter name so I just left it in package.json with the '*' pattern... It's now easy to replace that with, for example, "AnnualReport", to run those tests only.

coops-ui/src/router.ts
severinbeauvais: VueRouter is the current way to import it.
I think "Router" was the old way, but it resolves to VueRouter anyway.

coops-ui/tests/unit/Directors.spec.ts
severinbeauvais: It doesn't work in afterEach() because it's not called in the context of the it() test itself.
As you can see in the new code, 'done' is a callback function to tell the it() when it's complete (which is why it's called the bottom of the test).
But if you call done() in afterEach(), you are only telling afterEach() when it's done.

coops-ui/tests/unit/Directors.spec.ts
severinbeauvais: It's stronger typing (esp since we're using TS, esp since a TS linter will complain about using var when you should be using const or let instead).

severinbeauvais:
@katiemcgoff Please see my responses above and come poke me if you want to discuss further. Thanks!

severinbeauvais:
@katiemcgoff I have fixed the merge conflicts.

PR #159, opened 2019-07-17
Integration with sbc common

leksmall:
*Issue #:* /bcgov/entity#886

*Description of changes:*
Integrated sbc common components for sign out functionality + common header and footer
Added more items to the session storage for the common components
Passing pay url as props to the sbc fee component
Changes to construct redirect url irrespective of whether '/' is present in the config


leksmall:
The props name for SbcFeeSummary component is payURL
The props name for SbcLayout is authURL

PR #165, opened 2019-07-18
Updated office address component to use BaseAddress

WalterMoar:
*Issue #:* /bcgov/entity/851

*Description of changes:*

Overhaul of the RegisteredOfficeAddress component.
 - conversion to a TypeScript component class
 - now uses BaseAddress components for the delivery and mailing addresses
 - removed use of the vuex store and replaced with properties and events
 - renamed test file to match component. it's weak but coverage is good? Should be extended once the networking code is moved out of the component and the modified eventing with BaseAddress starts working.

Does include some workarounds to deal with the API call being done in the component after mounting. When BaseAddress is fixed and the API call is moved outside the component, it will be much simpler.


coops-ui/package.json
```diff
@@ -13,9 +13,10 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.0.0-rc.1",
+    "@types/vuelidate": "^0.7.5",
```
severinbeauvais: Should Vuelidate version be updated also?
I now the ^ means minor versions are updated, but visually matching the versions would look consistent.

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
```diff
@@ -5,631 +5,372 @@
         <div class="meta-container">
           <label>Delivery Address</label>
           <div class="meta-container__inner">
-            <!-- START: Static Details (Delivery Address) -->
+            <delivery-address
```
severinbeauvais: I noticed that Province is now a drop-down list and I cannot set any value that's not in it. This means I cannot enter a non-Canadian address. Is this OK?

coops-ui/package.json
WalterMoar: Oh, good catch! I'll drop the types/vuelidate to 0.7.4, since that's what version of vuelidate I'm actually running (there is no vuelidate 0.7.5).

coops-ui/src/views/AnnualReport.vue
```diff
@@ -69,7 +69,13 @@
                 <p>Verify or change your Registered Office Addresses.</p>
               </header>
               <v-card flat id="AR-step-2-container">
-                <RegisteredOfficeAddress ref="registeredAddress"/>
+                <RegisteredOfficeAddress
+                  ref="registeredAddress"
+                  :changeButtonDisabled="!agmDateValid"
+                  :legalEntityNumber="corpNum"
+                  @modified="setRegOffAddrChange($event)"
```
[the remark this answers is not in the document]
severinbeauvais: Changed in later commit -- looks good to me!

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
severinbeauvais: This is being discussed separately. This PR can proceed as-is.

PR #167, opened 2019-07-22
replaced deprecated @babel/polyfill package

severinbeauvais:
*Issue #:* no ticket for this

*Description of changes:*

- replaced deprecated @babel/polyfill package (see https://babeljs.io/docs/en/babel-polyfill)
- also bumped vue and vue-template-compiler versions to latest
- npm run build passes
- npm run test:unit passes

### DIFF
--- a/coops-ui/package.json
+++ b/coops-ui/package.json
@@ -12,12 +12,13 @@
     "test:unit-silent": "vue-cli-service test:unit --silent"
   },
   "dependencies": {
-    "@babel/polyfill": "^7.0.0-rc.1",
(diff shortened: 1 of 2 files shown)


WalterMoar:
We should open a ticket for this so that it can be tracked for time, etc.

Does everything still work with IE and Edge? I think the address component layout might have had problems with them.

severinbeauvais:
@WalterMoar Looks OK in Edge (but didn't fix address component ;)
I couldn't test in IE11 due to an auth issue.

PR #194, opened 2019-07-30
resume a Change Address filing

severinbeauvais:
*Issue #:* /bcgov/entity#982

*Description of changes:*

- used prop and watch to pass draft addresses to component
- conditionally set 'current' addresses
- used emitter to update addresses in parent
- added validation when loading draft data
- implemented PUT for draft filings
- updated unit tests

### DIFF
--- a/coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
+++ b/coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
@@ -310,6 +310,8 @@ export default class RegisteredOfficeAddress extends Vue {
 
   /**
    * Loads the office addresses from an API call.
+   *
+   * @param legalEntityNumber the identifier for the legal entity whose office addresses are fetched.
    */
   private loadAddressesFromApi (legalEntityNumber: string): void {
     if (legalEntityNumber) {
--- a/coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
+++ b/coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
@@ -83,10 +83,9 @@
 <script lang="ts">
 
 import Vue from 'vue'
-import { Component, Emit, Prop } from 'vue-property-decorator'
-
+import { Component, Emit, Prop, Watch } from 'vue-property-decorator'
 import axios from '@/axios-auth'
-
+import isEmpty from 'lodash.isempty'
 import BaseAddress from 'sbc-common-components/src/components/BaseAddress.vue'
 
 @Component({
@@ -99,12 +98,22 @@ export default class RegisteredOfficeAddress extends Vue {
   /**
    * The identifier for the legal entity that is to have its addresses retrieved from the API.
    */
-  @Prop({ default: '' }) readonly legalEntityNumber: string
+  @Prop({ default: '' })
+  readonly legalEntityNumber: string
 
   /**
    * Indicates whether the change button should be disabled or not
    */
-  @Prop({ default: false }) readonly changeButtonDisabled: boolean
+  @Prop({ default: false })
+  readonly changeButtonDisabled: boolean
+
+  /**
+   * Addresses object from the parent page.
+   * If this is null then this is a new filing; otherwise these are the addresses from a draft filing.
+   * This will be emitted back to the parent page when the addresses are updated.
+   */
+  @Prop({ default: null })
+  readonly addresses: object
 
   // The two addresses that come from the API. These are used to reset the address.
   private deliveryAddressOriginal: object = {}
@@ -128,14 +137,6 @@ export default class RegisteredOfficeAddress extends Vue {
 
   // State of the form checkbox for determining whether or not the mailing address is the same as the delivery address.
   private inheritDeliveryAddress: boolean = true
-
-  // TODO: these should be properties - no coupling to internal state.
-  public getDeliveryAddress () { return this.deliveryAddress }
-  public getMailingAddress () { return this.mailingAddress }
-
-  // load data from existing filing
-  public setDeliveryAddress (address) { this.deliveryAddress = address }
-  public setMailingAddress (address) { this.mailingAddress = address }
 
   /**
    * Lifecycle callback to set up the component when it is mounted.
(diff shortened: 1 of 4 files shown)


coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
```diff
@@ -99,12 +98,22 @@ export default class RegisteredOfficeAddress extends Vue {
   /**
    * The identifier for the legal entity that is to have its addresses retrieved from the API.
    */
-  @Prop({ default: '' }) readonly legalEntityNumber: string
+  @Prop({ default: '' })
+  readonly legalEntityNumber: string
```
WalterMoar: Much better/correct/consistent formatting - that's what happens when I take code from the web and don't think too much about it!

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
```diff
@@ -166,6 +167,23 @@ export default class RegisteredOfficeAddress extends Vue {
...
+   */
+  @Emit('update:addresses')
+  private emitAddresses (): object {
+    return { deliveryAddress: this.deliveryAddress, mailingAddress: this.mailingAddress }
+  }
+
+  /**
+   * Called when addresses property changes (ie, when parent page has loaded a draft filing).
+   */
+  @Watch('addresses')
```
WalterMoar: Instead of watching addresses and updating deliveryAddress and mailingAddress, can't we just use the sub-objects within addresses?

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
```diff
@@ -285,38 +304,42 @@ export default class RegisteredOfficeAddress extends Vue {
...
       axios.get(url)
         .then(response => {
           if (response && response.data) {
-            if (response.data.deliveryAddress) {
-              this.deliveryAddressOriginal = { ...response.data.deliveryAddress }
-              this.deliveryAddress = { ...response.data.deliveryAddress }
+            const deliveryAddress = response.data.deliveryAddress
+            if (deliveryAddress) {
+              this.deliveryAddressOriginal = { ...deliveryAddress }
+              // If parent page loaded draft before this API call, don't overwrite draft delivery address.
```
WalterMoar: Hopefully it won't be long until we can move this API call out of the class, and into the parent. It will make things so much simpler.

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
```diff
@@ -285,38 +304,42 @@ export default class RegisteredOfficeAddress extends Vue {
...
     this.deliveryAddress = { ...this.deliveryAddressOriginal }
     this.mailingAddress = { ...this.mailingAddressOriginal }
+    this.emitAddresses()
     this.emitModified()
   }
 
   /**
    * Loads the office addresses from an API call.
-   *
-   * @param legalEntityNumber the identifier for the legal entity whose office addresses are fetched.
```
WalterMoar: Why remove this JSDoc?

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
```diff
@@ -285,38 +304,42 @@ export default class RegisteredOfficeAddress extends Vue {
...
+            if (deliveryAddress) {
+              this.deliveryAddressOriginal = { ...deliveryAddress }
+              // If parent page loaded draft before this API call, don't overwrite draft delivery address.
+              // Otherwise this API call finished before parent page loaded draft, or parent page won't
+              // load draft (ie, this is a new filing) -> initialize delivery address.
+              if (isEmpty(this.deliveryAddress)) this.deliveryAddress = { ...deliveryAddress }
             } else {
-              console.log('loadAddressesFromApi() error - invalid Delivery Address')
-              console.log(response.data.deliveryAddress)
+              console.log('loadAddressesFromApi() error - invalid Delivery Address =', deliveryAddress)
```
WalterMoar: Nice - I just learned this recently, so much better and still able to inspect the object in the browser console!

coops-ui/src/views/AnnualReport.vue
```diff
@@ -408,25 +426,51 @@ export default {
...
+          let authStub: string = this.authURL
+          if (!(authStub.endsWith('/'))) {
+            authStub = authStub + '/'
+          }
+          let payURL = authStub + 'makepayment/' + payRequestId + '/' + encodeURIComponent(returnURL)
+          // TODO: need to check if pay UI is reachable, else display modal dialog
+          window.location.href = payURL
+        }).catch(error => {
+          console.error('submit() error =', error)
+          if (error.response && error.response.status === 402) {
```
WalterMoar: We should consider the value of using the http-status-codes library, but I think at some point we're probably going to want a consolidation of our API-calling code.

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
severinbeauvais: Yes we can read the sub-objects but we can't write to them (Vue error). See:
https://stackoverflow.com/questions/39868963/vue-2-mutating-props-vue-warn
https://michaelnthiessen.com/avoid-mutating-prop-directly/

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
severinbeauvais: Hmm, yes and no. This component needs to fetch the "current" addresses, either as a starting point for a new filing or a comparison for a draft filing. To get around this, we would have to pass in 2 addresses (the draft address and the current address). I don't know which is cleaner...

coops-ui/src/components/AnnualReport/RegisteredOfficeAddress.vue
severinbeauvais: Oops. At one point I used the legalEntityNumber directly in this function (instead of passing it in), and then I reverted that code and forgot to restore the @param. I have push another commit to restore it.

coops-ui/src/views/AnnualReport.vue
severinbeauvais: Great idea!
Done.

PR #198, opened 2019-07-31
Standalone filings - Directors 

katiemcgoff:
*Issue #:* /bcgov/entity#946

*Description of changes:*
First batch of changes for standalone filings - directors only. 

There are known bugs and no tests yet written for the standalone directors filing, however all existing Directors component pass (and all others). I will continue to work on this, but this batch of changes enables others to work on the Directors component without too many conflicts.


coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -223,11 +440,82 @@ export default {
   computed: {
     ...mapState(['corpNum', 'agmDateValid']),
 
+    // today's date as YYYY-MM-DD, current timezone
+    today () {
+      let today = new Date()
+      today = '' +
+        today.getFullYear() + '-' +
+        (today.getMonth() * 1 + 1).toString().padStart(2, '0') + '-' +
+        today.getDate().toString().padStart(2, '0')
```
WalterMoar: Just an aside - I think we should eventually look into using some sort of date library, like Moment. We'll be doing a lot of date manipulation, comparison, etc. Something to ponder, for future work.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+            to="/dashboard">
+            Cancel
+          </v-btn>
+        </div>
+      </v-container>
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
```
WalterMoar: I think that from now on, any new components should use the vue-class-component style. Everything is going that way eventually, so it will be less re-work later.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+      // Temp fix Added to remove cors error while invoking mock endpoint.
+      // To be removed when the API is in place
+      let config = {
+        headers: {
+          'Content-Type': 'text/plain'
+        }
+      }
+
+      axios.post(this.corpNum + '/filings', filingData, config).then(res => {
+        let payRequestId:String = res.data.filing.header.paymentToken
```
WalterMoar: Should be "let payRequestId: string" - the uppercase types should not be used.

I've been leaving a space after the colon, but that just seems to be the common convention and there is unfortunately no linting around it. I'm in favour of this convention, but I can be convinced otherwise - I do feel strongly about choosing one way and using it consistently (and that's where the linter would help).

> https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

(Actually should be "const" rather than "let", but in this case the next line looks like temporary code that changes the value, so const won't work until the temporary assignment is removed).

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+        payRequestId = '189'
+        let returnURL = window.location.origin + '/standalone-directors?pay_id=' + payRequestId
```
WalterMoar: We should always be using const, except for cases where let is actually needed.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+        let payURL = authStub + 'makepayment/' + payRequestId + '/' + encodeURIComponent(returnURL)
+        window.location.href = payURL
+      }).catch((error) => {
+        // TODO : To Decide how and where to display the error message from API
+        console.log(error)
+        this.dialog = true
+      })
+    },
+
+    toggleFiling (setting, filing) {
```
WalterMoar: Something for us to eventually figure out: this (and other code) is duplicated from AnnualReport.vue. Can we have this kind of common code in a base TypeScript class (or Component? does that work?).

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+        this.toggleFiling('remove', 'OTCDR')
+      }
+    },
+
+    directorFormValid (val) {
+      this.setValidateFlag()
+    },
+
+    filingData: function (val) {
+      console.log('StandaloneDirectorsFiling, filingData =', val)
```
WalterMoar: Yet again something to ponder - we don't want console.log code going to production. I think there are ways to strip it out, but if we do so in this instance would we still have the watcher? Just thinking about keeping it as simple as possible / how valuable is this code (yes, it's probably a carryover from AnnualReport).

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+    Directors,
+    SbcFeeSummary,
+    Affix
+  },
+
+  data () {
+    return {
+      showLoading: false,
+      loadingMsg: 'Redirecting to PayBC to Process Your Payment',
+      directorsChange: false,
```
WalterMoar: It looks like this is set by a callback, and then action is taken in a watcher. Would it be simpler to drop the data variable, and just do the watcher action in the callback?

coops-ui/src/views/StandaloneDirectorsFiling.vue
katiemcgoff: Like instead of StandaloneDirectorsFiling, use standalone-directors-filing as the name of the component?

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+  components: {
+    Directors,
+    SbcFeeSummary,
+    Affix
+  },
+
+  data () {
+    return {
+      showLoading: false,
+      loadingMsg: 'Redirecting to PayBC to Process Your Payment',
```
WalterMoar: This looks like a static value, does it need to be in data or can it just be a variable (that is, does it need to be reactive?). Or, can it just be static text in the template itself?

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import axios from '@/axios-auth'
+import Directors from '@/components/AnnualReport/Directors.vue'
+import { Affix } from 'vue-affix'
+import SbcFeeSummary from 'sbc-common-components/src/components/SbcFeeSummary.vue'
+import { mapState, mapActions, mapGetters } from 'vuex'
```
WalterMoar: It looks like mapGetters is unused.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+
+  components: {
+    Directors,
+    SbcFeeSummary,
+    Affix
+  },
+
+  data () {
+    return {
+      showLoading: false,
```
WalterMoar: (Going to assume this is in progress - it is always false and doesn't seem to get set anywhere)

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
+<template>
+  <div>
+    <v-dialog  v-model="dialog" width="50rem">
+      <v-card>
+        <v-card-title>Error</v-card-title>
+
+        <v-card-text>
+          An error occured during processing. Please try later
```
WalterMoar: Typo in "occurred".

coops-ui/src/views/StandaloneDirectorsFiling.vue
katiemcgoff: I might (by the end of this implementation) be able to break this out into a generic "Filing" component or class. For now you are right, there's a lot copied and I don't love it either but time is crunched. I'm also adding another standalone after this batch is approved, for standalone change of address, so it will be worse :)

coops-ui/src/views/StandaloneDirectorsFiling.vue
katiemcgoff: Will do. I actually copied this from Annual Report so will make the change in both.

You've also alerted me to the hardcoded payRequestId that snuck in (below) - I will take that out it's not supposed to be there.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -0,0 +1,301 @@
...
+      loadingMsg: 'Redirecting to PayBC to Process Your Payment',
+      directorsChange: false,
+      filingData: [],
+      dialog: false,
+      lastFilingDate: 'something' // TODO
+    }
+  },
+
+  computed: {
+    ...mapState(['noAGM', 'regOffAddrChange',
```
WalterMoar: Looks like the following aren't used: noAGM, regOffAddrChange, validated, currentARStatus. Not sure why we'd need something like addressesFormValid and directorFormValid, so I assume this is still in progress?

coops-ui/src/components/AnnualReport/Directors.vue
katiemcgoff: Good point - I actually think that the AGMDate component might do this better, but I was doing this at midnight last night :) Will re-evaluate with the next batch of changes because I have more date stuff to add.

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -25,7 +25,7 @@ describe('Directors.vue', () => {
     store.state.corpNum = 'CP0001191'
 
     // GET directors
-    sinon.stub(axios, 'get').withArgs('CP0001191/directors')
+    sinon.stub(axios, 'get').withArgs('CP0001191/directors?date=2019-04-01')
```
WalterMoar: If this file is being touched, I'd love to see the three lines of "data-app" code from, say, RegisteredOfficeAddress (lines 17-19) added to make the console output a little bit quieter.

coops-ui/src/views/StandaloneDirectorsFiling.vue
katiemcgoff: going to have to rework/ideally reuse the errors to match more up-to-date AR filing, so ignoring for now.

coops-ui/src/views/StandaloneDirectorsFiling.vue
katiemcgoff: nice, I changed it

coops-ui/src/views/StandaloneDirectorsFiling.vue
WalterMoar: (as discussed - use actual TypeScript classes).

coops-ui/tests/unit/Directors.spec.ts
WalterMoar: Yay!

coops-ui/tests/unit/RegisteredOfficeAddress.spec.ts
```diff
@@ -19,7 +19,6 @@ app.setAttribute('data-app', 'true')
 document.body.append(app)
 
 describe('RegisteredOfficeAddress', () => {
-
```
WalterMoar: Thanks for catching this - I'll check with Serevin about his linter functioning or not.

severinbeauvais: @WalterMoar Yes my linter is functioning, but I have to explicitly call `npm run lint:nofix` before submitting ;)

PR #229, opened 2019-08-08
Resume Draft not loading directors correctly

leksmall:
*Issue #:* /bcgov/entity#983

*Description of changes:*
Issue fix - Resume Draft not loading directors correctly
Changed the styling of certify to be in sync with the rest of the page
Corrected the logic for enabling save and save as draft to exclude the certify section


coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -447,7 +448,7 @@ export default {
   },
 
   computed: {
-    ...mapState(['corpNum', 'lastPreLoadFilingDate', 'currentDate', 'filingHistory']),
+    ...mapState(['corpNum', 'lastPreLoadFilingDate', 'currentDate', 'filingHistory', 'currentARStatus']),
```
katiemcgoff: So the Directors component is used in the standalone filing too now, which does not use currentARStatus. I'd suggest either a) passing the status to the directors component as a prop, or b) change currentARStatus to be for all filings, so currentFilingStatus, and set it in the TodoList component for standalone filings as well.

leksmall: This is just in the context of the AR as we discussed yesterday

leksmall: Related to this, The director component is now adding properties required for various UI purposes in the director object. As a result, all these properties which are invalid from  director schema perspective gets pushed as part of filing data. But some of these properties may be needed when the filing is saved as a draft. This needs fixes on both Legal API and UI

leksmall: I dont think Legal API should be enforcing schema validation on a draft. We need to discuss this

katiemcgoff: Oh you're right about the AR status, I have such a short memory :S 

The work currently in development by me will remove the need for the various flags (isNew, isDirectorActive, that kind of thing) and replace it with the "actions" array, which is new in the schema. So I think we're okay. Otherwise yes you're right, we'd need to handle this otherwise from the API but for now, let's leave it as is.

So I'm approving as-is. I will handle showing the individual direct changes (ceased/name changed/etc) in my code now that your code will be in.

PR #242, opened 2019-08-12
1036 filter off legacy filings

thorwolpert:
Signed-off-by: thor wolpert <thor@wolpert.ca>

*Issue #:* /bcgov/entity#1036

*Description of changes:*
- added tests for filing additions
- added filter for go-live-date
- added go-live-date to deployment config and configmap

### DIFF
--- a/legal-api/tests/unit/models/test_filing.py
+++ b/legal-api/tests/unit/models/test_filing.py
@@ -21,6 +21,8 @@ import json
 from http import HTTPStatus
 
 import pytest
+from flask import current_app
+from sqlalchemy_continuum import versioning_manager
 
 from legal_api.exceptions import BusinessException
 from legal_api.models import Filing, User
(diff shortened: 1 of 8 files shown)


legal-api/tests/unit/models/test_filing.py
```diff
@@ -274,3 +276,59 @@ def test_get_filing_by_payment_token(session):
...
+    filing.save()
+
+    rv = Filing.get_filings_by_status(business.id, [Filing.Status.COMPLETED.value])
+
+    assert rv
+    assert rv[0].status == Filing.Status.COMPLETED.value
+
+
+# testdata pattern is ({str: environment}, {expected return value})
+TEST_FILING_GO_LIVE_DATE = [
```
severinbeauvais: Because dates in JS (and possibly Python) are fickle, how about testing "on the day of" and "one day before" to include/exclude filings?
In other words, if Go Live Date is 2019-08-12, test with a filing date of 2019-08-12 (should be included) and 2019-08-11 (should not be included)?
I just think +10 and -10 won't catch edge errors :)

severinbeauvais:
^^^ The conversations above were resolved face-to-face. In short, it was agreed that no change is needed because:
- there are expected to be no edge cases (ie, same dates)
- Python dates have time 00:00 already
- this is temporary code that will be removed after MVP and rework

PR #254, opened 2019-08-15
implemented AR PDF page numbering

severinbeauvais:
*Issue #:* /bcgov/entity#1045

*Description of changes:*
- implemented Annual Report PDF page numbering using the CSS page-margin at-rule approach

### DIFF
--- a/legal-api/report-templates/annualReport.html
+++ b/legal-api/report-templates/annualReport.html
@@ -281,7 +281,7 @@
         <tr class="footer-gold">
           <td class="spacer"></td>
           <td>Incorporation Number {{ business.identifier }}</td>
-          <td class="align-right"></td>
+          <td></td>
         </tr>
       </table>
     </footer>
--- a/legal-api/report-templates/annualReport.html
+++ b/legal-api/report-templates/annualReport.html
@@ -8,12 +8,12 @@
     <style>
 
       @page {
+        font-family: Arial, Helvetica, sans-serif;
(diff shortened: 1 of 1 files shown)


legal-api/report-templates/annualReport.html
```diff
@@ -260,21 +269,13 @@
       <tfoot>
         <tr class="bcgov-footer-row bcgov-blue-bg">
           <td></td>
-          <td colspan="2">
-            <span>BC Registries and Online Services</span>
-          </td>
-          <td class="align-right">
-            {{ environment }}
-          </td>
+          <td colspan="2"><span>BC Registries and Online Services</span></td>
```
WalterMoar: You might have had an old file - this span isn't needed.

severinbeauvais: You're right. I fixed my base and now it's OK!

legal-api/report-templates/annualReport.html
```diff
@@ -276,24 +274,14 @@
...
-          </td>
-          <td>
-            Incorporation Number {{ business.identifier }}
-          </td>
-          <td>
-            <div class="page-number align-right"></div>
-          </td>
+          <td class="spacer"></td>
+          <td>Incorporation Number {{ business.identifier }}</td>
+          <td class="align-right"></td>
```
WalterMoar: Can you remove the class on this?

PR #281, opened 2019-08-21
replaced deprecated XMLHttpRequest

severinbeauvais:
*Issue #:* /bcgov/entity#855

*Description of changes:*
-changed synchronous XMLHttpRequest() to axios.get()
-used promise to load Vue after fetching config
-added helper to fetch config
-saved config in session storage
-related updates and cleanup

### DIFF
--- a/coops-ui/src/util/config-helper.ts
+++ b/coops-ui/src/util/config-helper.ts
@@ -1,0 +1,42 @@
+import axios from '@/axios-auth'
+
+export default {
+  /**
+   * fetch config from API
+   */
+  fetchConfig () {
+    const url = '/config/configuration.json'
+    const headers = {
+      'Accept': 'application/json',
+      'ResponseType': 'application/json',
+      'Cache-Control': 'no-cache'
+    }
+
+    return axios
+      .get(url, { headers })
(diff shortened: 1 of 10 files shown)


coops-ui/src/util/config-helper.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        const payApiUrl = response.data['PAY_API_URL']
+        sessionStorage.setItem('PAY_API_URL', payApiUrl)
+        console.log('Set payApiUrl to: ' + payApiUrl)
+
+        const addressCompleteKey = response.data['ADDRESS_COMPLETE_KEY']
+        window['addressCompleteKey'] = addressCompleteKey
+        console.log('Set addressCompleteKey')
+      })
+      .catch(err => {
```
WalterMoar: Can this catch be removed? Otherwise everything looks good.

severinbeauvais: Yes, if it removed then the error bubbles up.

In ticket 913 I'm reworking some of the data loading and I have removed the catch from here. I have added a catch in main.ts instead.

PR #310, opened 2019-09-04
Simplify the Certify component and add tests (#1266)

WalterMoar:
*Issue #:* /bcgov/entity#1266

*Description of Certify changes:*
  - Removed the shadow property fields to simplify the component.
  - All the watchers could be removed once above simplification done.
  - Added a property for the `currentDate`, so that (1) the component isn't tied to VueX, and (2) the parent component isn't required to have a specific named variable in the store for the current date.
  - Added a `valid` event so the parent component knows if the form is complete or not. Will be returned on create so that the parent knows if the properties are valid or not.
  - Added unit tests.

*Description of AnnualReport changes:*
  - Passing in `currentDate` as a property.
  - Added a `certifyFormValid` variable - there shouldn't be any need to store any of these in the store, but leaving the others as-is for now.

### DIFF
--- a/coops-ui/src/components/AnnualReport/Certify.vue
+++ b/coops-ui/src/components/AnnualReport/Certify.vue
@@ -78,15 +78,15 @@ export default class Certify extends Vue {
   // Emit an update event.
   @Emit('update:isCertified')
   private emitIsCertified (val: boolean): boolean {
-    this.emitValid(this.certifiedBy && !!val)
+    this.emitValid(this.trimmedCertifiedBy && !!val)
 
-    return !!val
+    return val
   }
 
   // Emit an event indicating whether or not the form is valid.
   @Emit('valid')
   private emitValid (val: boolean): boolean {
-    return !!val
+    return val
   }
 }
 
--- a/coops-ui/src/components/AnnualReport/Certify.vue
+++ b/coops-ui/src/components/AnnualReport/Certify.vue
@@ -7,14 +7,18 @@
         </label>
         <div class="value certifiedby">
           <v-text-field
-            id="certified-by-textfield"
-            v-model="certifyTextField"
-            label="Name of current director, officer, or lawyer of the association"
             box
+            id="certified-by-textfield"
+            label="Name of current director, officer, or lawyer of the association"
+            :value="certifiedBy"
+            @input="emitCertifiedBy"
           />
         </div>
       </div>
-      <v-checkbox v-model="certifyCheckbox">
+      <v-checkbox
+        :value="isCertified"
+        @change="emitIsCertified"
+      >
         <template slot="label">
           <div class="certify-stmt">
             I, <b>{{trimmedCertifiedBy || '[Legal Name]'}}</b>, certify that I have relevant knowledge of the
@@ -33,24 +37,11 @@
 </template>
 
 <script lang="ts">
-import { Component, Vue, Prop, Watch, Emit } from 'vue-property-decorator'
-import { mapState } from 'vuex'
 
-@Component({
-  computed: {
-    // Property definition for runtime environment.
-    ...mapState({ currentDate: 'currentDate' })
-  }
-})
+import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
+
+@Component
 export default class Certify extends Vue {
-  // Local definition of computed property for static type checking.
-  // Use non-null assertion operator to allow use before assignment.
-  readonly currentDate!: string
-
-  // Model properties for the Text Field and Checkbox controls.
-  private certifyTextField: string = ''
-  private certifyCheckbox: boolean = false
-
   // Props passed into this component.
   @Prop({ default: '' })
   private certifiedBy: string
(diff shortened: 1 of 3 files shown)


coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -33,83 +37,59 @@
...
+
     return val
   }
 
   // Emit an update event.
   @Emit('update:isCertified')
-  private emitIsCertified (val) {
-    return val
+  private emitIsCertified (val: boolean): boolean {
+    this.emitValid(this.certifiedBy && !!val)
```
severinbeauvais: This should check this.trimmedCertifiedBy instead.

coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -33,83 +37,59 @@
...
+
+    return !!val
+  }
+
+  // Emit an event indicating whether or not the form is valid.
+  @Emit('valid')
+  private emitValid (val: boolean): boolean {
+    return !!val
```
severinbeauvais: Why is !! needed here? The val passed in is already boolean, right?

coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -33,83 +37,59 @@
...
   }
 }
+
```
severinbeauvais: Interesting. My linter complains about whitespace below the <script> tag so I don't include one here. You included whitespace in both places. Did `npm run lint:nofix` pass?

coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -33,83 +37,59 @@
...
-    this.emitIsCertified(val)
+  private get trimmedCertifiedBy (): string {
+    return this.certifiedBy && this.certifiedBy.trim()
   }
 
   // Emit an update event.
   @Emit('update:certifiedBy')
-  private emitCertifiedBy (val) {
+  private emitCertifiedBy (val: string): string {
+    this.emitValid(val && val.trim() && this.isCertified)
```
severinbeauvais: In other places you use this.trimmedCertifiedBy and here you don't. I'd prefer to see consistency.

coops-ui/tests/unit/Certify.spec.ts
```diff
@@ -0,0 +1,172 @@
+//
```
severinbeauvais: Is this header required? If so, please create a ticket to update all the other files. Thx!

coops-ui/tests/unit/Certify.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+import Vue from 'vue'
+import Vuetify from 'vuetify'
+
+import { mount, Wrapper } from '@vue/test-utils'
+
+import Certify from '@/components/AnnualReport/Certify.vue'
+
+Vue.use(Vuetify)
+
+// Boilerplate to prevent the complaint "[Vuetify] Unable to locate target [data-app]"
```
severinbeauvais: This boilerplate is not needed in every file. Did you check without it?

coops-ui/tests/unit/Certify.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+ * Returns the last event for a given name, to be used for testing event propagation in response to component changes.
+ *
+ * @param wrapper the wrapper for the component that is being tested.
+ * @param name the name of the event that is to be returned.
+ *
+ * @return the value of the last named event for the wrapper.
+ */
+function getLastEvent (wrapper: Wrapper<Certify>, name: string): any {
+  const eventsList: Array<any> = wrapper.emitted(name)
+  const events: Array<any> = eventsList[eventsList.length - 1]
```
severinbeauvais: What happens if an event was never emitted? :P

coops-ui/src/components/AnnualReport/Certify.vue
WalterMoar: Yes! Also Indicates that there are tests missing for all-whitespace certifiedBy, so I added them.

coops-ui/src/components/AnnualReport/Certify.vue
WalterMoar: Not needed, likewise with the one in emitIsCertified.

coops-ui/src/components/AnnualReport/Certify.vue
WalterMoar: trimmedCertifiedBy is the computed value for certifiedBy, but in this case val != certifiedBy yes. It has to be emitted first.

coops-ui/tests/unit/Certify.spec.ts
WalterMoar: Required? No idea. Probably a decision for the team to make.

coops-ui/tests/unit/Certify.spec.ts
WalterMoar: I didn't check - works fine without it.

coops-ui/tests/unit/Certify.spec.ts
WalterMoar: Good question - "it explodes" is the answer.

The tests always test for existence before calling getLastEvent,, but why not just let it fail? The resulting stack trace is enough information to find the caller and determine which event is missing.

coops-ui/src/components/AnnualReport/Certify.vue
WalterMoar: I never run the linter manually, the IDE indicates when there are problems. However, running it manually does not produce a problem. The team should probably pick a lane when it comes to the IDE.

coops-ui/tests/unit/Certify.spec.ts
WalterMoar: It is required. Opened #1291 as tech debt.

WalterMoar:
> Looks good EXCEPT that we should discuss the copyright header as you may be advised to remove it.

We need the copyright/license. Opened #1291 as tech debt.

PR #323, opened 2019-09-16
1271 use new authorization api

thorwolpert:
*Issue #:* /bcgov/entity#1271

*Description of changes:*
- implemented call to new authorization API calls


legal-api/src/legal_api/services/queue.py
```diff
@@ -92,12 +92,8 @@ def init_app(self, app, loop=None,
 
     def teardown(self, exception):  # pylint: disable=unused-argument; flask method signature
         """Destroy all objects created by this extension."""
-        try:
-            this_loop = asyncio.get_event_loop()
-            this_loop.run_until_complete(self.close())
-        except:
-            pass
-        # await self.close()
+        this_loop = asyncio.get_event_loop()
```
katiemcgoff: This undoes the change that we approved from Freshworks to wrap teardown in try/catch. I believe we still want the try/catch.

thorwolpert: I'll change the way this does it in another PR. A bare except wouldn't pass any of the Linters ... so we'll need to check other PRs in the future to ensure clean linting is enforced

PR #324, opened 2019-09-16
refactored Directors from JS to TS

severinbeauvais:
*Issues #:* /bcgov/entity#1094

*Description of changes:*

This is the first part of the subject 'modernization' ticket (which is the only thing that can be delivered in this sprint). There will be a continuation ticket to further modernize the subject component.

The changes include:
- refactored Directors from JS to TS
- implemented `@Emit` functions
- renamed util/ to utils/
- renamed/refactored DateUtils to date-mixin
- created mixin to allow TS inheritance of external mixins (eg, Vue2Filters)
- removed Vue2Filters external mixin
- created mixins/ and moved some files there
- changed misc `@return` to `@returns`


coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -61,7 +61,7 @@ export default class Certify extends Vue {
 
   /**
    * Computed value.
-   * @return The trimmed "Certified By" string (may be '').
+   * @returns The trimmed "Certified By" string (may be '').
```
WalterMoar: Just curious: VSCode doesn't like @return? PyCharm is fine with the synonym (it originated in JavaDoc). Good to be consistent and use the more common one, though :)

severinbeauvais: Actually, VS Code allows anything that starts with @ :/
So I changed all of these to use the JSDoc tag (https://jsdoc.app/tags-returns.html).

severinbeauvais:
> Style-wise I didn't look too closely as Directors.vue as it seems to be a work in progress. The big thing I'm wondering about is if it should be broken into two components: a person (who in this case is a director); and a list of those persons.
> 
> While I understand the desire to fix little things like the JsDoc @returns, it does make merging more tricky when unrelated files are touched. Hopefully the other team is aware of it.

We have been talking with the other team and they aren't working on this component. The minor changes to the other components should be trivial merges (I hope).

As for splitting up this component, that makes sense (and it's documented in the ticket). I only had a day to do something here so I thought I'd attack the TS refactor since it required solving a few more technical hurdles, meaning all UI devs can now profit from this :)

PR #338, opened 2019-09-24
Entity Filter/ARDate & BackEnd provisions for both

cameron-freshworks:
*Issue #:* /bcgov/fwben 

*Description of changes:*

*FRONT-END*
* Implemented an entity filter mixin to use throughout the frontend components to handle to varying entity types and what will be rendered.

* Implemented the use of the filter on the AGMDate Component and ARDate Component.

* Implemented a ts folder that houses an enum used for entityTypes. This is for readability and consistency throughout the use of these entity types. When updating or changing the name of entity types it only has to be done in one location.

* Updated unit testing.

* Set the entityType to state along with other business related info in the App.vue.

** WARNING **

* Currently the data being returned from the API with regards to legalType is null. To get around the lack of a value here, the type has been set to a COOP. This placeholder will be swapped out when the legalType data is being returned with values.

*BACK-END*

* added logic for AR Date and legalType for BCORP 
* modified task list for BCORP
* remove AGM checking/validation for non-coop entities


severinbeauvais:
This PR has a lot of changes in common with https://github.com/bcgov/lear/pull/337. That means one PR can't be merged without the other one (or without merge conflicts).
Please see comments in PR 337 as many of them apply here.
Please let me know when all is resolved and ready for re-review. Thanks! :)

cameron-freshworks:
Thanks @severinbeauvais , good catch. That was a small mistake on his part and he is withdrawing that PR. It's destination was not meant for bcgov/master branch.

coops-ui/public/config/local-configuration.json
```diff
@@ -1,7 +1,7 @@
 {
   "API_URL": "http://localhost:2015/api/v1/businesses/",
   "AUTH_URL": "http://localhost:2015/auth/",
-  "AUTH_API_URL": "https://auth-api-test.pathfinder.gov.bc.ca/api/v1/",
+  "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "PAY_API_URL": "https://pay-api-test.pathfinder.gov.bc.ca/api/v1/",
```
severinbeauvais: Should you be use the dev Pay API URL also?

coops-ui/src/App.vue
```diff
@@ -226,6 +228,9 @@ export default {
     storeEntityInfo (response) {
       if (response && response.data && response.data.business) {
         this.setEntityName(response.data.business.legalName)
+        // TODO: Replace placeholder with `response.data.business.legalType` when legalTypes are seeded.
```
severinbeauvais: Severin: Production code (ie, code committed to master branch) should not contains TODOs (although IMHO they can contain 'FUTURE' comments for work to be done in a future task).

Cameron: Definitely agree. A lot of work going forwards depends on legalTypes having values. Until that value is seeded in production i'm not sure how else to move forward, when the only repo we have is to PR into bcgov/master. Should we wait until that is done before moving forward?

Severin: As Katie mentioned in RC, perhaps this code change should be in a separate branch to keep all functionality together, and then merged into master as a whole? We can discuss as a team if helpful.

coops-ui/src/components/AnnualReport/BCorp/ARDate.vue
```diff
@@ -0,0 +1,41 @@
...
+import { mapState } from 'vuex'
+import DateMixin from '@/mixins/date-mixin'
+
+@Component({
+  computed: {
+    ...mapState(['nextARDate'])
+  }
+})
+export default class ARDate extends DateMixin {
+    readonly nextARDate: string
```
severinbeauvais: Severin: You probably need to use the non-null assertion operator here so you don't get an error in the template when you use 'nextARDate' before it's assigned here. See AnnualReport/AGMDate.vue for an example of this.

Cameron: Easily done, but not sure where this error is occurring in this instance with regards to ARDate

Severin: I can see in the template that you use `this.nextARDate`. I'm surprised you're not getting an error (lint or runtime) about using that while it's technically undefined.

coops-ui/src/components/AnnualReport/BCorp/ARDate.vue
```diff
@@ -0,0 +1,41 @@
...
+
+    .timeline
+      display: flex
+      justify-content space-between
+
+      label
+        font-weight: 500
+
+      .date
+        font-family: "BC Sans"
```
severinbeauvais: This is incorrect because there is no fallback font stack.
Please discuss this with @jeznorth on the reasoning for this font here. This should probably be controlled by the application instead of in this (sub)component.

coops-ui/src/mixins/date-mixin.ts
```diff
@@ -51,4 +51,15 @@ export default class DateMixin extends Vue {
...
+
+  /**
+   * Formats a simple date string (YYYY-MM-DD) to (Month Day, Year) for readability.
+   *
+   * @param date The date string to format.
+   * @returns The re-formatted date string without the day name.
+   */
+  toReadableDate (date: string): string {
+    const regex = / (?!.* )/
+    return (new Date(date).toDateString().split(' ').slice(1).join(' ').replace(regex, ', '))
```
severinbeauvais: Severin: This statement is hard to read and understand. If you're updating this file, I recommend you separate the parts of the algorithm, as is done with other functions in this module :)

Cameron: Certainly can break down the string manipulation and regex, absolutely!

coops-ui/src/ts/enums/entityTypes.ts
```diff
@@ -0,0 +1,4 @@
+export enum EntityTypes {
```
severinbeauvais: Severin: I see there is a 'ts' folder under 'src' now. I don't think we need to differentiate between source code types. Was this discussed as a team and agreed to? I will go along if the team(s) think this is the way to go but otherwise I think we can just have src/enums, src/interfaces, etc. Thoughts?

Cameron: I'm open to ideas, imo i thought was a good idea to consolidate our enums, constants and interfaces opposed to them living individually scattered throughout components, we can certainly name the root folder for these values anything we like.

Severin: No, I'm just asking to remove the /ts folder level. By all means let's use enums/, interfaces/ etc to keep those specific components together.

coops-ui/src/views/AnnualReport.vue
```diff
@@ -234,7 +250,10 @@ export default {
       filingPaying: false,
       haveChanges: false,
       saveErrors: [],
-      saveWarnings: []
+      saveWarnings: [],
+
+      // Constant properties
+      EntityTypes
```
severinbeauvais: Severin: Is this a constant, a property, a type or an enum? This has been called all of these :) May I suggest a consistent label/comment for this?

Cameron: This could definitely be more clear, good catch. Will update this!

coops-ui/tests/unit/ARDate.spec.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint promise/param-names: 0, prefer-promise-reject-errors: 0 */
```
severinbeauvais: Severin: Overriding ESLint warnings/errors should only be used as a last resort. Is it not possible to fix the flagged warnings/errors?

Cameron: Yes, not sure how this got in here. Will remove! Thanks for the catch.

coops-ui/tests/unit/ARDate.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    wrapper.destroy()
+  })
+
+  it('succeeds when the Annual report date outputs are correct', () => {
+    const wrapper = shallowMount(ARDate, { store })
+    const vm: any = wrapper.vm
+    const regex = / (?!.* )/
+    const today = new Date().toDateString().split(' ').slice(1).join(' ').replace(regex, ', ')
+
+    expect(vm.$el.querySelector('.ar-date').textContent).toContain('Sep 18, 2020')
```
severinbeauvais: Severin: Does this test pass today (Sep 24)?
To fix this, instead of doing new Date(), you can create a specified date and then check that the element contains it.

Cameron: Certainly, that change could be made! This field however validates against the current day and is unique everyday, by hardcoding the input and output were not exactly testing the function behind the calculation. What are your thoughts?

Severin: I was thinking you'd do something like `new Date(2020, 09, 18, 23, 15, 53)`, which still tests the function.

cameron-freshworks:
@severinbeauvais  I tried to address your comments on the other PR. When you are able could you review this PR, and make any more comments locally. Thanks sir!

coops-ui/public/config/local-configuration.json
cameron-freshworks: We were not completing any filings as of yet and the payment calculator was not working in test. This config file is used locally only in our environments.

coops-ui/src/components/AnnualReport/BCorp/ARDate.vue
cameron-freshworks: K sounds good, i will do that. It was a request of our designer, I will research further. Thanks!

coops-ui/src/ts/enums/entityTypes.ts
cameron-freshworks: Would you recommend another name for a directory opposed to just having all the enums, constants, interfaces  living in the root level of src?

coops-ui/tests/unit/ARDate.spec.ts
cameron-freshworks: Not sure i entirely follow this recommendation, the method behind that field is that of the date-mixins `toReadableDate`, of which the argument passed is of a specific format returned from the back end and the output is specific to the return of that method. I'm certainly open to adjusting this, but i'm not entirely sure what you're recommending.

coops-ui/src/App.vue
```diff
@@ -226,6 +228,8 @@ export default {
     storeEntityInfo (response) {
       if (response && response.data && response.data.business) {
         this.setEntityName(response.data.business.legalName)
+        this.setEntityType(EntityTypes.Coop)
```
cameron-freshworks: So with the dependency on legalTypes, i set it to 'CP' so it had no impact on the current UI. When there is data present in prod for legalType it will be `response.data.business.legalType`. I set this like this for a couple reasons..
1) To have no impact on current state
2) So there is no pressure at which this data is implemented in prod 
3) Also i wanted to avoid logic that would assign a default legalType if null since we're going to have many different types in the future and i'm not sure 1 UI would fit for all types. As well it would logic we would have to go back and change anyhow 

I completely understand if we need to hold off on this feature until the data is present, so we can make this 100% complete. Love to hear your thoughts!

thorwolpert:
Too many lint errors to paste or comment in:
make pylint|wc -l
75
make flake8|wc -l
76
Please fix so that there are 0 lint errors.

thorwolpert:
There are 2 failing pytests:
- test_calc_annual_report_date
- test_business_json
Please fix.

cameron-freshworks:
Thanks Thor, he is tackling this asap!

coops-ui/src/ts/enums/entityTypes.ts
severinbeauvais: I'm thinking of this:
/coops-ui
   /src
      /asssets
      /components
      /enums
      /interfaces
      /mixins
      /store
etc

coops-ui/tests/unit/ARDate.spec.ts
severinbeauvais: On line 30, initialize date as per my example.

coops-ui/src/App.vue
severinbeauvais: Will this code work when we have other entity types coming from the BE? Or will this task/story need additional PR(s) to function? If the latter then let's see if we can wait to get a full feature to merge.
Of course, I'm open to ideas and flexibility :)

coops-ui/src/ts/enums/entityTypes.ts
cameron-freshworks: Alrighty!

coops-ui/tests/unit/ARDate.spec.ts
cameron-freshworks: If we hard code the `today` date, it no longer tests that todays date is the output & in the correct format. The purpose of this test is two-fold,
1) It is `today's` date.
2) It is output in the correct format.

I'm not saying the test is perfect by any means, but by hard coding a specific date, it fails the test, as that field expects todays date.  Maybe there is a little confusion as per the `nextARDate`? That date is predefined and validated to ensure the output is the correct format, whereas today's date is dynamic.

coops-ui/src/App.vue
cameron-freshworks: Absolutely, the idea will be to have the `response.data.business.legalType` capture the type and set it to store. Then we can use the entityFilter anywhere when we want to encapsulate components specific to certain entities, controlling the UI output. It's pretty straight forward and can allow the AR to scale with a growing number of entity types. When new types are added, we simply add the corresponding type to our `entityTypes` Enum ( so it is consistent throughout the codebase ) and use it where we have type specific components or elements in the dom.  Would love to sit down and chat about it! :)

coops-ui/tests/unit/ARDate.spec.ts
severinbeauvais: Oops! Mea culpa. I was looking at wrong date.

coops-ui/src/App.vue
severinbeauvais: It was agreed to implement this code, but now this PR is dependent on https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/1370.

coops-ui/src/components/AnnualReport/BCorp/ARDate.vue
severinbeauvais: The font declaration has been removed in the latest commit. It was decided that this requirement would be handled application-wide in https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/1369

coops-ui/src/interfaces/BCorpsIF.ts
```diff
@@ -0,0 +1,4 @@
+export interface ARDateIF {
```
severinbeauvais: Is this used anywhere?

coops-ui/src/App.vue
severinbeauvais: It was agreed to implement this change with data from the API, but now this PR is dependent on https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/1370 (which is planned for implementation ASAP).

coops-ui/src/interfaces/BCorpsIF.ts
cameron-freshworks: Thanks Severin, good catch. In this current Iteration the interface was no longer required after the way the dates were output changed. I will remove this interface, rename the file to `interfaces.ts` for us to use in the ( near ) future.

thorwolpert:
I still see 2 failing tests in the legal_api, so wont look further:
- test_bcorps_get_tasks_no_filings
- test_business_json
Please ensure these are fixed before updating PR and moving the review forward

legal-api/tests/unit/api/test_business_tasks.py
```diff
@@ -76,6 +76,17 @@ def test_get_tasks_no_filings(session, client):
     assert len(rv.json.get('tasks')) == 1  # To-do for the current year
 
 
+def test_bcorps_get_tasks_no_filings(session, client):
```
thorwolpert: This test is failing.
Review halted until it passes

legal-api/tests/unit/models/test_business.py
```diff
@@ -201,7 +202,7 @@ def test_business_find_by_identifier_no_identifier(session):
     assert b is None
 
 
-def test_business_json():
+def test_business_json(session):
```
thorwolpert: This test is now failing.
Review is halted  until tests pass

legal-api/src/legal_api/resources/business/business_tasks.py
```diff
@@ -15,15 +15,16 @@
 
 Provides all the search and retrieval from the business filings datastore.
 """
-from datetime import datetime
+from datetime import date, datetime
+import datedelta
```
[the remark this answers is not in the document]
peter-freshworks: this should be fixed in the most recent version

legal-api/src/legal_api/models/business.py
```diff
@@ -86,6 +85,12 @@ def identifier(self, value: str):
         else:
             raise BusinessException('invalid-identifier-format', 406)
 
+    @property
+    def nextAnniversary(self):
+        arFilings = Filing.get_filings_by_type(self.id, Filing.FilingType.AR.value)
+        count = len(arFilings)
+        return self.founding_date+relativedelta(years=count+1)
+
```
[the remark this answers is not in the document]
peter-freshworks: I had done that initially, but I realised last_ar_date doesn't necessarily track the anniversary (correct me if I'm wrong), but that is a good point I'll review the test scenarios with Nittin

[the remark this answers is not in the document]
thorwolpert: I'm also referring 2 things beyond that.
1. how things are counted. for a corp with 50years of history, all 48+ ARs are retrieved, serialized, sent across the network, deserialized, objects graphs created and then just size of the array is used? Either count on the server, or use founding date, the calculated field last_ar_date, etc.
2. we are currently using datedelta, not relativedelta. Always open to switching libraries, but since these 2 work a little differently it'll confuse future maintainers to have 2 packages to do the same, although slightly different things.

coops-ui/src/components/AnnualReport/BCorp/ARDate.vue
severinbeauvais: I see this is fixed now.

coops-ui/src/mixins/date-mixin.ts
severinbeauvais: Better, thanks!

coops-ui/src/views/AnnualReport.vue
severinbeauvais: Fixed, thanks.

PR #343, opened 2019-09-27
Added Delete Draft functionality to UI

katiemcgoff:
*Issue #:* /bcgov/entity#1289

*Description of changes:*
Added Delete Draft functionality to UI.

### DIFF
--- a/coops-ui/src/components/Dashboard/TodoList.vue
+++ b/coops-ui/src/components/Dashboard/TodoList.vue
@@ -405,7 +405,7 @@ export default {
     async doDeleteDraft (item) {
       let url = this.entityIncNo + '/filings/' + item.id
       await axios.delete(url).then(res => {
-        if (!res || !res.data || !res.data.filing) { throw new Error('invalid API response') }
+        if (!res) { throw new Error('invalid API response') }
 
         // reload dashboard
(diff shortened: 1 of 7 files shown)


coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -336,6 +377,57 @@ export default {
...
+      })
+    },
+
+    async doDeleteDraft (item) {
+      let url = this.entityIncNo + '/filings/' + item.id
+      await axios.delete(url).then(res => {
+        if (!res) { throw new Error('invalid API response') }
+
+        // reload dashboard
+        this.setTriggerDashboardReload(true)
```
severinbeauvais: In other places, the code simply routes to the Home URL, which reloads everything.
```
this.$router.push('/')
```
Do you think that could work here instead of setting a variable that is watched?

katiemcgoff: It didn't work when I was testing without the actual api, but actually let me check again now that I can run the api locally with that delete action implemented...

katiemcgoff: Nope just tested the router push, didn't work - no change on screen evidenced at all.

PR #354, opened 2019-10-01
Add office types 44

peter-freshworks:
*Issue #:* /bcgov/entity###

*Description of changes:*

Create a model and db migration for offices (i.e -registered/records offices), each having one or more address records. Change the json schema for change of address to use this model. mailing_address and delivery_address properties of business model moved out of model definition and into properties (to simplify code). Entity filer changed to process more than one set of addresses.


coops-ui/public/config/configuration.json
```diff
@@ -1,6 +1,6 @@
 {
-  "API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses/",
-  "AUTH_URL": "https://coops-dev.pathfinder.gov.bc.ca/auth/",
+  "API_URL": "http://localhost:2015api/v1/businesses/",
```
severinbeauvais: Please don't check in these changes. There is a second config file for using `localhost:2015`.

coops-ui/src/mixins/entityFilter-mixin.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, Vue } from 'vue-property-decorator'
```
severinbeauvais: Not a big deal, but I think the name of this file should be `entity-filter-mixing.ts`.

PR #359, opened 2019-10-02
bug fix - standalone filings only checking certify checkbox and not c…

katiemcgoff:
*Issue #:* /bcgov/entity###

*Description of changes:*
bug fix - standalone filings only checking certify checkbox and not certified-by name for validity.

### DIFF
--- a/coops-ui/src/views/StandaloneDirectorsFiling.vue
+++ b/coops-ui/src/views/StandaloneDirectorsFiling.vue
@@ -157,7 +157,7 @@ export default {
       earliestDateToSet: 'your last filing',
       isCertified: false,
       certifiedBy: '',
-      certifyFormValid: null,
+      certifyFormValid: false,
       directorFormValid: true,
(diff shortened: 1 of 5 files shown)


coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -155,6 +157,7 @@ export default {
       earliestDateToSet: 'your last filing',
       isCertified: false,
       certifiedBy: '',
+      certifyFormValid: null,
```
severinbeauvais: Should be false?

katiemcgoff: I was going with what AnnualReport does with it, but yes it could totally be false instead.

katiemcgoff: I'll change it and recommit.

PR #367, opened 2019-10-07
Add mailing address to director

peter-freshworks:
Issue #: /bcgov/entity#1413

Description of changes:
Add optional mailing address to director to facilitate directors with mailing/delivery addresses. Should only be an additive change. API and entity filer changes and their unit tests will be in a future PR


legal-api/src/legal_api/models/director.py
```diff
@@ -61,6 +63,14 @@ def json(self):
             if 'addressType' in director_address:
                 del director_address['addressType']
             d['deliveryAddress'] = director_address
+        if self.mailing_address:
+            director_mailing_address = self.mailing_address.json
+            if 'addressTYpe' in director_mailing_address:
```
thorwolpert: Should this be addressTYpe or addressType

peter-freshworks: good catch!

peter-freshworks: I was planning to add tests in a separate pull request in the interest of keeping pull requests small. I do have tests surrounding this that go along with the actual change of director filing work

PR #394, opened 2019-10-11
Dashboard UI Change for Directors

cameron-freshworks:
*Issue #:* /bcgov/entity  #1387  & #1417

*Description of changes:*

* Updated dashboard UI to accordion style for Directors.
* Implemented mailing address to directors for Bcorps.
* Refactored method of importing Mixins.
* Created a Mixin for address utility.
* Updated unit tests.
* Regression testing.

Sometimes the issue linking is off, so i've included links to the tickets & designs..

https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/1417

https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/1387

Designs: https://projects.invisionapp.com/share/38U4FN3WA2B#/screens/384879517

### DIFF
--- a/coops-ui/src/components/Dashboard/DirectorListSm.vue
+++ b/coops-ui/src/components/Dashboard/DirectorListSm.vue
@@ -15,7 +15,7 @@
               <ul class="address-details">
                 <li>{{ director.deliveryAddress.streetAddress }}</li>
                 <li>{{ director.deliveryAddress.addressCity }} {{ director.deliveryAddress.addressRegion }}
-                  {{ director.deliveryAddress.postalCode}}</li>
+                  {{ director.deliveryAddress.postalCode }}</li>
                 <li>{{ director.deliveryAddress.addressCountry }}</li>
               </ul>
             </div>
@@ -31,7 +31,7 @@
               <ul v-else class="address-details" >
                 <li>{{ director.mailingAddress.streetAddress }}</li>
                 <li>{{ director.mailingAddress.addressCity }} {{ director.mailingAddress.addressRegion }}
-                  &nbsp;&nbsp;{{ director.mailingAddress.postalCode}}</li>
+                  {{ director.mailingAddress.postalCode }}</li>
                 <li>{{ director.mailingAddress.addressCountry }}</li>
               </ul>
             </div>
--- a/coops-ui/src/components/Dashboard/DirectorListSm.vue
+++ b/coops-ui/src/components/Dashboard/DirectorListSm.vue
@@ -25,7 +25,7 @@
           <div class="director-info">
             <div class="list-item_title">Mailing Address</div>
             <div class="list-item__subtitle">
-              <span v-if="isSameAddress(director.deliveryAddress.streetAddress, director.mailingAddress.streetAddress)">
+              <span v-if="isSameAddress(director.deliveryAddress, director.mailingAddress)">
                 Same as above
               </span>
               <ul v-else class="address-details" >
--- a/coops-ui/src/components/Dashboard/DirectorListSm.vue
+++ b/coops-ui/src/components/Dashboard/DirectorListSm.vue
@@ -1,71 +1,114 @@
 <template>
-  <ul class="list list-container">
-    <li class="list-item" v-for="director in directors" v-bind:key="director.id">
-      <v-avatar color="primary" size="25">
-        <span class="white--text small">{{ director.officer.firstName.substring(0,1)}}</span>
-      </v-avatar>
-      <div class="director-info">
+  <v-expansion-panels class="list list-container" :accordion=true :multiple=true>
+    <v-expansion-panel class="address-panel" v-for="director in directors" v-bind:key="director.id">
+      <v-expansion-panel-header>
+        <v-avatar color="primary" size="25">
+        <span class="">{{ director.officer.firstName.substring(0,1)}}</span>
+        </v-avatar>
         <div class="list-item__title">{{ director.officer.firstName }} {{ director.officer.lastName }}</div>
-        <div class="list-item__subtitle">
-          <ul class="address-details">
-            <li>{{ director.deliveryAddress.streetAddress }}</li>
-            <li>{{ director.deliveryAddress.addressCity }} {{ director.deliveryAddress.addressRegion }}
-              &nbsp;&nbsp;{{ director.postalCode}}</li>
-            <li>{{ director.deliveryAddress.addressCountry }}</li>
-            </ul>
-        </div>
-      </div>
-    </li>
-  </ul>
+      </v-expansion-panel-header>
+      <v-expansion-panel-content class="list-container-content">
+        <li class="list-item" >
+          <div class="director-info">
+            <div class="list-item_title">Delivery Address</div>
+            <div class="list-item__subtitle">
+              <ul class="address-details">
+                <li>{{ director.deliveryAddress.streetAddress }}</li>
+                <li>{{ director.deliveryAddress.addressCity }} {{ director.deliveryAddress.addressRegion }}
+                  {{ director.deliveryAddress.postalCode}}</li>
+                <li>{{ director.deliveryAddress.addressCountry }}</li>
+              </ul>
+            </div>
+          </div>
+        </li>
+        <li class="list-item" v-if="entityFilter(EntityTypes.BCorp)">
+          <div class="director-info">
+            <div class="list-item_title">Mailing Address</div>
+            <div class="list-item__subtitle">
+              <span v-if="isSameAddress(director.deliveryAddress.streetAddress, director.mailingAddress.streetAddress)">
+                Same as above
+              </span>
+              <ul v-else class="address-details" >
+                <li>{{ director.mailingAddress.streetAddress }}</li>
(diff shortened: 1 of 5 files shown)


coops-ui/src/components/Dashboard/DirectorListSm.vue
```diff
@@ -1,71 +1,114 @@
 <template>
-  <ul class="list list-container">
-    <li class="list-item" v-for="director in directors" v-bind:key="director.id">
-      <v-avatar color="primary" size="25">
-        <span class="white--text small">{{ director.officer.firstName.substring(0,1)}}</span>
-      </v-avatar>
-      <div class="director-info">
+  <v-expansion-panels class="list list-container" :accordion=true :multiple=true>
```
severinbeauvais: Normally, a prop's value should be in quotes, eg, "true".
Here, since these props aren't reactive, you don't need the colon.
But really, in this case, the props just needs to exist, so you can just write "accordion multiple".

coops-ui/src/components/Dashboard/DirectorListSm.vue
```diff
@@ -1,71 +1,114 @@
...
+    <v-expansion-panel class="address-panel" v-for="director in directors" v-bind:key="director.id">
+      <v-expansion-panel-header>
+        <v-avatar color="primary" size="25">
+        <span class="">{{ director.officer.firstName.substring(0,1)}}</span>
```
severinbeauvais: Indent please.

coops-ui/src/components/Dashboard/DirectorListSm.vue
```diff
@@ -1,71 +1,114 @@
...
+    </v-expansion-panel>
+  </v-expansion-panels>
 </template>
 
 <script lang="ts">
-import { Component, Vue } from 'vue-property-decorator'
+
+// Vue Libraries
+import { Component } from 'vue-property-decorator'
+import { mixins } from 'vue-class-component'
```
severinbeauvais: Please use Mixins from vue-property-decorator instead. You can see an example of this in Directors.vue.

coops-ui/src/mixins/address-mixin.ts
```diff
@@ -0,0 +1,20 @@
...
+export default class AddressMixin extends Vue {
+  /**
+   * Compares two address objects and returns a boolean indicating
+   * if they match.
+   *
+   * @param addressA An address object.
+   * @param addressB An address object.
+   * @return boolean True if they match, false if they do not.
+   */
+  isSameAddress (addressA: Object, addressB: Object): boolean {
```
severinbeauvais: I thought about this and I like it!

But now we have a second way to compare addresses (see sameAddress() in RegisteredOfficeAddress.vue). (We also compare directors that way, but I'm undecided which way is best.)

Can you decide which way we should use to compare addresses and update the code to use the same mechanism everywhere?

coops-ui/src/components/Dashboard/DirectorListSm.vue
cameron-freshworks: Interesting, i will give it a try and change it!

coops-ui/src/components/Dashboard/DirectorListSm.vue
cameron-freshworks: Good catch, thanks!

coops-ui/src/mixins/address-mixin.ts
```diff
@@ -0,0 +1,20 @@
...
+    return isEqual(addressA, addressB)
```
severinbeauvais: So we found a problem with the "other" address comparison code: it didn't ignore the address type field. But the code sometimes worked because we weren't initializing the address type.
All this to say that I think we should initialize address type when we create the address objects, and the comparison functions will have to handle that.

coops-ui/src/components/Dashboard/DirectorListSm.vue
cameron-freshworks: Ok can do. What is the main reasoning?

coops-ui/src/mixins/address-mixin.ts
cameron-freshworks: Absolutely, i will look into the other implementation and we can discuss offline!

coops-ui/tests/unit/DirectorListSm.spec.ts
```diff
@@ -59,12 +62,90 @@ describe('DirectorListSm', () => {
...
+  it('displays multiple directors as a bcorp', done => {
+    // init store
+    store.state.entityType = EntityTypes.BCorp
+    store.state.directors = [
+      {
+        'officer': {
+          'firstName': 'Peter',
+          'lastName': 'Griffin'
+        },
+        'deliveryAddress': {
```
severinbeauvais: Can you please add a unit test to confirm that "Same as above" is displayed when the mailing address is the same as the delivery address?

coops-ui/src/mixins/address-mixin.ts
cameron-freshworks: Hmm, i like the other method but using a library like Lodash really abstracts the complexit. Also would have to agree, i like creating this sort of utility to use App wide opposed to local implementation in a specific component.  Would you prefer to see this refactor done in this ticket/pr, or should that be stand alone in another small ticket?

coops-ui/src/components/Dashboard/DirectorListSm.vue
severinbeauvais: That's how the other props work (eg, filled, flat, etc).

coops-ui/src/components/Dashboard/DirectorListSm.vue
severinbeauvais: If the linter doesn't catch this, we should add a rule.

coops-ui/src/components/Dashboard/DirectorListSm.vue
cameron-freshworks: cool, i like it!

coops-ui/src/components/Dashboard/DirectorListSm.vue
severinbeauvais: Consistency :)

I know that one package includes the other's functions, but another reviewer might question why it looks like we're using 2 different functions.

coops-ui/src/mixins/address-mixin.ts
severinbeauvais: No preference on making the change here or in a new ticket. I can see pros and cons to both.

coops-ui/src/components/Dashboard/DirectorListSm.vue
cameron-freshworks: Totally agree, makes perfect sense. Will update!

coops-ui/src/mixins/address-mixin.ts
cameron-freshworks: Ok, sounds good. Might as well do it now, will update and do some testing 👍

coops-ui/src/mixins/address-mixin.ts
cameron-freshworks: Okie doke, sounds good. Will look into that when i refactor and do the testing.

coops-ui/tests/unit/DirectorListSm.spec.ts
cameron-freshworks: Absolutely, will do!

coops-ui/src/mixins/address-mixin.ts
cameron-freshworks: After our discussion this afternoon, i've decided to leave the refactor of the existing code for a new ticket due to there being some more complexity and more thorough testing. If you are ok with that we can move this forward and refactor the `other` comparison in a new ticket?

coops-ui/tests/unit/DirectorListSm.spec.ts
cameron-freshworks: Done!

coops-ui/src/mixins/address-mixin.ts
severinbeauvais: Also, in that new ticket, please request to delete the addressType when fetching office address (in App.vue) and also in the unit tests that reference it.

Once the new ticket is created, please resolve this conversation. Thx.

coops-ui/src/mixins/address-mixin.ts
severinbeauvais: As discussed earlier, we will delete/ignore addressType in the FE.

Do you need to do any further refactoring/testing or can I resolve this conversation?

coops-ui/src/mixins/address-mixin.ts
cameron-freshworks: To follow up on this, this refactor work has been done in the following PR

coops-ui/src/mixins/address-mixin.ts
cameron-freshworks: This work and refactor has been implemented in the follow up PR. Since we concluded to not initialize `addressType` and instead filter it out, the changes were separate from this particular work, so was included in the follow up PR.  I will make notes of these changes on PR #401

PR #396, opened 2019-10-11
1484 paperfiling

thorwolpert:
*Issue #:* /bcgov/entity#1484

*Description of changes:*
Added availableOnPaperOnly boolean to the filing/header
Updated setup.cfg so that isort supports () imports


schemas/tests/unit/test_filings.py
```diff
@@ -44,25 +48,30 @@ def test_invalid_ar_filing():
...
-    # if errors:
-    #     for err in errors:
-    #         print(err.message)
+    if errors:
+        for err in errors:
+            print(err.message)
     print(errors)
 
-    assert not is_valid
+    assert is_valid
```
katiemcgoff: This one is trying to test invalid ("test_invalid_ar_filing"). You've switched it to a test for *valid* filing.

PR #400, opened 2019-10-14
1357 director name change & 1363 director address change

katiemcgoff:
*Issue #:* /bcgov/entity#1357 & /bcgov/entity#1363

*Description of changes:*
Director name change and address change - free filings - in UI.

### DIFF
--- a/coops-ui/src/components/AnnualReport/Directors.vue
+++ b/coops-ui/src/components/AnnualReport/Directors.vue
@@ -618,7 +618,7 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin) {
    */
   private get directorsFreeChange (): boolean {
     return this.directors.filter(director =>
-      director.actions.indexOf(NAMECHANGED) >= 0 || director.actions.indexOf(ADDRESSCHANGED) >= 0
+      this.isNameChanged(director) || this.isAddressChanged(director)
     ).length > 0
   }
 
@@ -743,7 +743,7 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin) {
     this.draftDate = date
   }
 
-  public getOriginalDirectors () {
+  private getOriginalDirectors () {
     this.getDirectors(true)
   }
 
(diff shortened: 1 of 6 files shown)


coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -599,6 +609,16 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin) {
     return this.directors.filter(director => director.isFeeApplied).length > 0
   }
 
+  /**
+   * Computed value.
+   * @returns Whether at least one director has a free change (name change, address change) applied.
+   */
+  private get directorsFreeChange (): boolean {
+    return this.directors.filter(director =>
+      director.actions.indexOf(NAMECHANGED) >= 0 || director.actions.indexOf(ADDRESSCHANGED) >= 0
```
severinbeauvais: You could call the helpers below instead of duplicating the logic.

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -740,11 +743,15 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin) {
     this.draftDate = date
   }
 
+  public getOriginalDirectors () {
```
severinbeauvais: This should be private unless you REALLY need to call it externally (which I would discourage since we should be using props/events to control components).

coops-ui/src/components/AnnualReport/Directors.vue
katiemcgoff: Oh ya! I'll leave it for now to get it merged but good point :S

coops-ui/src/components/AnnualReport/Directors.vue
katiemcgoff: Done.

coops-ui/src/components/AnnualReport/Directors.vue
katiemcgoff: Actually I made the change. Done.

PR #404, opened 2019-10-16
PWA first commit

leksmall:
*Issue #:* /bcgov/entity#1225

*Description of changes:*
Created custom service worker
Updated the register service worker to show the confirmation pop up and post the message to the service worker. Added call back to force reload the window once user confirms to update the app.
Updated the vue config to point to custom service worker

Note : This commit is to test whether the confirmation popup is displayed when a new version of the app is available. May need a few iterations to get this working 

### DIFF
--- a/coops-ui/src/registerServiceWorker.ts
+++ b/coops-ui/src/registerServiceWorker.ts
@@ -19,8 +19,10 @@ if (process.env.NODE_ENV === 'production') {
     updatefound () {
       console.log('New content is downloading.')
     },
-    updated () {
+    updated (registration) {
       console.log('New content is available; please refresh.')
+      let confirmationResult = confirm('New content found! Do you want to reload the app?')
+      if (confirmationResult) { registration.waiting.postMessage({ action: 'skipWaiting' }) }
     },
     offline () {
       console.log('No internet connection found. App is running in offline mode.')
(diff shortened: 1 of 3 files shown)


coops-ui/src/registerServiceWorker.ts
```diff
@@ -19,8 +19,10 @@ if (process.env.NODE_ENV === 'production') {
     updatefound () {
       console.log('New content is downloading.')
     },
-    updated () {
+    updated (registration) {
       console.log('New content is available; please refresh.')
+      let confirmationResult = confirm('New content found! Do you want to reload the app?')
+      if (confirmationResult) { registration.waiting.postMessage({ action: 'skipWaiting' }) }
```
severinbeauvais: I prefer 'if' blocks to be on a separate line unless it's really short (like the return on line 37 below, which is ok!).
In fact, I think you'll get a lint error because your have the { } block on the same line as the if.

coops-ui/src/registerServiceWorker.ts
```diff
@@ -29,4 +31,11 @@ if (process.env.NODE_ENV === 'production') {
       console.error('Error during service worker registration:', error)
     }
   })
+
+  let refreshing
```
severinbeauvais: You should probably initialize this to false.

coops-ui/src/service-worker.js
```diff
@@ -0,0 +1,10 @@
+self.__precacheManifest = [].concat(self.__precacheManifest || [])
+workbox.precaching.suppressWarnings()
+workbox.precaching.precacheAndRoute(self.__precacheManifest, {})
+
+// install new service worker when ok, then reload page.
```
severinbeauvais: See the comment I wrote in https://github.com/severinbeauvais/lear/blob/1225/coops-ui/src/service-worker.js.

PR #414, opened 2019-10-18
Fix queue problem

katiemcgoff:
*Issue #:* /bcgov/entity#1560

*Description of changes:*
CURRENTLY RUNNING IN DEV

In entity filer, if filing cannot be found (because payment ID hasn't been assigned yet) try again up to 5 times, then add the message back to the queue to be processed again.

~~The original ticket says to wait 200 ms but the code does not wait, it just tries again. So far this has been sufficient.~~

### DIFF
--- a/entity-filer/src/entity_filer/worker.py
+++ b/entity-filer/src/entity_filer/worker.py
@@ -38,7 +38,9 @@ from sqlalchemy_continuum import versioning_manager
 
 from entity_filer.config import get_named_config
 from entity_filer.filing_processors import annual_report, change_of_address, change_of_directors
-from entity_filer.service_utils import QueueException, logger
+from entity_filer.service_utils import FilingException, QueueException, logger
+
+from . import service
 
 
 def extract_payment_token(msg: nats.aio.client.Msg) -> dict:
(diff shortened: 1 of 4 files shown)


entity-filer/src/entity_filer/worker.py
```diff
@@ -57,7 +59,16 @@ def process_filing(payment_token, flask_app):
...
     with flask_app.app_context():
-        filing_submission = get_filing_by_payment_id(payment_token['paymentToken'].get('id'))
+
+        # try to find the filing 5 times before putting back on the queue - in case payment token ends up on the queue
+        # before it is assigned to filing.
+        counter = 1
+        filing_submission = None
+        while not filing_submission and counter <= 5:
+            filing_submission = get_filing_by_payment_id(payment_token['paymentToken'].get('id'))
+            counter += 1
```
kialj876: This might not matter if get_filing_by_payment_id takes some time, but don't you need to wait a second before incrementing the counter? Otherwise won't it loop through super fast and then throw the exception?

katiemcgoff: Yep you're right. I've added the check. From a few tests (with print statements) it gets to cycle 2 before it can process, so that's good, and better than running through all 5 iterations and putting it back on the queue.

PR #429, opened 2019-10-22
Reload dashboard when we return too quickly from completed filing.

katiemcgoff:
*Issue #:* /bcgov/entity#1117

*Description of changes:*
Reload dashboard when we return too quickly from completed filing.
This covers several scenarios, especially around free/unpaid filings that do not redirect to payment system.

The steps to reproduce and (rough) acceptance criteria - as described solution, not formal AC - are in the ticket.

### DIFF
--- a/coops-ui/src/components/Dashboard/TodoList.vue
+++ b/coops-ui/src/components/Dashboard/TodoList.vue
@@ -32,6 +32,11 @@
                 <span v-else-if="isPending(item)">FILING PENDING</span>
                 <span v-else-if="isError(item)">FILING PENDING</span>
               </div>
+              <div class="filing-status-2" v-if="inProcessFiling === item.id && item.id !== undefined">
+                <span>
+                  PROCESSING...
+                </span>
+              </div>
               <div class="filing-status-2" v-if="isPending(item)">
                 <span>
                   PAYMENT INCOMPLETE
@@ -47,7 +52,12 @@
             </template>
 
             <div class="list-item__actions">
-              <span v-if="isDraft(item)">
+              <span v-if="inProcessFiling === item.id && item.id !== undefined">
+                <!-- hidden button to maintain layout -->
+                <v-btn style="visibility: hidden"></v-btn>
+              </span>
+
+              <span v-else-if="isDraft(item)">
                 <v-btn id="btn-draft-resume"
                   color="primary"
                   :disabled="!item.enabled"
@@ -167,6 +177,10 @@ export default {
     }
   },
 
+  props: {
+    inProcessFiling: null
+  },
+
   computed: {
     ...mapState(['tasks', 'entityIncNo']),
 
@@ -196,6 +210,7 @@ export default {
       })
 
       this.$emit('todo-count', this.taskItems.length)
+      this.$emit('todo-filings', this.taskItems)
 
       // if this is a draft/pending/error item, emit the has-blocker-filings event to the parent component
       // this indicates that a new filing cannot be started because this one has to be completed first
--- a/coops-ui/src/views/Dashboard.vue
+++ b/coops-ui/src/views/Dashboard.vue
@@ -12,13 +12,14 @@
               <header>
                 <h2 class="mb-3">To Do <span class="text-muted">({{todoCount}})</span></h2>
               </header>
-              <todo-list @todo-count="todoCount = $event" @has-blocker-filing="hasBlockerFiling = $event" />
+              <todo-list @todo-count="todoCount = $event" @has-blocker-filing="hasBlockerFiling = $event"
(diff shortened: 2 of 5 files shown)


coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -32,6 +32,11 @@
                 <span v-else-if="isPending(item)">FILING PENDING</span>
                 <span v-else-if="isError(item)">FILING PENDING</span>
               </div>
+              <div class="filing-status-2" v-if="inProcessFiling === item.id && item.id !== undefined">
```
severinbeauvais: You probably don't need the second check since the first check would already return false (null !== undefined).

coops-ui/src/views/Dashboard.vue
```diff
@@ -12,13 +12,14 @@
               <header>
                 <h2 class="mb-3">To Do <span class="text-muted">({{todoCount}})</span></h2>
               </header>
-              <todo-list @todo-count="todoCount = $event" @has-blocker-filing="hasBlockerFiling = $event" />
+              <todo-list @todo-count="todoCount = $event" @has-blocker-filing="hasBlockerFiling = $event"
```
severinbeauvais: Given all the logic here, could you please put each on its own line?

coops-ui/src/views/Dashboard.vue
```diff
@@ -93,7 +104,82 @@ export default {
...
+      // cancel any existing timer so we can start fresh here
+      clearTimeout(this.refreshTimer)
+
+      let filingId = null
+      if (this.$route !== undefined) filingId = +this.$route.query.filing_id // if missing, this is NaN
+
+      // only consider refreshing the dashboard if we came from a filing
+      if (!filingId) return
+
+      let isInFilingHistory = this.historyFilings.filter(el => el.filingId === filingId).length > 0
```
severinbeauvais: Trivial, but using Array.prototype.find() might be a bit faster since it should return when a match is found instead of processing all elements.

coops-ui/src/views/Dashboard.vue
```diff
@@ -93,7 +104,82 @@ export default {
...
+    },
+
+    checkFilingStatus (filingId) {
+      // check whether this filing's status has changed - recursive, runs approx. every 1 second for up to 10 seconds
+
+      this.checkFilingStatusCount++
+
+      // stop this cycle after 10 iterations
+      if (this.checkFilingStatusCount > 10) {
+        this.inProcessFiling = null
```
severinbeauvais: This stops formatting the item differently, right?

coops-ui/src/views/Dashboard.vue
```diff
@@ -93,7 +104,82 @@ export default {
...
+        if (!res) {
+          // quietly fail - this error is not worth showing the customer an error
+          return false
+        }
+        // if the filing status is now COMPLETE, reload the dashboard
+        if (res.data.filing.header.status === 'COMPLETED') {
+          this.setTriggerDashboardReload(true)
+        } else {
+          // call this function again in 1 second
+          let vue = this
```
severinbeauvais: Do you really need this workaround (since you're using an arrow function in setTimeout)?
Also is there a precedent for using `vue = this`? I've only seen `'that = this` before. Just curious.

coops-ui/src/components/Dashboard/TodoList.vue
katiemcgoff: Let me check... I ran into the pending to-dos (tasks, not filings) being marked as processing but that might not be an issue anymore. I'll experiment.

coops-ui/src/views/Dashboard.vue
katiemcgoff: Yes. If 10 seconds elapses and we haven't reloaded the dashboard, setting this to null passes down to the Todo component and removes the special handling/formatting.

coops-ui/src/views/Dashboard.vue
katiemcgoff: I'll experiment. And I saw the `vue=this` in sample code, and this indicates that it's the vue component that this refers to, but I could go either way.

katiemcgoff:
> Looks really good.
> I have made a few comments but no changes are required. Up to you whether you make any more changes or merge as is.

I will check out a few of them, I want to make sure the code is as clean as possible since it's hard-ish to follow anyway. Will msg you when it's ready for re-review.

coops-ui/src/views/Dashboard.vue
severinbeauvais: As I understand it, arrow functions know about this so you don't need to save it.

PR #433, opened 2019-10-23
COD SummaryUI/Functionality

cameron-freshworks:
*Issue #:* /bcgov/entity#1414  & /bcgov/entity#1415 

*Description of changes:*

This feature adds summarized content for the change of directors filing.

Implements a Summary of Directors component and a Summary of Certification component.

Implemented TS files ( interfaces & constants).

Some Style changes to the directors list.

Added new unit tests for summary components.

Updated unit testing for modified components.

### DIFF
--- a/coops-ui/tests/unit/StandaloneDirectorsFiling.spec.ts
+++ b/coops-ui/tests/unit/StandaloneDirectorsFiling.spec.ts
@@ -152,77 +152,6 @@ describe('Standalone Directors Filing - Part 1 - UI', () => {
 
     wrapper.destroy()
   })
-
-  it('enables File & Pay button when Validated is true', () => {
-    const localVue = createLocalVue()
-    localVue.use(VueRouter)
-    const router = mockRouter.mock()
-    router.push({ name: 'standalone-directors', params: { id: '0' } }) // new filing id
-    const wrapper = mount(StandaloneDirectorsFiling, {
-      store,
-      localVue,
-      router,
-      stubs: {
-        Directors: true,
-        Certify: true,
-        Affix: true,
-        SbcFeeSummary: true,
-        ConfirmDialog: true,
-        PaymentErrorDialog: true,
-        ResumeErrorDialog: true,
-        SaveErrorDialog: true
-      },
-      vuetify
-    })
-    const vm: any = wrapper.vm
-
-    // set flag
-    vm.directorFormValid = true
-    vm.certifyFormValid = true
-
-    // set stub list of filings
-    vm.filingData.push({})
-
-    // confirm that button is enabled
-    expect(wrapper.find('#cod-file-pay-btn').attributes('disabled')).not.toBe('true')
-
-    wrapper.destroy()
-  })
-
-  it('disables File & Pay button when Validated is false', () => {
-    const localVue = createLocalVue()
-    localVue.use(VueRouter)
-    const router = mockRouter.mock()
-    router.push({ name: 'standalone-directors', params: { id: '0' } }) // new filing id
-    const wrapper = mount(StandaloneDirectorsFiling, {
-      store,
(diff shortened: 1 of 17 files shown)


katiemcgoff:
Looking at the UI first... looks like there are three lint errors
No unit tests failure, good

katiemcgoff:
The director component changes seem to stray outside the scope of these changes. I checked with Scot re. some of the button and layout changes, for example, and he says stick with the original layout. See screenshots below for changes. 

Also there is a non-functional code change that seem unnecessary, or please explain why: changed v-show changed to v-if on some buttons.

Original: 
image

New layout, please go back to original:
image

coops-ui/tests/unit/StandaloneDirectorsFiling.spec.ts
```diff
@@ -400,6 +329,81 @@ describe('Standalone Directors Filing - Part 3 - Submitting', () => {
     sinon.restore()
   })
 
+  it('enables File & Pay button when Validated is true', () => {
```
katiemcgoff: I don't think these tests should have been moved - they're UI checks (button enable/disable) rather than functions of submitting.

coops-ui/tests/unit/StandaloneDirectorsFiling.spec.ts
```diff
@@ -440,7 +445,7 @@ describe('Standalone Directors Filing - Part 3 - Submitting', () => {
       // TODO: verify that new filing was created
 
       const button = wrapper.find('#cod-file-pay-btn')
-      expect(button.attributes('disabled')).toBeUndefined()
+      expect(wrapper.find('#cod-file-pay-btn').attributes('disabled')).not.toBe('true')
```
katiemcgoff: can use the constant "button" from previous line rather than wrapper.find() again.

katiemcgoff: Same on line 512.

katiemcgoff:
I don't see any backend changes (which isn't surprising) except a requirement removed. Is this correct? 

The backend change (removing the attrs include) actually can't happen - it's true that the package isn't directly included in the project, but it's included by a different package, and if it exceeds a specific version we get some error (I can't remember what the error is but it was bad).

cameron-freshworks:
> The director component changes seem to stray outside the scope of these changes. I checked with Scot re. some of the button and layout changes, for example, and he says stick with the original layout. See screenshots below for changes.
> 
> Also there is a non-functional code change that seem unnecessary, or please explain why: changed v-show changed to v-if on some buttons.
> 
> Original:
> image
> 
> New layout, please go back to original:
> image

This new design was done as per the designs Billin had approved from Scott and Kaine, prior to the building of the UI.

Attached is a picture for reference. 
Screen Shot 2019-10-29 at 8 05 21 AM

Also, here is the link to Invision to reference all the designs. ( If you don't have permission, please let me know and i can invite you! ) 
https://projects.invisionapp.com/d/main/#/console/18216776/386689479/preview

coops-ui/tests/unit/StandaloneDirectorsFiling.spec.ts
cameron-freshworks: Sounds good, great point. I will adjust!

coops-ui/tests/unit/StandaloneDirectorsFiling.spec.ts
cameron-freshworks: Good catch, thank you!

cameron-freshworks:
> Some UI changes requested by Scott, plus other notes throughout.

Comments addressed, though we do have a discrepancy on the designs. We can sort those out after speaking with the designers :)

PR #466, opened 2019-10-31
Change validation on business identifiers to be less strict

peter-freshworks:
*Issue #:* /bcgov/entity1610

*Description of changes:*
Remove Co-op hardcoding for schema validation on business identifiers

### DIFF
--- a/schemas/src/registry_schemas/schemas/business.json
+++ b/schemas/src/registry_schemas/schemas/business.json
@@ -87,7 +87,7 @@
                     "examples": [
                         "CP1234567"
                     ],
-                    "pattern": "^[A-Z]+[0-9]{7}$"
+                    "pattern": "^[A-Z]{1,3}[0-9]{7}$"
                 },
                 "legalName": {
                     "$ref": "#/definitions/legalName"
--- a/schemas/src/registry_schemas/schemas/business.json
+++ b/schemas/src/registry_schemas/schemas/business.json
@@ -87,7 +87,7 @@
                     "examples": [
                         "CP1234567"
                     ],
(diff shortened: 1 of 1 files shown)


schemas/src/registry_schemas/schemas/business.json
```diff
@@ -87,7 +87,7 @@
                     "examples": [
                         "CP1234567"
                     ],
-                    "pattern": "^(CP)[0-9]{7}$"
+                    "pattern": "^[A-Z]+[0-9]{7}$"
```
severinbeauvais: The new regex allows from 1 to many capital alpha characters at the start of the identifier. Is this the intent? Or did you mean to allow only 2 characters? Please respond to this comment or update the code accordingly.

peter-freshworks: That is the intent. I could do a 1-3 as well if that is preferable (my thinking is this would have to be updated every time a longer prefix is added). Right now the shortest prefix is 1 character and the longest is 3

severinbeauvais: I prefer to see {1,3} at the moment. We can change it again later if needed.

Are there specific requirements for identifier format? Would it make sense to be more explicit about what is allowed, eg, (CP|BC) ?

peter-freshworks: There are 44 legal types (I think) so I think {1,3} is about as strict as we should be here without getting messy. Anything more granular is probably better suited within code with some sort of lookup table or reference object.

severinbeauvais: Agreed!

PR #479, opened 2019-11-05
1389 with queue services

peter-freshworks:
*Issue #:* /bcgov/entity#1389

*Description of changes:*
Re-add changes and jobs for future effective change of address filing


jobs/future-effective-filings/openshift/cron-file-future-effective.yml
```diff
@@ -0,0 +1,70 @@
+---
```
thorwolpert: All of our other templates are in json, please re-export as json.

jobs/future-effective-filings/utils/__init__.py
```diff
@@ -0,0 +1,39 @@
+# Copyright © 2019 Province of British Columbia
```
thorwolpert: These don't seem to be used. Also, why not just use isoformat() ?

lear-db/Dockerfile
```diff
@@ -1,3 +1,5 @@
 FROM postgres:11
 COPY lear_db.sql /docker-entrypoint-initdb.d/
 EXPOSE 5432
+RUN apt update
+RUN apt install sudo
```
thorwolpert: not sure why these are added

legal-api/src/legal_api/models/filing.py
```diff
@@ -67,6 +67,7 @@ class Status(Enum):
     _status = db.Column('status', db.String(10), default='DRAFT')
     paper_only = db.Column('paper_only', db.Boolean, unique=False, default=False)
 
+    effective_date = db.Column('effective_date', db.DateTime(timezone=True), default=None)
```
thorwolpert: This is already in line 63, effective_date should not be defined twice.

legal-api/src/legal_api/models/filing.py
```diff
@@ -291,6 +294,13 @@ def get_completed_filings_for_colin():
                    Filing._status == Filing.Status.COMPLETED.value).all()
         return filings
 
+    @staticmethod
```
thorwolpert: consider setting some limit on this, as the next guy might just call it withA status like COMPLETED and get a few hundred thousand rows

queue_services/entity-filer/src/entity_filer/filing_processors/annual_report.py
```diff
@@ -24,7 +24,7 @@ def process(business: Business, filing: Dict):
     """Render the annual_report onto the business model objects."""
     agm_date = filing['annualReport'].get('annualGeneralMeetingDate')
     ar_date = filing['annualReport'].get('annualReportDate')
-    if agm_date and validations.annual_report.requires_agm(business):
+    if agm_date and validations.annual_report.requires_agm(business):  # pylint: disable=no-member;
```
thorwolpert: why is this lint disabled here?

jobs/future-effective-filings/file_future_effective.py
```diff
@@ -0,0 +1,242 @@
...
+
+def setup_jwt_manager(app, jwt_manager):
+    """Use flask app to configure the JWTManager to work for a particular Realm."""
+    def get_roles(a_dict):
+        return a_dict['realm_access']['roles']  # pragma: no cover
+    app.config['JWT_ROLE_CALLBACK'] = get_roles
+
+    jwt_manager.init_app(app)
+
+def register_shellcontext(app):
```
thorwolpert: this isn't needed as now shell debug context is provided

jobs/future-effective-filings/file_future_effective.py
```diff
@@ -0,0 +1,242 @@
...
+from dotenv import find_dotenv, load_dotenv
+import config
+
+
+from utils.logging import setup_logging
+
+setup_logging(os.path.join(os.path.abspath(os.path.dirname(__file__)), \
+    'logging.conf'))  # important to do this first
+
+class QueueHelper():
```
thorwolpert: consider just using the queue services common for this.

lear-db/Dockerfile
peter-freshworks: in order to shell into a postgres container you need to have sudo installed on it, it just got annoying to do it every time so we put it into the build

legal-api/src/legal_api/models/filing.py
peter-freshworks: this is true, but it's also not necessarily incorrect/undesired behaviour...I'll chew on this

queue_services/entity-filer/src/entity_filer/filing_processors/annual_report.py
peter-freshworks: because pylint was throwing up an error for requires_agm not being a member (which it is)

jobs/future-effective-filings/utils/__init__.py
peter-freshworks: you are right! They don't get used. The structure of the folder was a copy over from the other job(s) and this was  dangling code

queue_services/entity-filer/src/entity_filer/filing_processors/annual_report.py
thorwolpert: We should check how you're running lint, passes here. 
Our convention is to try and leave a comment as to why we're overriding / disabling something for future reviewers.
I think you can either remove this, or place a comment as to why it was added.

jobs/future-effective-filings/openshift/cron-file-future-effective.yml
peter-freshworks: These are yaml for the other jobs?

legal-api/src/legal_api/models/filing.py
peter-freshworks: good catch, I had this defined before the filer split went in, and when I was cherry-picking commits one of them must've had this

jobs/future-effective-filings/openshift/cron-file-future-effective.yml
thorwolpert: Ugh, seems some got through as YAML. Leave it for now then, and I'll change them in another ticket.

PR #490, opened 2019-11-07
LaunchDarkly Integration for coops ui

leksmall:
*Issue #:* /bcgov/entity#1482,  /bcgov/entity#1602

*Description of changes:*
LaunchDarkly Integration changes
Removed CDN reference for MDI 

Todo : Replace the LD Client Key with the one applicable for coops ui


leksmall:
> I think the code looks fine, but how do you use it?
> 
> Also, did you create a note on what config map changes to make in Openshift?



> I think the code looks fine, but how do you use it?
> I will give a walk through on Monday

> Also, did you create a note on what config map changes to make in Openshift?
Yes. But I am not sure of the value to be used. The one in the local config now is the trial account client id. That needs to be replaced  with the one for coops

PR #498, opened 2019-11-08
Change Of Address Filing 

cameron-freshworks:
*Issue #:* /bcgov/entity#1399 & /bcgov/entity#1393 & /bcgov/entity#1835

*Description of changes:*

* Implementation of a new officeAddresses component used in standaloneAddress filing.
* New component has an updated UI and is functional with both entity types.
* Refactor of StandAloneAddress filing.
* Refactor of Annual Report Address Component.
* Updated unit tests for affected components.
* Unit test for the officeAddresses component.


coops-ui/src/views/AnnualReport.vue
```diff
@@ -286,13 +287,13 @@ export default {
       saveWarnings: [],
 
       // EntityTypes Enum
-      EntityTypes
+      EntityTypes: EntityTypes
```
severinbeauvais: This change is not needed in ES6. See http://es6-features.org/#PropertyShorthand.

coops-ui/src/views/AnnualReport.vue
```diff
@@ -452,17 +453,17 @@ export default {
             }
 
             // load Change of Address fields
-            const changeOfAddress = filing.changeOfAddress
-            if (changeOfAddress) {
-              if (changeOfAddress.deliveryAddress && changeOfAddress.mailingAddress) {
-                this.addresses = {
-                  deliveryAddress: changeOfAddress.deliveryAddress,
-                  mailingAddress: changeOfAddress.mailingAddress
+            const changeOfAddress = filing.changeOfAddress.offices
```
severinbeauvais: 1. I think you need a check that filing.changeOfAddress is not null before you dereference office.
2. You should change the const name to offices.
3. Isn't code needed below to handle recordsOffice?

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -35,15 +35,28 @@
...
           <header>
-            <h1 id="filing-header">Address Change</h1>
+            <h1 id="filing-header">Change of Office Addresses</h1>
+            <p>Please change your Registered Office Address
+              <span v-if="entityFilter(EntityTypes.BCorp)">and Record Address</span>
+            </p>
+            <v-alert
+              v-if="entityFilter(EntityTypes.BCorp)"
+              type="info"
+              :value="true"
```
severinbeauvais: Since value doesn't need to be reactive (ie, based on a property that changes), try this (ie, no colon):
```
value="true"
```

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -35,15 +35,28 @@
...
-          <!-- Registered Office Addresses -->
+          <!-- Office Addresses -->
           <section>
-            <RegisteredOfficeAddress
+            <OfficeAddresses
               :changeButtonDisabled="false"
-              :legalEntityNumber="entityIncNo"
               :addresses.sync="addresses"
+              :registeredAddress.sync="registeredAddress"
+              :recordsAddress.sync="recordsAddress"
```
severinbeauvais: This is confusing. If I understand correctly, for Coop I use 'addresses' but for BCorp I use 'registeredAddress' and 'recordsAddress'??? Um, can we either use a single set of properties, or at least make this much clearer (eg, by renaming the props)?

coops-ui/src/constants/index.ts
```diff
@@ -1,4 +1,5 @@
 import * as DirectorConst from './director-constants'
+export * from './address-constants'
```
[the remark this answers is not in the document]
cameron-freshworks: 100% Agree, i will refactor this and import/export properly. I noticed this and was looking for a good place to do the refactor, thanks for commenting on it!

coops-ui/src/schemas/index.ts
```diff
@@ -0,0 +1 @@
+export * from './address-schema'
```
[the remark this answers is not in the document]
cameron-freshworks: Works for me, this was an inconsistency on my part. Will correct!

coops-ui/src/schemas/address-schema.ts
```diff
@@ -0,0 +1,33 @@
+import { required, maxLength } from 'vuelidate/lib/validators'
```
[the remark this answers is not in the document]
cameron-freshworks: 100% agree, i have also been a victim of this confusion when i was digging into the validation. I would like to explore the option of extracting it and i can refactor it a little further to bring a little more clarity and consistency. I will apply the changes to the directors schema as well, if it doesn't bring any value to us though, i will revert to the schemas in the components themselves. We can decide upon re-review!

coops-ui/src/views/AnnualReport.vue
```diff
@@ -200,7 +201,7 @@
 import axios from '@/axios-auth'
 import AGMDate from '@/components/AnnualReport/AGMDate.vue'
 import ARDate from '@/components/AnnualReport/BCorp/ARDate.vue'
-import RegisteredOfficeAddress from '@/components/AnnualReport/RegisteredOfficeAddress.vue'
+import { OfficeAddresses } from '@/components/Common'
```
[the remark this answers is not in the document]
cameron-freshworks: 100% agree, i would love to do a little refactoring to move the common components to the common directory. 

I think it makes sense that the common directory would be larger and the annual report a little smaller, as the only components that are truly specific to annual report thus far are AGM date and ARDate, imo. 

This refactor is a little bit of work, not huge, but due to the scope of this ticket could we create a ticket specific for this refactor and pursue it behind this one? 

oops, that is just a text blunder on my part, will rename appropriately, thanks!

coops-ui/src/views/AnnualReport.vue
cameron-freshworks: Good catch, thanks!

coops-ui/src/views/AnnualReport.vue
cameron-freshworks: 1. Could certainly add this check!
2. Sure can do, was just following the pattern defined earlier when directors are defined
line 425 `const changeOfDirectors = filing.changeOfDirectors`. Would you prefer that to be refactored as well to `directors`?
3. Bcorps cannot file office address changes through the annual report flow.

severinbeauvais:
Please reduce vertical whitespace in AddressListSm.vue (to look similar to director addresses).

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
cameron-freshworks: oops, this may not even need to be there. I believe i just left this in while testing.

severinbeauvais:
Please add missing class so that v-alert at top of COA page (and presumably AR page???) have white background -- same as v-alert at top of COD page.

severinbeauvais:
In App.vue:storeAddresses(), the code to omit the addressType property does not work (2 instances of this).

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
cameron-freshworks: Both entities use Addresses as it is used to retrieve filings, from a draft for instance. 
The registered and records props are what is retrieved from the state providing our baseline addresses, in case a user wants to reset it default.
I could definitely try and make this a little clearer, would some comments help perhaps?

coops-ui/src/views/AnnualReport.vue
cameron-freshworks: 1. Could certainly add this check!
2. Sure can do, was just following the pattern defined earlier when directors are defined
line 425 `const changeOfDirectors = filing.changeOfDirectors`. Would you prefer that to be refactored as well to `directors`?
3. Bcorps cannot file office address changes through the annual report flow.

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
cameron-freshworks: oops, this may not even need to be there. I believe i just left this in while testing.

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
cameron-freshworks: Both entities use Addresses as it is used to retrieve filings, from a draft for instance. 
The registered and records props are what is retrieved from the state providing our baseline addresses, in case a user wants to reset it default.
I could definitely try and make this a little clearer, would some comments help perhaps?

coops-ui/src/views/AnnualReport.vue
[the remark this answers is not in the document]
severinbeauvais: OK, so if you don't want to move the other components to common/ then please create a new ticket for that. Thanks!

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
severinbeauvais: Actually, you don't need value at all, since by default it's true. And while you're there, could you please fix this in StandaloneDirectorsFiling.vue as well? Oops and thanks.

coops-ui/src/components/Common/OfficeAddresses.vue
```diff
@@ -0,0 +1,813 @@
...
+     */
+    @Prop({ default: null })
+    private recordsAddress: BaseAddressObjIF
+
+    // The two addresses that come from the store. These are used to reset the address.
+    private deliveryAddressOriginal = {} as AddressIF
+    private mailingAddressOriginal = {} as AddressIF
+
+    // The two addresses that come from the store. These are used to reset the address.
+    private recDeliveryAddressOriginal = {} as AddressIF
```
severinbeauvais: Any idea why we need two separate address properties to reset the address? If there's a reason then please add comments here to discern these. Otherwise I'm confused. If there's no reason then please, please, please simplify this code. Thanks either way!

coops-ui/src/components/Common/OfficeAddresses.vue
```diff
@@ -0,0 +1,813 @@
...
+    private mailingAddress = {} as AddressIF
+
+    // The two addresses that are the current state of the BaseAddress components.
+    private recDeliveryAddress = {} as AddressIF
+    private recMailingAddress = {} as AddressIF
+
+    // The two addresses for Registered Office where the above are stored prior to an edit. These allow a cancel to
+    // the address prior to edit, which if there was a prior edit will not be the data that originally
+    // came from the store.
+    private deliveryAddressTemp = {} as AddressIF
```
severinbeauvais: Ah, so we have THREE separate addresses for reset purposes. Sigh...

Does the comment above make sense? I mean, if you cancel an edit, shouldn't we fall back to the address from the store? (I'm honestly not sure what the answer is.)

coops-ui/src/components/Common/OfficeAddresses.vue
```diff
@@ -0,0 +1,813 @@
...
+    /**
+     * Compares two address objects while omitting specified properties from the comparison.
+     *
+     * @param addressA The first address to compare
+     * @param addressB The second address to compare
+     * @param prop The property to omit during the comparison
+     *
+     * @return boolean A boolean indicating a match of addresses
+     */
+    private isSameWithoutProp (addressA: AddressIF, addressB: AddressIF, ...prop: Array<string>): boolean {
```
severinbeauvais: There's a similar function in a mixin. Can we please have a single method for this?

coops-ui/src/components/Common/OfficeAddresses.vue
```diff
@@ -0,0 +1,813 @@
...
+     * Assign the Registered address objects actions and omit the addressType
+     *
+     * @param addressBase The base address object to be parsed and assigned to its Office Address class property
+     */
+    private assignRegisteredAddresses (addressBase: BaseAddressObjIF): void {
+    // Assign Delivery Address
+      const deliveryAddress = addressBase.deliveryAddress
+      if (deliveryAddress) {
+        deliveryAddress.actions = []
+        this.deliveryAddressOriginal = { ...this.omitProp(deliveryAddress, ['addressType']) as AddressIF }
```
severinbeauvais: Should prop be omitted earlier, like wherever addressBase is fetched?

Also in assignRecordAddresses() below.

severinbeauvais:
Do the unit tests pass?

coops-ui/src/views/AnnualReport.vue
severinbeauvais: Re: 2.

The pattern is that the const is a shortcut to the property, so `const changeOfDirectors = filing.changeOfDirectors` makes sense. For this reason, on line 456 above, I think it should be `const offices = filing.changeOfAddress.offices`.

cameron-freshworks:
> Do the unit tests pass?

Yes sir, Linting too ;)

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
severinbeauvais: I'm sitll confused, so yes please, more code comments!

coops-ui/tests/unit/AnnualReport.spec.ts
```diff
@@ -691,8 +690,26 @@ describe('AnnualReport - Part 3 - Submitting', () => {
 
     // stub address data
     vm.addresses = {
-      deliveryAddress: {},
-      mailingAddress: {}
+      'registeredOffice': {
+        'deliveryAddress': {
+          'streetAddress': 'delivery street address',
```
severinbeauvais: Do the actual data properties need to exist for these tests? I mean, the tests worked before with only the stub addresses right?

Also below.

coops-ui/tests/unit/OfficeAddresses.spec.ts
```diff
@@ -0,0 +1,411 @@
...
+        addressRegion: 'mailRegionDraft',
+        deliveryInstructions: 'mailInstructionsDraft',
+        postalCode: 'mailPostalDraft',
+        streetAddress: 'mailStreetDraft',
+        streetAddressAdditional: 'mailStreetAddDraft'
+      }
+    }
+  }
+
+  it('loads the current office addresses properly', async done => {
```
severinbeauvais: Async is not needed since nothing in here awaits anything. I think this applies to all tests in this file.

coops-ui/tests/unit/OfficeAddresses.spec.ts
```diff
@@ -0,0 +1,411 @@
...
+    const mailingAddress = vm.registeredAddress.mailingAddress
+    expect(mailingAddress['streetAddress']).toEqual('mailStreet')
+    expect(mailingAddress['streetAddressAdditional']).toEqual('mailStreetAdd')
+    expect(mailingAddress['addressCity']).toEqual('mailCity')
+    expect(mailingAddress['addressRegion']).toEqual('mailRegion')
+    expect(mailingAddress['postalCode']).toEqual('mailPostal')
+    expect(mailingAddress['addressCountry']).toEqual('mailCountry')
+    expect(mailingAddress['deliveryInstructions']).toEqual('mailInstructions')
+
+    Vue.nextTick(() => {
```
severinbeauvais: Is there a need to wait for the next tick here, or can you simply call done() ?

Actually, since you're not waiting for anything to complete before you tell this test it's done, you can remove done() and then just pass () into the arrow function on line 71.

Also below.

coops-ui/tests/unit/StandaloneOfficeAddressFiling.spec.ts
```diff
@@ -520,14 +617,8 @@ describe('Standalone Office Address Filing - Part 3 - Submitting', () => {
 
     // click the File & Pay button
     button.trigger('click')
```
severinbeauvais: It works now? Hooray!

coops-ui/tests/unit/OfficeAddresses.spec.ts
```diff
@@ -0,0 +1,411 @@
+import Vue from 'vue'
```
severinbeauvais: Need some extra tests in here to check when address are "same as ..." (different flavours of this).

Should also add check that "same as" is NOT displayed in other cases.

coops-ui/src/constants/index.ts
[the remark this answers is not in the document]
cameron-freshworks: So i refactored these a little different than mixins. To avoid having to refer to the named `as` export where we are using these constants i did an `export * `. It cleans up the code a little and also simplified the constants file. Refer to the changes on SummaryDirectors to see an example.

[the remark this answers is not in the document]
cameron-freshworks: I'm going to do the same Refactor to the interfaces, as it makes the code much cleaner when implemented. I should have done it this way to start :)

coops-ui/src/schemas/index.ts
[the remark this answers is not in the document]
cameron-freshworks: Same thing here as above, since the mixins are classes they export cleaner, but i think we should export our constants, enums, interfaces and schemas etc as a batch. So when we use all of the above we don't have to refer to them as `someExportName.constant` or `someExportName.schema`. This is just my opinion on it, but we can discuss further!

coops-ui/src/views/AnnualReport.vue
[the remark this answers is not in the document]
cameron-freshworks: Created 👍

coops-ui/src/views/AnnualReport.vue
cameron-freshworks: Cool, done and done 👍

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
cameron-freshworks: Sounds good, thanks!

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
cameron-freshworks: Expanded comments inside of the `OfficeAddresses` component where the props are declared :)

coops-ui/src/components/Common/OfficeAddresses.vue
cameron-freshworks: Make perfect sense to remove these, i think the original thought process to prevent having to do another api call ( which was also redundant and was also removed ). Great comment!

coops-ui/src/components/Common/OfficeAddresses.vue
cameron-freshworks: Great point, i did a significant refactor. I removed the address mixin, and created a method in the common mixin that can compare any objects against each other, whilst having the option to pass in a property to omit during the comparison.    I did this because the address comparison was only specific to addresses by name only.

coops-ui/src/components/Common/OfficeAddresses.vue
cameron-freshworks: Yes!  I implemented the omitProps at the app level. EntityType will no longer be based into the children components.

coops-ui/tests/unit/AnnualReport.spec.ts
cameron-freshworks: They did indeed, they don't hurt the quality of the test but they certainly aren't necessary. I will clean that up!

coops-ui/tests/unit/OfficeAddresses.spec.ts
cameron-freshworks: Correct indeed, there is nothing being awaited. Removing!

coops-ui/tests/unit/OfficeAddresses.spec.ts
cameron-freshworks: 100% right. I saw this implementation of nextTick and callbacks used consistently throughout the tests and just wanted to be congruent with the code as it was. But i whole heartedly agree, let's simplify as much as we can and remove it!

coops-ui/tests/unit/OfficeAddresses.spec.ts
cameron-freshworks: Will do this!

coops-ui/src/constants/index.ts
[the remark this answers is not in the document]
severinbeauvais: Looks good.

coops-ui/src/schemas/index.ts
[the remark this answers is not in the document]
severinbeauvais: OK, I see only the mixins use a different format now. That's OK I guess.

coops-ui/src/schemas/address-schema.ts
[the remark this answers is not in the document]
severinbeauvais: @cameron-freshworks We still have the issue of 2 separate schemas (one in schemas/ and the other in `Directors.vue`) with no way to distinguish between them. How about renaming the subject schema to "officeAddressSchema" ?

coops-ui/src/mixins/common-mixin.ts
```diff
@@ -1,18 +1,47 @@
...
+  /**
+   * Removes the specified properties from nested objects.
+   *
+   * @param baseObj The base object.
+   * @param keys The nested object keys which to omit properties from.
+   * @param prop The properties to be removed.
+   */
+  omitProps (baseObj: Object, keys: Array<string>, prop: Array<string>): object {
+    let parsedObj = {}
+    Object.keys(baseObj).forEach(keys => {
```
severinbeauvais: For future, this looks like a textbook example that `map` could solve :)

coops-ui/src/mixins/common-mixin.ts
```diff
@@ -1,18 +1,47 @@
...
+   * Compares two objects while omitting specified properties from the comparison.
+   *
+   * @param addressA The first object to compare
+   * @param addressB The second object to compare
+   * @param prop The property to omit during the comparison
+   *
+   * @return boolean A boolean indicating a match of objects
+   */
+  isSame (objA: {}, objB: {}, prop: string = null): boolean {
+    return isEqual({ ...this.omitProp(objA, [prop]) }, { ...this.omitProp(objB, [prop]) })
```
severinbeauvais: Cool!

coops-ui/src/views/AnnualReport.vue
```diff
@@ -452,12 +453,14 @@ export default {
             }
 
             // load Change of Address fields
-            const changeOfAddress = filing.changeOfAddress
-            if (changeOfAddress) {
-              if (changeOfAddress.deliveryAddress && changeOfAddress.mailingAddress) {
+            if (filing.changeOfAddress) {
+              const offices = filing.changeOfAddress.offices
+              if (offices.registeredOffice) {
```
severinbeauvais: You need a check that "offices" isn't null before you dereference it.
`if (offices && offices.registeredOffice) {`

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
severinbeauvais: Thanks.

coops-ui/src/components/Common/OfficeAddresses.vue
severinbeauvais: Out of scope.

coops-ui/src/components/Common/OfficeAddresses.vue
severinbeauvais: Thanks for the cleanup.

coops-ui/src/components/Common/OfficeAddresses.vue
severinbeauvais: Yay, thanks! (Original) business address is only fetched in 1 place now!

@cameron-freshworks Can you confirm `addressType` is not returned anywhere else (eg, director data)? I don't see it in Dev but you may have different BE code than me.

coops-ui/src/components/Common/OfficeAddresses.vue
severinbeauvais: Yes, nicely done. I like it!

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
severinbeauvais: Fixed, thank you.

coops-ui/tests/unit/OfficeAddresses.spec.ts
severinbeauvais: Looks good. Keep the tests simple, yeah!

coops-ui/tests/unit/OfficeAddresses.spec.ts
severinbeauvais: Awesome, thanks!

coops-ui/src/mixins/common-mixin.ts
cameron-freshworks: Very cool, i'll look into that, thanks!

coops-ui/src/views/AnnualReport.vue
cameron-freshworks: 100%, i missed the point of that one, my bad. Will correct.

coops-ui/src/schemas/address-schema.ts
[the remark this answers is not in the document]
cameron-freshworks: Absolutely, can do.. OR.... would you prefer i apply this schema to the `Directors.vue` as well?  ( I guess the question is do the `isBc()` and `isCanada` methods apply to the office address business rules as well? )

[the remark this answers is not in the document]
severinbeauvais: The schemas are different.
As discussed, just rename this schema.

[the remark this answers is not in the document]
cameron-freshworks: Changed!

coops-ui/src/components/Common/OfficeAddresses.vue
cameron-freshworks: I cannot find `addressType` locally past the point it's filtered out ( `app.vue`).

coops-ui/src/views/AnnualReport.vue
cameron-freshworks: Changed!

severinbeauvais: This fix needs to be made in AnnualReport.vue on line 458 also.

coops-ui/src/components/Common/OfficeAddresses.vue
severinbeauvais: Perfect, thanks.

coops-ui/src/views/AnnualReport.vue
severinbeauvais: Oops, you don't need both lines 456 and 458 😛

cameron-freshworks:
> I closed most conversations but there are a few still open. Please have a look.

Think i got them all, thanks!

coops-ui/src/views/AnnualReport.vue
severinbeauvais: Fixed.

PR #501, opened 2019-11-12
Bugfix - Filings completed after 4pm show as the next day, Report has wrong Datetime

leksmall:
*Issue #:* /bcgov/entity#1818, /bcgov/entity#1405

*Description of changes:*
Changes to use local timezone datetime for filing_date and completion_date rather than utc 
Changes to correct the date time in reports
Changes to pass time to receipt

### DIFF
--- a/coops-ui/src/components/Dashboard/FilingHistoryList.vue
+++ b/coops-ui/src/components/Dashboard/FilingHistoryList.vue
@@ -131,7 +131,8 @@ export default {
       for (let i = 0; i < this.filings.length; i++) {
         const filing = this.filings[i].filing
         if (filing && filing.header) {
-          if (filing.header.date < '2019-03-08' || filing.header.availableOnPaperOnly) {
+          let filingDate = filing.header.date.slice(0, 10)
+          if (filingDate < '2019-03-08' || filing.header.availableOnPaperOnly) {
             this.loadPaperFiling(filing)
           } else {
             switch (filing.header.name) {
--- a/coops-ui/src/components/Dashboard/FilingHistoryList.vue
+++ b/coops-ui/src/components/Dashboard/FilingHistoryList.vue
@@ -89,6 +89,7 @@
 
 <script lang="ts">
 import ExternalMixin from '@/mixins/external-mixin'
+import DateMixin from '@/mixins/date-mixin'
 import axios from '@/axios-auth'
 import { mapState } from 'vuex'
 import DownloadErrorDialog from '@/components/Dashboard/DownloadErrorDialog.vue'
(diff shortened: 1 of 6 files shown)


legal-api/src/legal_api/models/filing.py
```diff
@@ -58,10 +58,10 @@ class Status(Enum):
 
     id = db.Column(db.Integer, primary_key=True)
     _completion_date = db.Column('completion_date', db.DateTime(timezone=True))
-    _filing_date = db.Column('filing_date', db.DateTime(timezone=True), default=datetime.utcnow)
+    _filing_date = db.Column('filing_date', db.DateTime(timezone=True), default=datetime.now)
```
severinbeauvais: Are there unit tests for these?

coops-ui/src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -304,7 +319,7 @@ export default {
       const url = filing.paymentToken + '/receipts'
       const data = {
         corpName: this.entityName,
-        filingDateTime: filing.filingDate, // TODO: format as needed
+        filingDateTime: this.convertISOToCommonTime(filing.filingDateTime), // TODO: format as needed
```
severinbeauvais: Is this now formatted as needed? Just wondering if the comment can be deleted or if more formatting is required (and should it be done here/now)?

coops-ui/src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -294,6 +298,17 @@ export default {
       })
     },
 
+    convertISOToCommonTime (isoTime) {
```
severinbeauvais: We have not imported the 'moment' package yet, but if we start doing a lot with dates/times/formatting then we should consider it. I think we're not yet at that point, but just keep this in mind for later.

legal-api/src/legal_api/models/filing.py
severinbeauvais: Lekshmi answered that, yes, there is a unit test that check for the existence of of this property, but it does not validate the time (which is kind of hard when you use "now", since it's always changing). So I'm OK with the status quo.

coops-ui/src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -174,18 +174,20 @@ export default {
     loadAnnualReport (filing) {
       if (filing.annualReport) {
         const date = filing.annualReport.annualGeneralMeetingDate
+        const filingDate = filing.header.date.slice(0, 10)
```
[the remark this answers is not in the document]
leksmall: I have added a temporary solution to slice the datetime to get the date.

PR #502, opened 2019-11-13
remove error if director list is empty

peter-freshworks:
*Issue #:* /bcgov/entity1839
*Description of changes:*
Remove 404 being thrown if director list is empty. Empty list is valid

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_directors.py
+++ b/legal-api/src/legal_api/resources/business/business_directors.py
@@ -55,8 +55,6 @@ class DirectorResource(Resource):
                 del director_json['mailingAddress']
             res.append(director_json)
 
-        if not res:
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/resources/business/business_directors.py
```diff
@@ -55,8 +55,6 @@ def get(identifier, director_id=None):
                 del director_json['mailingAddress']
             res.append(director_json)
 
-        if not res:
```
severinbeauvais: Are there unit tests that need to be updated (or written)?

peter-freshworks: you know what, I asked myself the same question but didn't see any specific coverage for this and pytest ran clean...but on second look I do see a test case for this. Updating now... (I ran pytest without the make command and it didn't run the suite that would have failed)

PR #505, opened 2019-11-14
default checkmark to unchecked for same-address on directors

peter-freshworks:
*Issue #:* /bcgov/entity#1732

*Description of changes:*


peter-freshworks:
> 1. Please update the PR issue reference to "/[bcgov/entity#1732](https://github.com/bcgov/entity/issues/1732)" so it links correctly.
> 2. Is there a unit test for this initial state? Could you please add one?

After much tinkering and discussion, we have decided against any additional test cases. As for the bug itself, I've been told this is confirmed to be incorrect behaviour so the defect is valid.

PR #511, opened 2019-11-18
add script to fix coops after offices deployed

peter-freshworks:
*Issue #:* /bcgov/entity#1876
*Description of changes:*
Once the story branch is merged, and deployed, coops will need to be updated to have an office association. New coops will need one created on initial load from colin. 

### DIFF
--- a/lear-db/test_data/add_offices.py
+++ b/lear-db/test_data/add_offices.py
@@ -1,0 +1,46 @@
+# Copyright © 2019 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


lear-db/test_data/add_offices.py
```diff
@@ -0,0 +1,46 @@
...
+def add_business_office_records():
+    """Iterate all business entries and associate a records office if needed."""
+
+    businesses = db.session.query(Business).all() #pylint: disable=no-member; needed by SQLAlchemy
+
+    for business in businesses:
+        if not any(business.offices):
+            office = Office()
+            office.office_type = 'registeredOffice'
+            office.addresses = [business.mailing_address.one_or_none(),
```
kialj876: After migrating the won't all the business mailing/delivery addresses be orphaned? 
I'm a bit out of the loop as to what your plan is for adding in the office, but I assume you're replacing the mailing/delivery addresses for businesses with the office right? In that case, business.mailing_address wouldn't exist. Or if you ran this before migrating then office wouldn't exist. We can chat tomorrow!

peter-freshworks: yes let's chat, the business_mailing address would still exist, but now it is a property (so as to retain code that already uses mailing_address etc.) You can see the behaviour in models/business.py

PR #515, opened 2019-11-19
check not only for a mailing address, but also that it is not empty

peter-freshworks:
*Issue #:* /bcgov/entity#1913

*Description of changes:*
Add empty object check for (optional) mailing address in change of director

### DIFF
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/change_of_directors.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/change_of_directors.py
@@ -37,7 +37,7 @@ def process(business: Business, filing: Dict):
                                        cessation_date=new_director.get('cessationDate'),
                                        delivery_address=address)
 
-            if 'mailingAddress' in new_director.keys():
+            if 'mailingAddress' in new_director and len(new_director['mailingAddress']):
                 mailing_address = create_address(new_director['mailingAddress'], Address.MAILING)
                 director_to_add.mailing_address = mailing_address
 
--- a/legal-api/src/legal_api/services/filings/validations/change_of_directors.py
+++ b/legal-api/src/legal_api/services/filings/validations/change_of_directors.py
@@ -34,7 +34,7 @@ def validate(business: Business, cod: Dict) -> Error:
 
     for idx, director in enumerate(directors):
         for address_type in Address.JSON_ADDRESS_TYPES:
-            if address_type in director: 
+            if address_type in director:
                 try:
                     pycountry.countries.search_fuzzy(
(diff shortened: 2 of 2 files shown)


queue_services/entity-filer/src/entity_filer/filing_processors/change_of_directors.py
```diff
@@ -37,7 +37,7 @@ def process(business: Business, filing: Dict):
                                        cessation_date=new_director.get('cessationDate'),
                                        delivery_address=address)
 
-            if 'mailingAddress' in new_director.keys():
+            if 'mailingAddress' in new_director and len(new_director['mailingAddress']):
```
severinbeauvais: I'm concerned that you removed the `.keys()` bit, but this may be OK in Python, I dunno.

I recommend that Katie or Lekshmi review this.

peter-freshworks: I can't request reviewers but if you'd like to, sure! I can also speak to the keys() issue, as I'm the one who added that to begin with. This way is actually a bit cleaner and accomplishes the same thing so I swapped it

severinbeauvais: Reviewers added.
OK, cool. How did you test this?

peter-freshworks: I ran the bad data that flagged the bug through the filer to confirm the edge case doesn't cause the filer to crash and fail. There are already specific test cases for Entity Filer COD (for both BCORP and COOP)...looking more deeply however I have found another possible issue (we assume the filing has been validated by the time we get to the filer...a mailing address that does not pass "Address" validation shouldn't validate this far)...digging into this

severinbeauvais: If you confirm that your code is OK and the validation would work (when fixed by the other ticket) then I'm OK to resolve this conversation.

peter-freshworks: yeah I mean this ticket wouldn't have been raised if JSONSchema validation was being run on nested properties (I tested the offending filing in a JSONSchema validator and it failed - as expected). I don't know when it stopped working, but I did add some code for validating mailing address that should probably be there any way and is attached to this PR.

Or, basically - Yeah go ahead and resolve

severinbeauvais: LGTM but still need a BE dev to review this PR.

PR #516, opened 2019-11-20
#1914 Remove Coops Directors Mailing Address

cameron-freshworks:
*Issue #:* /bcgov/entity#1914

*Description of changes:*

* Fixed Directors filing submission where the coops directors would contain an empty mailing address.
* Updated unit testing.

### DIFF
--- a/coops-ui/src/components/AnnualReport/Directors.vue
+++ b/coops-ui/src/components/AnnualReport/Directors.vue
@@ -911,7 +911,7 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
         lastName: this.director.officer.lastName
       },
       deliveryAddress: { ...this.inProgressAddress },
-      ...(!this.entityFilter(EntityTypes.Coop) && { mailingAddress: { ...this.inProgressMailAddress } }),
+      ...(this.entityFilter(EntityTypes.BCorp) && { mailingAddress: { ...this.inProgressMailAddress } }),
       appointmentDate: this.asOfDate, // when implemented: this.director.appointmentDate,
       cessationDate: null // when implemented: this.director.cessationDate
     }
--- a/coops-ui/src/components/AnnualReport/Directors.vue
+++ b/coops-ui/src/components/AnnualReport/Directors.vue
@@ -773,7 +773,7 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
   }
 
   /**
-   * Function called internall and externally to fetch the list of directors.
+   * Function called internally and externally to fetch the list of directors.
    * TODO: change this to a prop?
    */
   public getDirectors (getOrigOnly: Boolean = false): void {
@@ -801,11 +801,6 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
               directors[i].officer.prevFirstName = directors[i].officer.firstName
               directors[i].officer.prevLastName = directors[i].officer.lastName
               directors[i].officer.prevMiddleInitial = directors[i].officer.middleInitial
-
-              // if this is a Coop, copy delivery address to mailing address - directors only have delivery addresses
-              if (this.entityFilter(EntityTypes.Coop)) {
-                directors[i].mailingAddress = directors[i].deliveryAddress
-              }
 
               // ensure there is complete address data including missing/blank fields
               directors[i].deliveryAddress = this.formatAddress(directors[i].deliveryAddress)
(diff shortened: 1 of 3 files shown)


coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -802,11 +802,6 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
               directors[i].officer.prevLastName = directors[i].officer.lastName
               directors[i].officer.prevMiddleInitial = directors[i].officer.middleInitial
 
-              // if this is a Coop, copy delivery address to mailing address - directors only have delivery addresses
```
severinbeauvais: I assume the BE doesn't expect a director mailing address anywhere (for a Coop). Please do as much testing as possible (and/or leave notes in the ticket for QA to verify this). Are the output PDFs OK?

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -917,7 +911,7 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
         lastName: this.director.officer.lastName
       },
       deliveryAddress: { ...this.inProgressAddress },
-      mailingAddress: { ...this.inProgressMailAddress },
+      ...(!this.entityFilter(EntityTypes.Coop) && { mailingAddress: { ...this.inProgressMailAddress } }),
```
severinbeauvais: Well this is a doozy.
Can it be refactored to something easier to understand, hopefully with if-then-else logic (or x ? y : z) ?
ATM, I can't guess what the result of the spread is when type isn't Coop.
Also I recommend the opposite logic for the if condition: "if BComp then" (otherwise you assume all other types except Coop have a mailing address). In this case I think it's more concise to specify for which types this applies.

Thanks. Sorry, this one broke my rule, "if it's hard to understand then it should be rewritten".

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -132,6 +132,10 @@ describe('Directors as a COOP', () => {
     // spot-check second address
     expect(vm.directors[1].deliveryAddress.streetAddressAdditional).toEqual('Kirkintiloch')
     expect(vm.directors[1].deliveryAddress.deliveryInstructions).toEqual('go to the back')
+
+    // Verify no mailing address fo Coop directors
```
severinbeauvais: Great!

coops-ui/src/components/AnnualReport/Directors.vue
cameron-freshworks: I thought this was the cleanest solution to adding a conditional property but i can certainly try and refactor ( expand the code ) if that makes it simpler to read. 

Do we know which entity is the exception here? Either way, we are assuming one way or the other until we have business information about the other entities? Why don't we ask business about this!

coops-ui/src/components/AnnualReport/Directors.vue
cameron-freshworks: Correct, it does not expect a mailing address. I can certainly try and verify the outputs. This code was removed because it was redundant. Originally a work around i imagine. Will verify all of the above.

cameron-freshworks: It appears the back end outputs a mailing address as well for Coops currently, though i'm not sure if the back end tickets for the outputs have been done for COD. I will inform QA. Changes to the BE outputs are outside the scope of this ticket i think.

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -917,7 +911,7 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
         lastName: this.director.officer.lastName
       },
       deliveryAddress: { ...this.inProgressAddress },
-      mailingAddress: { ...this.inProgressMailAddress },
+      ...(this.entityFilter(EntityTypes.BCorp) && { mailingAddress: { ...this.inProgressMailAddress } }),
       appointmentDate: this.asOfDate, // when implemented: this.director.appointmentDate,
```
cameron-freshworks: @severinbeauvais 
I updated the condition to be specific to BCORPs as it is sort of arbitrary until we have more information and if you prefer it his, no problem at all! 

I disagree that this method is too complicated however. Using rest and spread operators are quite common now and it seems quite useful in this case. A ternary operator will not actually omit the mailingAddress property whereas wrapping the entire `newDirector` obj into an if else statement is writing duplicate code in my opinion. Is this perfect or the best way, i'm not sure,  so I'm certainly open to ideas though of what you think would be simpler than this.

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: OK, thanks for that.

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: Yes please.

You're right, but today what we know for sure is that Mailing Address applies only to BComps so IMHO I would rather the code "opt in" for that case instead of opting out Coops. I could be wrong in the long term, but positive logic is always easier to understand :)

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: I had a quick poll here and all agreed that this may work but it's difficult to understand. (Sorry.) The good news is we are all learning something from your fancy code :)

An option is to conditionally add this property to the newDirector object. Thoughts?

coops-ui/src/components/AnnualReport/Directors.vue
cameron-freshworks: Makes sense, i definitely agree. We can adjust it down the road if need be when we have other entities information.

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: I had a quick poll here and all agreed that this may work but it's difficult to understand. (Sorry.) The good news is we are all learning something from your fancy code :)

An option is to conditionally add this property to the newDirector object. Thoughts?

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: Ok, you changed the logic.

coops-ui/src/components/AnnualReport/Directors.vue
cameron-freshworks: well of course it works 😆  Yes we can assign the property conditionally after creating it, if that is what the team prefers :)

severinbeauvais: I didn't check with the team but I prefer this -- thanks!

PR #517, opened 2019-11-20
1550 certify component

peter-freshworks:
*Issue #:* /bcgov/entity#1550

*Description of changes:*
Looking for feedback on changes made for 1550 as this would be the first Vue work I've done

### DIFF
--- a/coops-ui/src/components/Common/SummaryCertify.vue
+++ b/coops-ui/src/components/Common/SummaryCertify.vue
@@ -8,7 +8,7 @@
       <p class="certify-content">
         Note: It is an offence to make a false or misleading statement in
         respect of a material fact in a record submitted to the Corporate Registry for filing.
-        See section 200 of the Cooperatives Association Act.
+        See section {{sectionCode}} of the Cooperatives Association Act.
       </p>
     </div>
   </v-card>
@@ -16,7 +16,7 @@
 
 <script lang="ts">
 import { Component, Vue, Prop } from 'vue-property-decorator'
-
+import { EntityFilterMixin } from '@/mixins'
 @Component({})
 export default class SummaryCertify extends Vue {
   // Props passed into this component.
@@ -25,6 +25,12 @@ export default class SummaryCertify extends Vue {
 
   @Prop({ default: '' })
   private currentDate: string
+
+  @Prop({ default: '' })
+  private sectionCode: string
+
+  @Prop({ default: '' })
+  private sectionString: string
 
   /**
    * Computed value.
--- a/coops-ui/src/components/Common/SummaryCertify.vue
+++ b/coops-ui/src/components/Common/SummaryCertify.vue
@@ -2,13 +2,13 @@
   <v-card flat id="certify-container">
     <div class="container">
       <p class="certify-content">I, <b>{{ trimmedCertifiedBy || '[Legal Name]'}}</b>,
-      certify that I have relevant knowledge of the association and that I am authorized to make this filing.
+      certify that I have relevant knowledge of the {{entityDisplayName}} and that I am authorized to make this filing.
       </p>
       <p class="certify-content">{{currentDate}}</p>
       <p class="certify-content">
         Note: It is an offence to make a false or misleading statement in
         respect of a material fact in a record submitted to the Corporate Registry for filing.
-        See section {{sectionCode}} of the Cooperatives Association Act.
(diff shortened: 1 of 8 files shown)


coops-ui/src/components/Common/SummaryCertify.vue
```diff
@@ -2,21 +2,21 @@
...
+        See section {{sectionCode}} of the {{sectionString}}.
       </p>
     </div>
   </v-card>
 </template>
 
 <script lang="ts">
 import { Component, Vue, Prop } from 'vue-property-decorator'
-
+import { EntityFilterMixin } from '@/mixins'
```
severinbeauvais: Why do you need this?

coops-ui/src/mixins/entityFilter-mixin.ts
```diff
@@ -21,4 +21,24 @@ export default class EntityFilterMixin extends Vue {
...
+      case 'BC':
+        return 'Benefits Company'
+    }
+    return ''
+  }
+
+  entityLegalSection (): string {
+    switch (this.entityType) {
+      case 'CP':
+        return 'Cooperative Association act'
```
severinbeauvais: Typo. Should be "Act". Also below.

coops-ui/src/mixins/entityFilter-mixin.ts
```diff
@@ -21,4 +21,24 @@ export default class EntityFilterMixin extends Vue {
   entityFilter (entityType: string): boolean {
     return this.entityType === entityType
   }
+
+  entityTypeToDisplay (): string {
+    switch (this.entityType) {
+      case 'CP':
+        return 'Cooperative'
```
severinbeauvais: I don't see the requirement for this. Should the text fall back to "association" as before?

coops-ui/src/mixins/entityFilter-mixin.ts
```diff
@@ -21,4 +21,24 @@ export default class EntityFilterMixin extends Vue {
...
+    }
+    return ''
+  }
+
+  entityLegalSection (): string {
+    switch (this.entityType) {
+      case 'CP':
+        return 'Cooperative Association act'
+      case 'BC':
+        return 'Benefits Corporations act'
```
severinbeauvais: Typos. Should be "Business" and should be "Act".

coops-ui/src/views/AnnualReport.vue
```diff
@@ -503,6 +506,16 @@ export default {
       this.toggleFiling(modified ? 'add' : 'remove', 'OTFDR')
     },
 
+    getSectionCode () :string {
+      switch (this.entityType) {
+        case EntityTypes.Coop:
```
severinbeauvais: Check with Cam on his change to how constants are imported and used. If his PR goes in first then you will have to update this code. If your PR goes in first then he will have to update this code.

coops-ui/src/views/AnnualReport.vue
```diff
@@ -503,6 +506,16 @@ export default {
       this.toggleFiling(modified ? 'add' : 'remove', 'OTFDR')
     },
 
+    getSectionCode () :string {
```
severinbeauvais: I recommend you move this function (and the ones in the other 2 filings below) to the mixin as well. I realized they don't need to be shared by different components but it would keep all this logic and data in one place.

peter-freshworks: I'd love to, but the section code actually changes based on which path we are in (AR vs Standalone Director change etc.), unless I were to make a switch-case of switch-cases and pass the path in as a parameter

severinbeauvais: Could the functions be named differently?

peter-freshworks: Yeah I mean I don't see a clean way of doing this without a lookup of some form...I'd imagine we have to revisit this, since every entity potentially has different section codes (and I don't think we wants hundreds of lines of code just for this)

coops-ui/src/views/AnnualReport.vue
severinbeauvais: I think this comment is obsolete. I believe EntityTypes is still an enum (not a bunch of constants) so this code is still OK. Check with Cam anyway :)

katiemcgoff:
Closing per discussion.

PR #520, opened 2019-11-20
fixed blank space bug for directors names

severinbeauvais:
*Issue #:* /bcgov/entity#1706

*Description of changes:*
- added rules to validate extra whitespace in Directors component
- added unit tests to verify Directors input values
- added "trim" to some v-models
- removed extra whitespace in Certify component
- fixed boolean errors in Certify component
- updated Certify unit tests re: above
- cleaned up misc unit tests

### DIFF
--- a/coops-ui/src/components/AnnualReport/Certify.vue
+++ b/coops-ui/src/components/AnnualReport/Certify.vue
@@ -57,7 +57,7 @@ export default class Certify extends Vue {
    * Lifecycle callback to always give the parent a "valid" event for its property values.
    */
   private created (): void {
-    this.emitValid(this.trimmedCertifiedBy && this.isCertified)
+    this.emitValid(Boolean(this.trimmedCertifiedBy && this.isCertified))
   }
 
   /**
@@ -65,30 +65,31 @@ export default class Certify extends Vue {
    * @returns The trimmed "Certified By" string (may be '').
    */
   private get trimmedCertifiedBy (): string {
-    return this.certifiedBy && this.certifiedBy.trim()
+    // remove repeated inline whitespace, and leading/trailing whitespace
+    return this.certifiedBy && this.certifiedBy.replace(/\s+/g, ' ').trim()
   }
 
-  // Emit an update event.
+  // Emit an update event when input changes.
   @Emit('update:certifiedBy')
-  private emitCertifiedBy (val: string): string {
-    this.emitValid(val && val.trim() && this.isCertified)
+  private emitCertifiedBy (certifiedBy: string): string {
+    // remove repeated inline whitespace, and leading/trailing whitespace
+    certifiedBy = certifiedBy && certifiedBy.replace(/\s+/g, ' ').trim()
+    this.emitValid(Boolean(certifiedBy && this.isCertified))
 
-    return val
+    return certifiedBy
   }
 
-  // Emit an update event.
+  // Emit an update event when checkbox changes.
   @Emit('update:isCertified')
-  private emitIsCertified (val: boolean): boolean {
-    this.emitValid(this.trimmedCertifiedBy && !!val)
+  private emitIsCertified (isCertified: boolean): boolean {
+    this.emitValid(Boolean(this.trimmedCertifiedBy && isCertified))
 
-    return val
+    return isCertified
   }
 
   // Emit an event indicating whether or not the form is valid.
   @Emit('valid')
-  private emitValid (val: boolean): boolean {
-    return val
-  }
+  private emitValid (valid: boolean): void { }
 }
 </script>
 
--- a/coops-ui/src/components/AnnualReport/Directors.vue
+++ b/coops-ui/src/components/AnnualReport/Directors.vue
@@ -49,18 +49,23 @@
                   v-model="directorFormValid" lazy-validation
                 >
                   <div class="form__row three-column">
-                    <v-text-field filled class="item" label="First Name" id="new-director__first-name"
+                    <v-text-field filled class="item"
+                      label="First Name"
+                      id="new-director__first-name"
                       v-model="director.officer.firstName"
                       :rules="directorFirstNameRules"
-                      required
                     />
-                    <v-text-field filled label="Initial" class="item director-initial"
+                    <v-text-field filled class="item director-initial"
+                      label="Initial"
+                      id="new-director__middle-initial"
                       v-model="director.officer.middleInitial"
+                      :rules="directorMiddleInitialRules"
                     />
-                    <v-text-field filled class="item" label="Last Name"
+                    <v-text-field filled class="item"
+                      label="Last Name"
+                      id="new-director__last-name"
                       v-model="director.officer.lastName"
                       :rules="directorLastNameRules"
-                      required
                     />
                   </div>
 
@@ -103,7 +108,7 @@
                       <template v-slot:activator="{ on }">
                         <v-text-field box class="item" label="Appointment / Election Date"
                           id="new-director__appointment-date"
-                          v-model="director.appointmentDate"
+                          v-model.trim="director.appointmentDate"
                           hint="YYYY/MM/DD"
                           append-icon="event"
                           v-on="on"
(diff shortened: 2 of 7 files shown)


coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -57,38 +57,39 @@ export default class Certify extends Vue {
    * Lifecycle callback to always give the parent a "valid" event for its property values.
    */
   private created (): void {
-    this.emitValid(this.trimmedCertifiedBy && this.isCertified)
+    this.emitValid(Boolean(this.trimmedCertifiedBy && this.isCertified))
```
cameron-freshworks: I like it. Another option for type coercion to a `Boolean` would be `!!(this.trimmedCertifiedBy && this.isCertified)`. I think it's debatable which is more clear or concise though. Just a thought!

coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -57,38 +57,39 @@ export default class Certify extends Vue {
...
-  // Emit an update event.
+  // Emit an update event when checkbox changes.
   @Emit('update:isCertified')
-  private emitIsCertified (val: boolean): boolean {
-    this.emitValid(this.trimmedCertifiedBy && !!val)
+  private emitIsCertified (isCertified: boolean): boolean {
+    this.emitValid(Boolean(this.trimmedCertifiedBy && isCertified))
 
-    return val
+    return isCertified
```
cameron-freshworks: Is this method just returning the parameter passed in?

coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -57,38 +57,39 @@ export default class Certify extends Vue {
...
   }
 
   // Emit an event indicating whether or not the form is valid.
   @Emit('valid')
-  private emitValid (val: boolean): boolean {
-    return val
-  }
+  private emitValid (valid: boolean): void { }
 }
```
cameron-freshworks: I didn't know the `Emit` decorator could work like this. Learning something all the time. Big fan of decorators.

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -598,7 +609,21 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
   private get directorFirstNameRules (): Array<Function> {
     return [
       v => !!v || 'A first name is required',
-      v => !/\s/g.test(v) || 'Invalid spaces'
+      v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
+      v => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
+      v => !/\s\s/g.test(v) || 'Invalid word spacing' // multiple inline spaces
```
cameron-freshworks: Did we want to allow multiple names in the single name input fields?

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -598,7 +609,21 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
   private get directorFirstNameRules (): Array<Function> {
     return [
       v => !!v || 'A first name is required',
-      v => !/\s/g.test(v) || 'Invalid spaces'
+      v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
+      v => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
```
cameron-freshworks: The original space rule prevented leading and trailing spaces. Were these implemented to allow multiple names in a single name input field?

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -598,7 +609,21 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
...
+
+  /**
+   * Computed value.
+   * @returns The array of validations rules for a director's middle initial.
+   */
+  private get directorMiddleInitialRules (): Array<Function> {
+    return [
+      v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
+      v => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
+      v => !/\s\s/g.test(v) || 'Invalid word spacing' // multiple inline spaces
```
cameron-freshworks: Can we create one ruleMethod() and apply them to all of the fields where required, it appears that all of the error prompts are similar. ie..   

`private get directorNameRules (nameValidated: string): Array<Function> {
    return [
      v => !!v || `A ${nameValidated} name is required`,
      v => !/\s/g.test(v) || 'Invalid spaces'
      v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
      v => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
      v => !/\s\s/g.test(v) || 'Invalid word spacing' // multiple inline spaces
    ]
  }

coops-ui/src/components/AnnualReport/Certify.vue
severinbeauvais: Thanks.

coops-ui/src/components/AnnualReport/Certify.vue
severinbeauvais: Yes and yes.

There are 2 ways to implement an emitter method in TS. I used this way because the body needs to emit the valid flag as well as emit the parameter passed in. An example of the other way is on line 92, below.

coops-ui/src/components/AnnualReport/Certify.vue
severinbeauvais: Yes. I recently learned that you can still use `this.$emit()` in TypeScript but I think we should use decorators whenever possible.

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: The original space rule prevented a space anywhere in the field. These rules were implemented to allow compound names. For example, my previous PO's first name is "Ann Marie".

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: Yes. "Ann Marie" (or Bobbi Jo, or take your pick) are valid first names. This rule allows them but only if separated by 1 whitespace character.

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: Yes, I could write a property that is an array of the common rules, and spread it within each first/middle/last name getter.
I suggest that I (or another dev) do that the next time this code is touched or rules are added. OK?

coops-ui/src/components/AnnualReport/Directors.vue
cameron-freshworks: Makes sense!👍

coops-ui/src/components/AnnualReport/Directors.vue
cameron-freshworks: Same as above, good job 👍

coops-ui/src/components/AnnualReport/Directors.vue
cameron-freshworks: 100%, fine by me!

katiemcgoff: FYI - please check in postgres (prod, ideally) whether there are director names that break these rules (leading spaces, trailing spaces, multiple spaces between words). If so, we should either fix the data, or not be so strict. Otherwise users will get errors without actually changing their data.

severinbeauvais: I checked in Prod, and yes there are some director names with trailing / multiple inline spaces.

I entered /bcgov/entity#1968 to update the Prod DB and Legal Updater accordingly.

severinbeauvais: Not being strict is another possibility, ie, don't allow only spaces, but allow leading / trailing / multiple inline spaces. But I would prefer to see sanitized data.

severinbeauvais: @katiemcgoff Do we have a decision on whether to allow spaces in a field (but not only spaces)?

PR #523, opened 2019-11-21
scss cleanup

severinbeauvais:
*Issue #:* /bcgov/entity#1943

*Description of changes:*
- changed style import mechanism so that order is respected (per Jer and same as auth-web)
- updated container max-width (per Jer and same as auth-web)
- stopped importing theme.scss when not needed
- fixed some scss typos
- removed unwanted l/r padding (per Jer and same as auth-web)
- updated component containers
- fixed drop-down menu text size
- updated some styles same as auth-web
- imported lodash instead of components because it's imported by other dependencies anyways
- moved a package from prod dependency to dev dependency
- removed fade enter transition (looks better without it)
- updated layout to v-rows and v-cols (per Jer)
- updated app name
- deleted obsolete styles
- fixed some unit tests
- fixed affix width issue
- fixed some duplicate ids
- deleted obsolete code


coops-ui/package.json
```diff
@@ -17,14 +17,11 @@
     "country-list": "^2.2.0",
     "http-status-codes": "^1.3.2",
     "ld-vue": "^0.1.1",
-    "lodash.isempty": "^4.4.0",
-    "lodash.isequal": "^4.4.0",
-    "lodash.omit": "^4.4.0",
+    "lodash": "^4.17.15",
```
cameron-freshworks: This is great 👍

coops-ui/public/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <title>Cooperatives</title>
+    <title>Cooperatives Online</title>
```
cameron-freshworks: Totally inconsequential, but why this change?

coops-ui/src/App.vue
```diff
@@ -54,9 +54,6 @@ import AccountAuthorizationDialog from '@/components/Dashboard/AccountAuthorizat
 // Mixins
 import { DateMixin, CommonMixin } from '@/mixins'
 
-// Enums
-import { EntityTypes } from '@/enums'
```
cameron-freshworks: Did i import the enum but not use it?  This was probably a result of me importing the enum for use when hardcoding the `setEntityType` during testing. Good job on the clean up, thank you.

coops-ui/src/App.vue
```diff
@@ -358,7 +354,5 @@ export default {
 </script>
 
 <style lang="scss">
-@import "@/assets/styles/base.scss";
-@import "@/assets/styles/layout.scss";
-@import "@/assets/styles/overrides.scss";
+// @import '@/assets/styles/theme.scss';
```
cameron-freshworks: Love it

coops-ui/src/assets/styles/base.scss
```diff
@@ -288,6 +289,27 @@ ul.list {
...
 }
 
-.inline-div{
-  display: inline
+.genErr {
+  font-size: 0.9rem;
+}
+
+.white-background {
+  background-color: white !important;
```
cameron-freshworks: What are we overriding here, if this is the base theme style?

coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -94,7 +92,7 @@ export default class Certify extends Vue {
 </script>
 
 <style lang="scss" scoped>
-@import '@/assets/styles/theme.scss';
+// @import '@/assets/styles/theme.scss';
```
cameron-freshworks: Why is the theme commented out here?

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -77,7 +77,7 @@
                       @update:address="updateBaseAddress"
                     />
                   </div>
-                  <div class="form__row" v-if="entityFilter(EntityTypes.BCorp)">
+                  <div class="form__row" v-if="entityFilter(EntityTypes.BCORP)">
```
cameron-freshworks: Great change. There essentially constants with a wrapper so i definitely like 'all-caps'

coops-ui/src/components/Common/OfficeAddresses.vue
```diff
@@ -57,13 +57,11 @@
       </li>
 
       <!-- Registered Mailing Address -->
-      <li class="container">
+      <li class="address-list-container">
         <div class="meta-container">
           <label v-if="showAddressForm">Mailing Address</label>
-          <label v-else></label>
```
cameron-freshworks: Have you tested how the columns arrange without this?

coops-ui/src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -174,7 +174,8 @@ export default {
     },
 
     // Method to extract date from a local datetime string
-    formatDate (dateString) {
+    formatDate (dateString): string {
+      if (!dateString) return null // safety check
```
cameron-freshworks: I like what you're doing here! It might be a lot to add in with this PR because you would probably want to do all the functions but in our future refactor we could update this too `formateDate (dateString: string): string | null { .... }` to go heavy on the type'ing.

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -525,16 +525,22 @@ export default {
...
-  border-top-left-radius: 0;
-  border-bottom-left-radius: 0;
-  margin-left: 1px;
+.list-item__actions {
+  .v-btn.actions__more-actions__btn {
+    min-width: 38px !important;
+    width: 38px;
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
+    margin-left: 1px;
```
cameron-freshworks: I understand the px for font sizes but did we want to keep spacing consistent by using rem?

coops-ui/src/enums/entityTypes.ts
```diff
@@ -1,4 +1,4 @@
 export enum EntityTypes {
-  Coop = 'CP',
-  BCorp = 'BC'
+  COOP = 'CP',
+  BCORP = 'BC'
```
cameron-freshworks: Once again, i like it 👍

coops-ui/src/main.ts
```diff
@@ -4,23 +4,28 @@ import '@mdi/font/css/materialdesignicons.min.css' // ensure you are using css-l
...
+import configHelper from '@/utils/config-helper'
 import router from '@/router'
 import store from '@/store/store'
-import configHelper from '@/utils/config-helper'
-import '@/registerServiceWorker'
 import { withFlagProvider } from 'ld-vue'
+import '@/registerServiceWorker'
+import '@/assets/styles/base.scss'
+import '@/assets/styles/layout.scss'
+import '@/assets/styles/overrides.scss'
```
cameron-freshworks: Have you considered creating an Index for the various scss files to help minimize/clean up the imports?

coops-ui/src/mixins/common-mixin.ts
```diff
@@ -1,6 +1,5 @@
 import { Component, Vue } from 'vue-property-decorator'
-import omit from 'lodash.omit'
-import isEqual from 'lodash.isequal'
+import { omit, isEqual } from 'lodash'
```
cameron-freshworks: 👍

coops-ui/src/views/AnnualReport.vue
```diff
@@ -34,115 +34,124 @@
...
-
-        <aside>
-          <affix relative-element-selector="#annual-report-article" :offset="{ top: 120, bottom: 40 }">
-            <sbc-fee-summary
-              v-bind:filingData="[...filingData]"
-              v-bind:payURL="payAPIURL"
-              @total-fee="totalFee=$event"
-            />
-          </affix>
-        </aside>
```
cameron-freshworks: We're going to have some major conflicts here 😢  Wish me luck.

coops-ui/package.json
severinbeauvais: FYI - importing just the functions you need is _usually_ better, but in this case Vuetify (and other packages) load the whole lodash package anyway so not only are we not smaller but we're actually adding extra packages! :/

coops-ui/public/index.html
severinbeauvais: Per Jeremy. Part of the branding/marketing of our apps.

coops-ui/src/App.vue
severinbeauvais: yw

coops-ui/src/App.vue
severinbeauvais: FYI, it might seems weird to import the SCSS in main.ts instead (in fact I prefer to keep presentation stuff together instead of mixing it with code) but it's the way for style order of precedence to work as intended.

coops-ui/src/assets/styles/base.scss
severinbeauvais: Some v-alert components are by default transparent (on a gray app background), so we use this style in a couple of places to make them have a white background and stand out a bit.

coops-ui/src/components/AnnualReport/Certify.vue
severinbeauvais: It's not needed in this section atm, but I left it (but commented out) for easier re-inclusion if we need to import any theme variables later.

coops-ui/src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I found this via unit test errors, but it's always good to write safe code 😛

coops-ui/public/index.html
cameron-freshworks: Very cool, happy to hear we're `Online` now 😄

coops-ui/src/assets/styles/base.scss
cameron-freshworks: right on, i've noticed a lot of styling issues are usually a clash between the component libraries and the designers goals 😄  Thanks for the info!

coops-ui/src/components/AnnualReport/Certify.vue
cameron-freshworks: Ah ok, sounds good to me. I assume this is the case for all the commented out themes. ( Hence why i didn't comment on the rest of them )

coops-ui/src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Absolutely!

coops-ui/src/components/Dashboard/TodoList.vue
severinbeauvais: This was existing code. The diff "forgot where it was" and highlighted a lot more than actually changed.

This style is for the "split menu" (action sub-menu with delete item). Vuetify hard-codes this type of button to be 64px wide, but we want it smaller, so we override it. I've added a code comment to this effect. (Relative EM units do not play a part here.)

As for fonts, they should all be in rems... so I fixed a couple in the next commit.

coops-ui/src/main.ts
severinbeauvais: It's probably not a good idea in this case because this is the only place we want to import all of them (and in this specific order). There's also a 4th scss file that we import by itself in many other places. I think it's safer this way in this case.

coops-ui/src/views/AnnualReport.vue
severinbeauvais: I wish you luck!

coops-ui/public/index.html
severinbeauvais: Hahaha

coops-ui/src/components/Common/OfficeAddresses.vue
severinbeauvais: I did not think of testing this, so I will do it now.

coops-ui/src/components/AnnualReport/Certify.vue
severinbeauvais: Yup, all unneeded ones are commented out atm.

coops-ui/src/components/Common/OfficeAddresses.vue
severinbeauvais: Good catch. Yes, this broke the column alignment, so I'm restoring the code (but in a way that makes it intentional).

PR #524, opened 2019-11-21
UI Features for future effective COA filings

cameron-freshworks:
*Issue #:* /bcgov/entity#1511 /bcgov/entity#1515

*Description of changes:*

* Implements some UI features on the dashboard when a recent filing is future effective.
( A future effective filing is when the filing doesn't take effect until the given date...
ie A COA for BCOMPS ) 

* Disables any more filings from being started until the effective date is met and the filing applied.

* Adds UI styling to display to user the state of the application with a pending tag and tooltip.

* Added an enum for FilingStatus.

* Updated unit testing.

* Updated seeding file for testing purposes.

* Design reference https://projects.invisionapp.com/d/main/#/console/18216776/387274614/preview


severinbeauvais:
I don't have access to the design reference. Could you attach a screenshot of it to the Zenhub ticket please?

coops-ui/tests/unit/Dashboard.spec.ts
```diff
@@ -70,6 +70,18 @@ describe('Dashboard - UI', () => {
...
   })
 
+  it('disables standalone filing buttons & toDo filings when there is a future effective filing pending', () => {
+    wrapper.find(FilingHistoryList).vm.$emit('filings-list', [{ 'status': 'PAID' }])
+    wrapper.find(TodoList).vm.$emit('has-blocker-filing', true)
+
+    expect(wrapper.vm.hasBlockerFiling).toEqual(true)
+    expect(wrapper.vm.coaPending).toEqual(true)
+    expect(wrapper.vm.$el.querySelector('#btn-standalone-addresses')
+      .getAttribute('disabled')).toBeTruthy()
```
severinbeauvais: Try to avoid toBeTruthy() and toBeFalsy() since those aren't type specific and might pass for the wrong reason.
Instead I recommend toBe(true) or toBeDefined() or toBeNull() as needed.

coops-ui/src/views/Dashboard.vue
```diff
@@ -91,7 +137,10 @@ export default {
       todoListFilings: [],
       refreshTimer: null,
       checkFilingStatusCount: 0,
-      inProcessFiling: null
+      inProcessFiling: null,
+      coaPending: false,
+      effectiveDate: null,
+      FilingStatus
```
severinbeauvais: Please add a comment for this one (same as in AnnualReport:297) cause it looks weird among the other 'real' data properties ;)

coops-ui/src/views/Dashboard.vue
```diff
@@ -12,40 +12,80 @@
...
+                        label
+                        color="yellow"
+                        text-color="black"
+                        v-show="coaPending"
+                        v-on="on"
+                      >Pending
+                      </v-chip>
+                    </template>
+                    <span>
+                      The updated office addresses will be legally effective on {{ effectiveDate }}, 12:01 AM(PST).
```
severinbeauvais: I don't think we can say this is "Standard" time (since we might currently be in Daylight time). In other places we just say "Pacific Time". Check with Scott or your designer on this.

coops-ui/src/views/Dashboard.vue
```diff
@@ -181,6 +247,7 @@ export default {
 
   watch: {
     historyFilings () {
+      this.checkPendingFilings(this.historyFilings)
```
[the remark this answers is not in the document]
cameron-freshworks: We could add a conditional but i'm not sure the value. It is additive and not destructive, possibly useable with future entities as well. If you feel it could impact performance, i could certainly add it in. Though which conditional is debatable, i was informed that in most instances COOPs will be the exception when compared with other entity types, whereas BCOMPs will be more similar. Thoughts on this?

coops-ui/tests/unit/Dashboard.spec.ts
cameron-freshworks: Great idea, thank you, will update!

coops-ui/src/views/Dashboard.vue
cameron-freshworks: Done and done, will do, thanks!

coops-ui/src/views/Dashboard.vue
cameron-freshworks: Ok great, thanks for pointing that out, i will update for consistency!

cameron-freshworks:
> I don't have access to the design reference. Could you attach a screenshot of it to the Zenhub ticket please?

Absolutely, heres a screenshot. 
Screen Shot 2019-11-28 at 8 27 43 AM

coops-ui/src/views/Dashboard.vue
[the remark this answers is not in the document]
severinbeauvais: I think performance isn't an issue; this will only be run once after all.

I'm just concerned about the possibility of accidentally applying (or appearing to apply) business rules to non-BCorp businesses. In the code one cannot tell that this applies only to BCorps. A conditional would clarify this, but a comment might suffice.

As for "if (BCorp)" vs "if (!Coop)", I prefer the first since today we know this applies only to BComps (instead of assuming it applies to all but Coops).

coops-ui/tests/unit/Dashboard.spec.ts
```diff
@@ -55,19 +55,19 @@ describe('Dashboard - UI', () => {
 
     expect(wrapper.vm.hasBlockerFiling).toEqual(false)
     expect(wrapper.vm.$el.querySelector('#btn-standalone-addresses')
-      .getAttribute('disabled')).toBeFalsy()
+      .getAttribute('disabled')).not.toBe('true')
     expect(wrapper.vm.$el.querySelector('#btn-standalone-directors')
-      .getAttribute('disabled')).toBeFalsy()
+      .getAttribute('disabled')).not.toBe('true')
```
severinbeauvais: Actually, I think in this case the attribute is undefined. Which is not true. But lots of things are not true. Anyway...

coops-ui/tests/unit/Dashboard.spec.ts
```diff
@@ -55,19 +55,19 @@ describe('Dashboard - UI', () => {
...
   })
 
   it('disables standalone filing buttons when there is a blocker filing in the to-do list', () => {
     wrapper.find(TodoList).vm.$emit('has-blocker-filing', true)
 
     expect(wrapper.vm.hasBlockerFiling).toEqual(true)
     expect(wrapper.vm.$el.querySelector('#btn-standalone-addresses')
-      .getAttribute('disabled')).toBeTruthy()
+      .getAttribute('disabled')).toEqual('true')
```
severinbeauvais: Does this work? When this attribute is set, I thought its value is actually "disabled" (ie, `disabled="disabled"`)?

coops-ui/src/views/Dashboard.vue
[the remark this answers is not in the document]
cameron-freshworks: Sounds good, done!

coops-ui/tests/unit/Dashboard.spec.ts
cameron-freshworks: What would you like to see here. The code we're changing was originally `toBeFalsy()` & `toBeTruthy()` in master, so i followed the pattern that was done. What is your preference of pattern?

cameron-freshworks: I specified `null` in this case, as that is what the attribute is when false.

coops-ui/tests/unit/Dashboard.spec.ts
cameron-freshworks: It works. It returns 'true' as the value of this attribute is a boolean.  `hasBlockerFiling`

coops-ui/tests/unit/Dashboard.spec.ts
severinbeauvais: Perfect, thanks.

coops-ui/tests/unit/Dashboard.spec.ts
severinbeauvais: OK, thanks.

coops-ui/src/views/Dashboard.vue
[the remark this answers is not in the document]
severinbeauvais: Comment is fine, thanks.

PR #530, opened 2019-11-22
#1508 COA Warning Modal

cameron-freshworks:
*Issue #:* /bcgov/entity#1508

*Description of changes:*

* Implements a warning modal for BCORPS before they enter the change of address filing.

* Updated Unit testing.

* Dependencies *

* This ticket can't be merged into 1388-coa-bcorp-story feature branch until  PR https://github.com/bcgov/lear/pull/524 is merged in. 


coops-ui/tests/unit/Dashboard.spec.ts
```diff
@@ -87,7 +94,7 @@ describe('Dashboard - UI', () => {
     localVue.use(VueRouter)
     const router = mockRouter.mock()
     router.push({ name: 'dashboard', query: { filing_id: '123' } })
-    const wrapper = mount(Dashboard, { localVue, store, router, vuetify })
+    const wrapper = shallowMount(Dashboard, { localVue, store, router, vuetify })
```
severinbeauvais: Do all unit tests pass? I thought we needed a deep mount for the sub-components, but maybe not!

coops-ui/tests/unit/Dashboard.spec.ts
```diff
@@ -142,19 +149,47 @@ describe('Dashboard - Click Tests', () => {
...
     const vm = wrapper.vm as any
+    vm.coaWarning = false
 
     Vue.nextTick(async () => {
       const button = vm.$el.querySelector('#btn-standalone-addresses')
-      expect(button.querySelector('.v-btn__content').textContent).toContain('Change')
+      expect(button.textContent).toContain('Change')
       await button.click()
 
+      expect(vm.coaWarning).toBe(true)
```
severinbeauvais: This doesn't prove the dialog is displaying. Is it possible to check that the dialog element(s) are defined and not display=none (or something like that)?

I agree that it's probably not worth having a separate unit test component for the dialog component, but if you think it's worthwhile then you could see how SaveErrorDialog.spec.ts does it.

coops-ui/src/components/Dashboard/CoaWarningDialog.vue
```diff
@@ -11,7 +11,7 @@
       <v-card-actions>
         <v-btn color="primary" text @click="toggle()">Return to dashboard</v-btn>
       <v-spacer></v-spacer>
-        <v-btn color="primary" text @click="proceed()">Continue to Change of Address</v-btn>
+        <v-btn id="btn-proceed-coa" color="primary" text @click="proceed()">Continue to Change of Address</v-btn>
```
severinbeauvais: Is this an id for QA? If so, cool. And could you also add a id to the toggle button?

coops-ui/src/components/Dashboard/CoaWarningDialog.vue
```diff
@@ -0,0 +1,42 @@
...
+
+  // Pass click event to parent.
+  @Emit() private proceed () {}
+}
+</script>
+
+<style lang="scss" scoped>
+  @import '@/assets/styles/theme.scss';
+
+  .genErr{
```
severinbeauvais: I moved this to `base.scss` in another PR. If your PR gets merged first, can you remind me to remove this from here in my PR?

coops-ui/src/components/Dashboard/CoaWarningDialog.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <v-dialog v-model="dialog" width="45rem" persistent>
+    <v-card>
+      <v-card-title>Address Change Effective 12:01 AM</v-card-title>
+      <v-card-text>
+        <p class="genErr">Address changes take effect at 12:01 AM (PST). No other filings can be performed until an
```
severinbeauvais: As in another comment, this should probably be changed to "Pacific Time" so we don't have to worry about Standard vs Daylight hours.

coops-ui/src/views/Dashboard.vue
```diff
@@ -236,6 +250,20 @@ export default {
           this.hasBlockerFiling = true
         }
       })
+    },
+    /**
+     * Toggle the Change of address warning dialog.
+     */
+    toggleCoaWarning () {
+      this.coaWarning = !this.coaWarning
```
severinbeauvais: Aren't you always setting coaWarning to false here? I'd prefer more explicit code (and function names), just to keep it simple to understand.
(not a mandatory change)

coops-ui/tests/unit/Dashboard.spec.ts
cameron-freshworks: They do indeed! The child components are stubbed/component placeholders in the shallowmount. Since they are not subject to any real testing here, just to ensure they are present, shallowmount seems to make sense. Shallowmount also helps us test the dashboard in isolation without any conflicts from child components. What are your thoughts on this?

coops-ui/tests/unit/Dashboard.spec.ts
cameron-freshworks: Good point, i will update this!

coops-ui/src/components/Dashboard/CoaWarningDialog.vue
cameron-freshworks: It is indeed, ok will do!

coops-ui/src/components/Dashboard/CoaWarningDialog.vue
cameron-freshworks: yes sir, will do!

coops-ui/src/components/Dashboard/CoaWarningDialog.vue
cameron-freshworks: k sounds great, will adjust np.

coops-ui/src/views/Dashboard.vue
cameron-freshworks: It is a toggle, so it will change it to the opposing state of whatever it currently is, to save having two methods ( one to close, one to open) or one method with conditionals.  I will certainly update the comments and method name!

coops-ui/tests/unit/Dashboard.spec.ts
severinbeauvais: Agreed!

coops-ui/src/views/Dashboard.vue
severinbeauvais: I get it, but we never using this function to display the warning, so it seems like overkill. Not a big deal though.

cameron-freshworks: Sorry, you were right. I wasn't calling the method to display the dialog, just to close it. So i adjusted it and now correctly am calling the toggle method to display it and calling the toggle method to close it. Thanks!

coops-ui/tests/unit/Dashboard.spec.ts
cameron-freshworks: How is this? I used the stubbed child component to emit the value to proceed with the routing?

severinbeauvais: After trying a few things myself, I find this is a decent way to do it.

As you know, the problem is that dialogs are injected dynamically at the top-most level of the Vue app element tree, and so you cannot "see" them from within a component further down the tree (as when testing that component in a unit test).

So, your approach to exercise the dialog and check for results seems OK. Thanks for this.

coops-ui/src/components/Dashboard/CoaWarningDialog.vue
severinbeauvais: Code fixed, thanks.

coops-ui/src/views/Dashboard.vue
severinbeauvais: IMO in this case I think it would be more clear to SHOW the dialog explicitly and then HIDE the dialog explicitly, instead of toggling it. The problem is you could get out of sync with the toggling, and then your call to show it might instead hide it and vice-versa. (JS errors cause interesting behaviours sometimes.)

However, let's leave it for now.

coops-ui/tests/unit/Dashboard.spec.ts
severinbeauvais: Next time please add comments explaining what your code is doing... especially in this case since it appears to be doing something unusual.

coops-ui/src/views/Dashboard.vue
cameron-freshworks: Totally fair point, i understand where you're coming from completely. Being explicit leaves less room for any weird interactions that may occur from errors. I may have overcomplicated something simple like this, i think in my head i always try to process solutions that are dynamic in nature. Appreciate your feedback on this!

cameron-freshworks:
> Thanks for the changes. You have conflicts to resolve. I'll have a quick look after you fix those and then I'll merge this PR.

Thanks sir, as always appreciate all the feedback and recommendations 👍

PR #540, opened 2019-11-26
UI - allow spaces in director names, but do not allow empty spaces only

katiemcgoff:
*Issue #:* /bcgov/entity#1824

*Description of changes:*
Allow spaces in name (such as "H. Mary") but do not allow only spaces (" ").

### DIFF
--- a/coops-ui/src/components/AnnualReport/Directors.vue
+++ b/coops-ui/src/components/AnnualReport/Directors.vue
@@ -598,7 +598,7 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
   private get directorFirstNameRules (): Array<Function> {
     return [
       v => !!v || 'A first name is required',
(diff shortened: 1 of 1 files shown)


katiemcgoff:
> I fixed this in #520 but it hasn't been merged yet.
> 
> However, if you want just the minimum fix, see https://github.com/bcgov/lear/pull/520/files#diff-211c31457df1cf3235e825b522ccf32fR612 and https://github.com/bcgov/lear/pull/520/files#diff-211c31457df1cf3235e825b522ccf32fR625.

Perfect. I'll push this through then, to fix the blocking error for users, and it can be replaced by your more thorough fix when that goes in. Also you've got tests so that's great.

FYI the blocking issue is that they can't save an address change if their name is now invalid, but there's no feedback to them because those fields aren't showing. Hence the urgency.

PR #565, opened 2019-12-03
updated dialogs (no contact info for staff, cleanup)

severinbeauvais:
*Issue #:* /bcgov/entity#1733

*Description of changes:*

- hid contact info for staff in dialogs
- updated some dialog text per wireframes
- changed Okay buttons to OK (per Scott)
- aligned dialog buttons consistently
- made all dialogs persistent (must click button to close)
- made Save Error dialog narrower (per Scott)
- added IDs to dialogs
- updated unit tests
- fixed missing enum rename
- changed Dashboard responsive breakpoint to medium (per Scott) (no change to main views/pages)

### DIFF
--- a/coops-ui/src/components/AnnualReport/PaymentErrorDialog.vue
+++ b/coops-ui/src/components/AnnualReport/PaymentErrorDialog.vue
@@ -1,20 +1,21 @@
 <template>
-  <v-dialog v-model="dialog" width="45rem">
+  <v-dialog id="payment-error-dialog" v-model="dialog" width="45rem" persistent>
     <v-card>
       <v-card-title>Unable to Process Payment</v-card-title>
 
       <v-card-text>
         <p class="genErr">We are unable to process payments at this time.</p>
-        <p class="genErr">Your filing has been saved as a DRAFT and you can resume your filing from
-          the Business Dashboard at a later time.</p>
-        <p class="genErr">PayBC is normally available:</p>
-        <p class="genErr">
-          Monday to Friday: 6:00am to 9:00pm
-          <br>Saturday: 12:00am to 7:00pm
-          <br>Sunday: 12:00pm to 12:00am
-        </p>
+        <p class="genErr">Your filing has been saved as a DRAFT and you can resume
+          your filing from the Business Dashboard at a later time.</p>
 
         <template v-if="!isRoleStaff">
+          <p class="genErr">PayBC is normally available:</p>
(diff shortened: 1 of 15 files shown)


coops-ui/src/components/AnnualReport/PaymentErrorDialog.vue
```diff
@@ -1,38 +1,54 @@
 <template>
-  <v-dialog v-model="dialog" width="45rem">
+  <v-dialog id="payment-error-dialog" v-model="dialog" width="45rem" persistent>
```
cameron-freshworks: I like the addition of `persistent` tho to be honest, i would have assumed it to be default! Good job.

coops-ui/src/components/AnnualReport/SaveErrorDialog.vue
```diff
@@ -44,20 +44,28 @@
         </v-card-actions>
         <v-card-actions v-else>
           <v-spacer></v-spacer>
-          <v-btn id="okay-btn" color="primary" text @click="okay()">Okay</v-btn>
+          <v-btn id="okay-btn" color="primary" text @click="okay()">OK</v-btn>
```
cameron-freshworks: Love the verbiage clarification ;)

severinbeauvais: This is the way Scott wants it :)

severinbeauvais:
> Looks good to me 👍
> 
> Just to clarify, what is the use case of the id's in the dialogs, is it mainly for testing?
> 
> Also, would this be a good time to reshuffle the location of the dialogs into a `Dialogs` folder, inside the common directory, use an index for cleaning imports app wide? It would require some import refactoring in quite a few components so it may fall outside the scope of this ticket, but just curious on your thoughts.
> 
> Approved!

I am working on a sub-ticket of /bcgov/entity#1721 which is specifically adding IDs to facilitate testing. My approach is:
- Things that should have IDs anyways will get `id=...` (and we can use those in our unit tests).
- Things that mainly require IDs for QA will get `data-test-id=...`. All our code should work without these (say, by stripping them all out with a preprocessor).

Actually, adding IDs to the v-dialogs won't work (because the dialogs are attached outside the components). But I figured out how to do this... and will refactor in my next PR!

My **next** PR will move the dialog components to a 'dialogs' folder! Great minds think alike :)

PR #568, opened 2019-12-03
Create Data Loader Service entity#1793 Take 2! (Rebasing was fubared)  

ConradGustafson:
*Issue #:* /bcgov/entity#1793

*Description of changes:*

Created an API that connects to a PostgreSQL database and imports or exports businesses. The details are in the README.md

Updated to take into account the model changes for offices and to do a few cleanups as requested on the previous pull request.

Did not refactor to pull in the model as a library. After investigation and discussion with DevOps (Brian), it looks like using the existing mechanism is broken on versioning. Doing something like this:
```
git+https://github.com/bcgov/lear.git#egg=registry_schemas&subdirectory=schemas
git+https://github.com/bcgov/lear.git#egg=legal_api&subdirectory=legal-api
```
-- creates a hard dependency on the versions that are checked into master. The intention of this service is that it can be built locally or on a PR based on whatever version of the model is also checked in at that tag or commit. If master changes, you can still build other branches or tags without being affected. This also required the /legal-test-fixture/legal-api path in the source, which does not conflict but does look a little weird.

Maybe in the future the model will be managed as an actual library, at which point the code in this (and /legal_api and test_data) will have to manage their version dependency in a proper way. I strongly suggest a refactoring of this stuff that allows it to be versioned properly **and** reusable. Until that happens I suggest that this test utility go in as is. 

Note: requirements.txt still includes this:
`git+https://github.com/bcgov/lear.git#egg=registry_schemas&subdirectory=schemas`
-- but the schemas aren't actually being used so the poor encapsulation doesn't matter.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0)


thorwolpert:
Logically if you copy in the source, or call it in as a library, there is little to no difference since the SQL will break as it migrates. If that was really a concern there are other libraries to use to load/unload data that can mitigate that or even just rely on column order of the csv/excel, such as the postgres built-in command COPY as suggested on the other PR. Also, if like the pytests Migrate was used instead of the model rebuild, then the lookups wouldn't need to be added to both migrations and the source code here (see conftest).

thorwolpert:
adding cross team SRE members for review.

thorwolpert:
As requested before, please provide a BC,DC at a minimum so that this can run in OpenShift, and maybe a pipeline if not already included in the new pipeline.

pwei1018:
Can you post pylint and flake8 result of this project?

katiemcgoff:
I've pulled this into the bcgov/bcrs-utilities repo as-is, to be adapted for all projects.

PR #575, opened 2019-12-04
fix for bug 2044

peter-freshworks:
*Issue #:* /bcgov/entity#2044

*Description of changes:*
Task list not taking pending filings into account for B-Comp. Fix the todo-list constructor method to account for pending/draft filings for non CO-OP entities as well.

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_tasks.py
+++ b/legal-api/src/legal_api/resources/business/business_tasks.py
@@ -80,27 +80,28 @@ class TaskListResource(Resource):
             tasks.append(task)
             order += 1
 
-        if check_agm:
-            last_ar_date = business.last_ar_date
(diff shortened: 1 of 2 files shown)


katiemcgoff:
Hi Peter, are there any tests that need to be updated or added?

peter-freshworks:
> y

We have coverage for tasks/todo lists, but I can add something a bit more specific for this bug

peter-freshworks:
> Hi Peter, are there any tests that need to be updated or added?

We have coverage for tasks/todo lists, but I can add something a bit more specific for this bug

peter-freshworks:
> Hi Peter, are there any tests that need to be updated or added?

I stand corrected, I see no coverage for b-comp task lists and pending filings

PR #580, opened 2019-12-05
release 2.3.0 merge cleanup

severinbeauvais:
*Issue #:* /bcgov/entity#2053

*Description of changes:*
- fixed unit tests
- misc cleanup
- added types to store actions, getters and state

### DIFF
--- a/coops-ui/src/mixins/common-mixin.ts
+++ b/coops-ui/src/mixins/common-mixin.ts
@@ -12,7 +12,7 @@ export default class CommonMixin extends Vue {
    * @param baseObj The base object.
    * @param prop The property to be removed.
    */
-  omitProp (baseObj: object, prop: Array<string>): Object {
+  omitProp (baseObj: object, prop: Array<string>): object {
     return omit(baseObj, prop)
   }
 
--- a/coops-ui/src/store/actions.ts
+++ b/coops-ui/src/store/actions.ts
@@ -1,74 +1,74 @@
 export default {
-  setKeycloakRoles ({ commit }, keycloakRoles) {
+  setKeycloakRoles ({ commit }, keycloakRoles: Array<string>) {
     commit('keycloakRoles', keycloakRoles)
   },
-  setAuthRoles ({ commit }, authRoles) {
+  setAuthRoles ({ commit }, authRoles: Array<string>) {
     commit('authRoles', authRoles)
   },
-  setCurrentDate ({ commit }, currentDate) {
+  setCurrentDate ({ commit }, currentDate: string) {
     commit('currentDate', currentDate)
   },
-  setLastAgmDate ({ commit }, lastAgmDate) {
+  setLastAgmDate ({ commit }, lastAgmDate: string) {
     commit('lastAgmDate', lastAgmDate)
   },
-  setNextARDate ({ commit }, nextARDate) {
+  setNextARDate ({ commit }, nextARDate: string) {
     commit('nextARDate', nextARDate)
   },
-  setARFilingYear ({ commit }, ARFilingYear) {
+  setARFilingYear ({ commit }, ARFilingYear: number) {
     commit('ARFilingYear', ARFilingYear)
   },
-  setEntityBusinessNo ({ commit }, entityBusinessNo) {
+  setEntityBusinessNo ({ commit }, entityBusinessNo: string) {
     commit('entityBusinessNo', entityBusinessNo)
   },
-  setEntityName ({ commit }, entityName) {
+  setEntityName ({ commit }, entityName: string) {
     commit('entityName', entityName)
   },
-  setEntityType ({ commit }, entityType) {
+  setEntityType ({ commit }, entityType: string) {
     commit('entityType', entityType)
   },
-  setEntityStatus ({ commit }, entityStatus) {
+  setEntityStatus ({ commit }, entityStatus: string) {
     commit('entityStatus', entityStatus)
   },
-  setEntityIncNo ({ commit }, entityIncNo) {
+  setEntityIncNo ({ commit }, entityIncNo: string) {
     commit('entityIncNo', entityIncNo)
   },
-  setEntityFoundingDate ({ commit }, entityFoundingDate) {
+  setEntityFoundingDate ({ commit }, entityFoundingDate: string) {
     commit('entityFoundingDate', entityFoundingDate)
   },
-  setBusinessEmail ({ commit }, businessEmail) {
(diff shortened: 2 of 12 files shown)


coops-ui/src/store/actions.ts
```diff
@@ -1,71 +1,71 @@
...
   },
-  setDirectors ({ commit }, directors) {
+  setDirectors ({ commit }, directors: Array<object>) {
     commit('directors', directors)
   },
-  setTriggerDashboardReload ({ commit }, value) {
+  setTriggerDashboardReload ({ commit }, value: boolean) {
     commit('triggerDashboardReload', value)
   },
   setLastAnnualReportDate ({ commit }, value) {
```
cameron-freshworks: Nice job!

coops-ui/src/components/Common/SummaryDirectors.vue
```diff
@@ -208,9 +208,9 @@ export default class SummaryDirectors extends Mixins(DateMixin, EntityFilterMixi
   EntityTypes: {} = EntityTypes
 
   /**
-    * On-Created hook to create two new arrays of directors for summary output
-    * One array will contain new/changed directors
-    * One array will contain all ceased directors
+    * On-Created hook to create two new arrays of directors for summary output.
+    * - "directorSummary" will contain new/changed directors
+    * - "directorsCeased" will contain all ceased directors
```
cameron-freshworks: Was the method itself changed in another PR?

coops-ui/src/mixins/common-mixin.ts
```diff
@@ -12,7 +12,7 @@ export default class CommonMixin extends Vue {
    * @param baseObj The base object.
    * @param prop The property to be removed.
    */
-  omitProp (baseObj: object, prop: Array<string>): Object {
+  omitProp (baseObj: object, prop: Array<string>): object {
```
cameron-freshworks: Good catch, i must have missed this when we discussed this before, thank you!

coops-ui/src/store/state.ts
```diff
@@ -1,33 +1,33 @@
 export default {
   // tombstone data
-  keycloakRoles: [],
-  authRoles: [],
-  username: null,
-  currentDate: null,
+  keycloakRoles: [] as Array<string>,
+  authRoles: [] as Array<string>,
+  username: null as string,
+  currentDate: null as string,
```
cameron-freshworks: Great job 👍 !! This goes for all the type refactoring on all the store files 💯

coops-ui/src/components/Common/SummaryDirectors.vue
severinbeauvais: No, I simply updated the comment. PR #577 actually changes this method. I'm expecting a conflict but this way I will remember to update the comment :)

coops-ui/src/mixins/common-mixin.ts
severinbeauvais: Yeah, I also learned that I should use `number` instead of `Number`.

The basic rule, I believe, is to use the lower case name for type declarations, but the upper case name when using `new Number()` or when using any prototype methods like `Number.toFixed(123)`

coops-ui/src/views/AnnualReport.vue
```diff
@@ -424,7 +422,7 @@ export default {
               // set the Initial AGM Date in the AGM Date component
               // NOTE: AR Filing Year (which is needed by agmDate component) was already set by Todo List
               this.initialAgmDate = annualReport.annualGeneralMeetingDate
-              if (annualReport.annualGeneralMeetingDate == null) {
+              if (!annualReport.annualGeneralMeetingDate) {
```
cameron-freshworks: Is `annualReport.annualGeneralMeetingDate` defined explicitly as a boolean or defined in general? Either way, this is a good refactor. It now covers null, undefined, NaN, false, or an empty string.  Good job!

severinbeauvais: It's a string. I was going to replace == with === but I realized that == also handles other falsy cases, so I just used ! directly.

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -165,20 +165,23 @@ describe('Directors as a COOP', () => {
...
 
-  it('disables buttons/actions when instructed by parent component', () => {
-    // invalidate AGM Date
-    const wrapper = createWrapper(vm)
-    wrapper.setProps({ componentEnabled: false })
-    // confirm that flag is set correctly
-    expect(vm.componentEnabled).toEqual(false)
+  it('disables buttons/actions when instructed by parent component', done => {
+    // clear enabled prop
+    vm.componentEnabled = false
```
cameron-freshworks: I don't know if this has any real value but if we weren't mounting the Component in the `beforeEach`  ( which we are in this test suite ) we could pass the propsData in when we mount the component to test the local property a little more organically. ie `
    const Constructor = Vue.extend(Directors)
    const instance = new Constructor({ store, vuetify, propsData: { componentEnabled: false } })
    vm = instance.$mount()`

No change at all required here, just curious if think this approach would cover more in the unit test opposed to setting it explicitly?

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -598,7 +606,7 @@ describe('Appoint New Director tests', () => {
     setTimeout(() => {
       expect(vm.$el.querySelectorAll('.v-messages')[0].textContent).toBe('')
       done()
-    }, 100)
+    }, 200)
```
cameron-freshworks: Just out of curiosity, what is the purpose of the timeout?

coops-ui/src/components/Common/SummaryDirectors.vue
severinbeauvais: Changed reverted.

coops-ui/tests/unit/Directors.spec.ts
severinbeauvais: Good idea... for future refactoring :)

coops-ui/tests/unit/Directors.spec.ts
severinbeauvais: It's because waiting for the next tick doesn't work in this scenario... unsure why but enough time was wasted on it... so this is a work-around but it works!
Since unit tests appear to be run concurrently, when there are more tests this number sometimes need to be increased.
Honestly, I would like to make this work correctly once and for all, but as I said, this was time-boxed.

PR #588, opened 2019-12-06
#1934 directors name change bug

cameron-freshworks:
*Issue #:* /bcgov/entity#1934

*Description of changes:*

Resets the directors name to original correctly when cancelling a free directors name change. 


cameron-freshworks:
> Looks good to me. I'm just going to pull it down to test locally.
> 
> Speaking of tests, there should be tests around this :) Keeping in mind that they're mainly so someone doesn't screw up your code (aka lose this functionality) in the future.

Great idea, I will write a specific unit test to confirm the data being restored and the actions being cleared 👍

cameron-freshworks:
> Looks good to me. I'm just going to pull it down to test locally.
> 
> Speaking of tests, there should be tests around this :) Keeping in mind that they're mainly so someone doesn't screw up your code (aka lose this functionality) in the future.

The unit tests have been expanded!

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -1093,8 +1094,11 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
 
   /**
    * Local helper to cancel a director that was edited.
+   * @param id Id of the director being edited.
    */
-  private cancelEditDirector (): void {
+  private cancelEditDirector (id = null): void {
+    if (id) this.restoreDirName(id - 1)
```
severinbeauvais: Should check that id > 0 otherwise "id - 1" could be negative. You could also move this check to restoreDirName().

PS null == 0

cameron-freshworks: Great idea, i will update this.

cameron-freshworks: Updated 👍

PR #592, opened 2019-12-06
Ui dashboard to do list

cameron-freshworks:
Issue #: /bcgov/entity#2050

Description of changes:
- Updates the UI for the ToDo list filings when they are Annual Reports for BCOMP's.
- Disables filing until an acknowledgement from the user about their current information.
- Minor Styling changes.
- New and updated unit testing.


coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -15,15 +15,31 @@
         v-for="(item, index) in orderBy(taskItems, 'order')"
         v-bind:key="index"
         expand-icon=""
-        :class="{ 'disabled': !item.enabled, 'draft': isDraft(item) }">
+        :class="{ 'disabled': !item.enabled, 'draft': isDraft(item),
```
severinbeauvais: Should these conditions be mutually exclusive -- what happens if all 3 classes are applied at once? Ref: https://vuejs.org/v2/guide/class-and-style.html

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -15,15 +15,31 @@
...
+                  label="All information about the Office Addresses and Current Directors is correct."
+                  :disabled=!item.enabled
+                  v-model=confirmCheckbox
+                  @click.native.stop
+                />
+              </div>
               <div class="list-item__subtitle">
-                <div v-if="item.subtitle" class="todo-status">{{item.subtitle}}</div>
-
+                <div v-if="item.subtitle && !entityFilter(EntityTypes.BCORP)" class="todo-status">
```
severinbeauvais: For now I'd prefer to see this element _included_ for Coops only instead of _excluded_ for BCorps.
The way I think of this is that I know what I know (ie, it's needed for Coops) but I don't know that it applies to everything except BCorps. Today the latter might be true but it will be an assumption for any future entity type, instead of considering future entity types separately.

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -118,12 +140,14 @@
...
               <v-btn v-else-if="!isCompleted(item)"
+                class="btn-file"
                 color="primary"
-                :disabled="!item.enabled || coaPending"
+                :disabled="(!item.enabled || coaPending) || !confirmCheckbox"
                 @click.native.stop="doFileNow(item)"
               >
                 File Now
               </v-btn>
+              </v-col>
```
severinbeauvais: Please fix block alignments.

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -210,6 +250,7 @@ export default {
 
     loadData () {
       this.taskItems = []
+      this.confirmCheckbox = this.entityFilter(EntityTypes.COOP)
```
severinbeauvais: I am unclear about the business rules around this statement. Please add a high level comment, something like "is not used for Coops so set this to true to make page logic work".
FWIW, although this logic is correct, it would be more clear to code "in the case of Coop, set this to true" (otherwise it stays initialized to false as per data() ).

coops-ui/tests/unit/TodoList.spec.ts
```diff
@@ -51,6 +58,9 @@ describe('TodoList - UI', () => {
               'name': 'annualReport',
               'ARFilingYear': 2017,
               'status': 'NEW'
```
severinbeauvais: Could status be initialized from enum?

coops-ui/src/components/Dashboard/TodoList.vue
cameron-freshworks: Good question, i do not have the answer :) I will look into this!

coops-ui/src/components/Dashboard/TodoList.vue
cameron-freshworks: Ok, that makes perfect sense to me. I will change this!

coops-ui/src/components/Dashboard/TodoList.vue
cameron-freshworks: Will do, thanks, good catch!

coops-ui/src/components/Dashboard/TodoList.vue
cameron-freshworks: I can definitely write this a little more clearly, at first glance it doesn't give you much information of the actual purpose of this line of code.

coops-ui/tests/unit/TodoList.spec.ts
cameron-freshworks: absolutely, i must have missed one. Thanks!

cameron-freshworks: Ohh, these are the in spec file! I'm not sure whats the way to go here, in my opinion i don't mind using the string values at all in tests as they can also double as a validation of the enums/constants used in the code. What are your thoughts on that?

coops-ui/src/components/Dashboard/TodoList.vue
cameron-freshworks: In this instance, if all three classes are applied at once the expansion panel would be `display: none` as it would supersede the other styles. That being said, the classes 'draft' and 'awaitConfirm' couldn't both exist on the panel due to logic that only enables the button on a BCOMPs New Annual Reports ( whom have no drafts ). The 'disabled' class can exist under many situations tho, so i would say that these three classes are not mutually exclusive. That is an interesting read though, and i dig into it a little deeper and see if there is more improvements to be made. I'm sure there is so much more we can do with Vue still yet :)

severinbeauvais: Thanks for looking into this.

coops-ui/tests/unit/TodoList.spec.ts
severinbeauvais: Good point! I agree with you.

coops-ui/src/components/Dashboard/TodoList.vue
severinbeauvais: Thanks, looks good.

coops-ui/src/components/Dashboard/TodoList.vue
severinbeauvais: Thanks for the change. I think this will help anyone else looking at that code.

PR #598, opened 2019-12-10
Configuration lookup for certify component

peter-freshworks:
*Issue #:* /bcgov/entity#1550

*Description of changes:*
Implement a configuration mechanism for composing front end components dynamically. This is for the certify component and can be expanded to, in the future, be used for other portions of the application with possible use alongside a localization library 

### DIFF
--- a/coops-ui/package.json
+++ b/coops-ui/package.json
@@ -9,7 +9,8 @@
     "lint:nofix": "vue-cli-service lint --no-fix",
     "test:e2e": "vue-cli-service test:e2e",
     "test:unit": "vue-cli-service test:unit --testPathPattern=.",
-    "test:unit-silent": "vue-cli-service test:unit --silent"
+    "test:unit-silent": "vue-cli-service test:unit --silent",
+    "test:debug": "node --inspect-brk node_modules/.bin/vue-cli-service test:unit --testPathPattern --no-cache --watch --runInBand"
   },
   "dependencies": {
     "@mdi/font": "^4.5.95",
--- a/coops-ui/package.json
+++ b/coops-ui/package.json
@@ -8,7 +8,7 @@
     "lint": "vue-cli-service lint",
     "lint:nofix": "vue-cli-service lint --no-fix",
     "test:e2e": "vue-cli-service test:e2e",
-    "test:unit": "vue-cli-service test:unit --testPathPattern",
+    "test:unit": "vue-cli-service test:unit --testPathPattern=.",
     "test:unit-silent": "vue-cli-service test:unit --silent",
     "test:debug": "node --inspect-brk node_modules/.bin/vue-cli-service test:unit --testPathPattern --no-cache --watch --runInBand"
   },
--- a/coops-ui/package.json
+++ b/coops-ui/package.json
@@ -8,7 +8,7 @@
     "lint": "vue-cli-service lint",
     "lint:nofix": "vue-cli-service lint --no-fix",
     "test:e2e": "vue-cli-service test:e2e",
-    "test:unit": "vue-cli-service test:unit --testPathPattern=.",
+    "test:unit": "vue-cli-service test:unit --testPathPattern",
     "test:unit-silent": "vue-cli-service test:unit --silent",
     "test:debug": "node --inspect-brk node_modules/.bin/vue-cli-service test:unit --testPathPattern --no-cache --watch --runInBand"
   },
--- a/coops-ui/src/components/AnnualReport/Certify.vue
+++ b/coops-ui/src/components/AnnualReport/Certify.vue
@@ -22,15 +22,13 @@
         <template slot="label">
           <div class="certify-stmt">
             I, <b>{{trimmedCertifiedBy || '[Legal Name]'}}</b>, certify that I have relevant knowledge of the
-            association and that I am authorized to make this filing.
+            {{ entityDisplay || 'association' }} and that I am authorized to make this filing.
           </div>
         </template>
       </v-checkbox>
       <p class="certify-clause">{{currentDate}}</p>
       <p class="certify-clause">
-        Note: It is an offence to make a false or misleading statement in
-        respect of a material fact in a record submitted to the Corporate Registry for filing.
-        See section 200 of the Cooperatives Association Act.
+        {{message}}
       </p>
     </div>
   </v-card>
@@ -38,7 +36,7 @@
 
 <script lang="ts">
 import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
-
+import { configJson } from '@/resources'
 @Component({})
 export default class Certify extends Vue {
   // Props passed into this component.
@@ -51,6 +49,11 @@ export default class Certify extends Vue {
   @Prop({ default: '' })
   private currentDate: string
 
+  @Prop({ default: '' })
+  private message: string
+
+  @Prop({ default: '' })
+  private entityDisplay: string
   /**
    * Lifecycle callback to always give the parent a "valid" event for its property values.
    */
--- a/coops-ui/src/components/AnnualReport/Certify.vue
+++ b/coops-ui/src/components/AnnualReport/Certify.vue
@@ -36,7 +36,6 @@
 
 <script lang="ts">
 import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
-import { configJson } from '@/resources'
 @Component({})
 export default class Certify extends Vue {
   // Props passed into this component.
--- a/coops-ui/src/main.ts
+++ b/coops-ui/src/main.ts
@@ -35,7 +35,6 @@ configHelper.fetchConfig()
     // ensure we have a Keycloak token
     if (!sessionStorage.getItem('KEYCLOAK_TOKEN')) {
       console.log('Redirecting to Auth URL...')
-      sessionStorage.setItem('REDIRECTED', 'true')
       const authUrl = sessionStorage.getItem('AUTH_URL')
       // assume Auth URL is always reachable
       window.location.assign(authUrl)
--- a/coops-ui/src/main.ts
+++ b/coops-ui/src/main.ts
@@ -34,12 +34,10 @@ configHelper.fetchConfig()
   .then(() => {
     // ensure we have a Keycloak token
     if (!sessionStorage.getItem('KEYCLOAK_TOKEN')) {
-      resetAuth(() => {
-        sessionStorage.setItem('REDIRECTED', 'true')
-        const authUrl = sessionStorage.getItem('AUTH_URL')
-        // assume Auth URL is always reachable
-        window.location.assign(authUrl)
-      })
+      sessionStorage.setItem('REDIRECTED', 'true')
+      const authUrl = sessionStorage.getItem('AUTH_URL')
+      // assume Auth URL is always reachable
+      window.location.assign(authUrl)
       return // do not execute remaining code
     }
     new Vue({
@@ -54,20 +52,3 @@ configHelper.fetchConfig()
     console.error('error fetching config -', error)
     alert('Fatal error loading app')
   })
-
-async function resetAuth (func) {
-  var xhr = new XMLHttpRequest()
-  xhr.open('POST', 'https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/token', true)
-  xhr.setRequestHeader('Content-Type', 'application/json;charset=UTF-8')
-  xhr.onload = function () {
-    // do something to response
-    var data = JSON.parse(this.responseText)
-    if (data && data.access_token) {
-      sessionStorage['KEYCLOAK_TOKEN'] = data['access_token']
-      sessionStorage['BUSINESS_IDENTIFIER'] = 'CP0001327'
-    } else {
-      func()
-    }
-  }
-  await xhr.send(JSON.stringify({ 'username': 'CP0001327', 'password': '874702467' }))
-}
--- a/coops-ui/src/main.ts
+++ b/coops-ui/src/main.ts
@@ -34,10 +34,12 @@ configHelper.fetchConfig()
   .then(() => {
     // ensure we have a Keycloak token
     if (!sessionStorage.getItem('KEYCLOAK_TOKEN')) {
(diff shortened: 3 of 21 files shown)


coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -22,23 +22,21 @@
         <template slot="label">
           <div class="certify-stmt">
             I, <b>{{trimmedCertifiedBy || '[Legal Name]'}}</b>, certify that I have relevant knowledge of the
-            association and that I am authorized to make this filing.
+            {{ entityDisplay || 'association' }} and that I am authorized to make this filing.
```
cameron-freshworks: Cool! If you plan on using `association` or any other label that is business approved verbiage, in other places you could always make it a constant.

coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -22,23 +22,21 @@
...
           </div>
         </template>
       </v-checkbox>
       <p class="certify-clause">{{currentDate}}</p>
       <p class="certify-clause">
-        Note: It is an offence to make a false or misleading statement in
-        respect of a material fact in a record submitted to the Corporate Registry for filing.
-        See section 200 of the Cooperatives Association Act.
+        {{message}}
```
cameron-freshworks: I like it. Not 100% sure but linters might want spaces in between the brackets and the template string.

coops-ui/src/components/AnnualReport/Certify.vue
```diff
@@ -22,23 +22,21 @@
...
       </p>
     </div>
   </v-card>
 </template>
 
 <script lang="ts">
 import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
-
+import { configJson } from '@/resources'
```
cameron-freshworks: Maybe just add a short comment above the imports like // Components  ...  // Enums   etc etc. Just to give a quick idea of what exactly the configJson is.

coops-ui/src/resources/config-objects.ts
```diff
@@ -0,0 +1,68 @@
+
+export class Business {
+     typeEnum: string;
+     displayName:String;
```
cameron-freshworks: just need a space here after the colon, and a lowercase s in `string`.

coops-ui/src/resources/config-objects.ts
```diff
@@ -0,0 +1,68 @@
...
+     certifyText:String;
+     // fields:Array<Fields>
+}
+
+export const configJson = [{
+  typeEnum: 'BC',
+  displayName: 'Benefits Company',
+  flows: [
+    {
+      feeCode: 'OTADD',
```
cameron-freshworks: We just recently introduced an enum for all the filing codes i believe, you could definitely use that for all the codes!

peter-freshworks:
> I feel that there is lot of duplication of strings in the resource json and it can be done in a better way by leveraging i18n support for vue and by using entity type prefixed strings as key for differing labels/ texts

I've looked into i18n and it looks like this JSON object I've used here should actually plug right in to that package (as the en-us locale lookup). The string duplication is not by accident, as technically there would only need to be a label for the different section codes and not the whole block of text, but keeping the entire message out of the code makes the code more concise and allows for a situation where a message for one entity is drastically different than another (or the scenario where there is no message, or the scenario where an entity has more than one message etc.). 

With that said, maybe I need to understand how the i18n package works better? From what I've seen it takes a javascript object with different locales and uses that as a lookup

leksmall:
> > I feel that there is lot of duplication of strings in the resource json and it can be done in a better way by leveraging i18n support for vue and by using entity type prefixed strings as key for differing labels/ texts
> 
> I've looked into i18n and it looks like this JSON object I've used here should actually plug right in to that package (as the en-us locale lookup). The string duplication is not by accident, as technically there would only need to be a label for the different section codes and not the whole block of text, but keeping the entire message out of the code makes the code more concise and allows for a situation where a message for one entity is drastically different than another (or the scenario where there is no message, or the scenario where an entity has more than one message etc.).
> 
> With that said, maybe I need to understand how the i18n package works better? From what I've seen it takes a javascript object with different locales and uses that as a lookup

All the messages will remain in the properties file itself. The common ones will not have a entity type prefix. Only the differing onces will be prefixed.

coops-ui/package.json
```diff
@@ -9,7 +9,8 @@
     "lint:nofix": "vue-cli-service lint --no-fix",
     "test:e2e": "vue-cli-service test:e2e",
     "test:unit": "vue-cli-service test:unit --testPathPattern=.",
-    "test:unit-silent": "vue-cli-service test:unit --silent"
+    "test:unit-silent": "vue-cli-service test:unit --silent",
+    "test:debug": "node --inspect-brk node_modules/.bin/vue-cli-service test:unit --testPathPattern --no-cache --watch --runInBand"
```
severinbeauvais: Could you please add a comment in coops-ui/README.md explaining what this does and how to use it?

coops-ui/src/components/AnnualReport/Certify.vue
severinbeauvais: I'm OK with the current code. It's legible and easy to understand :)

coops-ui/src/resources/config-objects.ts
severinbeauvais: Yes to all.

FYI, use `string` as a type to declare properties or `String` when you want to use the class prototype methods directly (eg, new String() or String.xxx).

coops-ui/src/resources/config-objects.ts
```diff
@@ -0,0 +1,68 @@
+
+export class Business {
+     typeEnum: string;
+     displayName:String;
+     flows: Array<Flow>;
+}
+
+export class Flow {
```
severinbeauvais: I know it doesn't matter in JS/TS that this class is declared after it's referenced above, but I still prefer to declare child classes first. Not a big deal though.

coops-ui/src/resources/config-objects.ts
```diff
@@ -0,0 +1,68 @@
+
```
severinbeauvais: I like that you've created a new folder for these things but I think the filename should be more specific to the type of resource... otherwise what would you call another resource (won't it also be a config object?)? Maybe "business-config.ts"?

coops-ui/src/views/AnnualReport.vue
```diff
@@ -159,6 +159,8 @@
                   :certifiedBy.sync="certifiedBy"
                   :currentDate="currentDate"
                   @valid="certifyFormValid=$event"
+                  :entityDisplay="this.configObject.displayName"
+                  :message="this.configObject.flows.find(x => x.feeCode === 'OTANN').certifyText"
```
severinbeauvais: I think the method to extract the data should be in a common module (similar to how we use store), so all components that need certifyText could call something like `getCertifyText(x.feeCode)`.

It also feels like this functionality is global/common/useful enough that any component in the app should be able to use it easily, possibly with a mixin? Or attached to Vue itself?

severinbeauvais:
In the title or description for this PR, please reference the ticket functionality that you are implementing (separate of how (POC) you implemented it).

Also, what will it take for this to no longer be a WIP?

Also, could you add some unit tests please?

coops-ui/src/components/AnnualReport/Certify.vue
peter-freshworks: Lint doesn't seem to complain on my system, but definitely let me know otherwise!

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -222,18 +224,17 @@ export default {
...
   },
 
   computed: {
     ...mapState(['currentDate', 'entityType', 'entityName', 'entityIncNo',
-      'entityFoundingDate', 'registeredAddress', 'recordsAddress']),
+      'entityFoundingDate', 'registeredAddress', 'recordsAddress', 'configObject']),
     ...mapGetters(['isRoleStaff']),
 
-    validated () {
+    validated () {
```
severinbeauvais: Sometimes this causes a lint error. Please check with `npm run lint:nofix`.

coops-ui/src/main.ts
```diff
@@ -35,6 +35,7 @@ configHelper.fetchConfig()
     // ensure we have a Keycloak token
     if (!sessionStorage.getItem('KEYCLOAK_TOKEN')) {
       console.log('Redirecting to Auth URL...')
+      sessionStorage.setItem('REDIRECTED', 'true')
```
severinbeauvais: This is old code that was removed a while back. Did you restore it intentionally?

peter-freshworks: no, I must have forgot to remove this, or I didn't see it in the rebase diff...my mistake

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
peter-freshworks: done

severinbeauvais:
> ~~In the title or description for this PR, please reference the ticket functionality that you are implementing (separate of how (POC) you implemented it), ie something to do with the certify component.~~ done
> 
> ~~Also, what will it take for this to no longer be a WIP?~~ PR is now complete
> 
> ~~Also, could you add some unit tests please?~~ done

coops-ui/src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -222,15 +224,14 @@ export default {
...
-
       // EntityTypes Enum
       EntityTypes
     }
   },
 
   computed: {
     ...mapState(['currentDate', 'entityType', 'entityName', 'entityIncNo',
-      'entityFoundingDate', 'registeredAddress', 'recordsAddress']),
+      'entityFoundingDate', 'registeredAddress', 'recordsAddress', 'configObject']),
```
severinbeauvais: This isn't necessary any more, right?

coops-ui/src/mixins/resource-lookup-mixin.ts
```diff
@@ -0,0 +1,41 @@
+import { Component, Vue } from 'vue-property-decorator'
+import { mapState } from 'vuex'
+
+/**
+ * Mixin that provides an entity filter utility.
```
severinbeauvais: Please update comment.

coops-ui/src/mixins/resource-lookup-mixin.ts
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Method to return certify message using the configuration lookup object.
+     *
+     * @param entity The entity type of the component.
+     * @return the appropriate message for the certify component for the current filing flow.
+     */
+    certifyText (feeCode: string): string {
+      if (this.configObject && this.configObject.flows) {
+        const flow = this.configObject.flows.find(x => x.feeCode === feeCode)
+        return flow.certifyText
```
severinbeauvais: What if `flow` is not found (null)? Should check this before dereferencing it.

coops-ui/src/mixins/resource-lookup-mixin.ts
peter-freshworks: d'oh ... good catch!

coops-ui/src/resources/business-config.ts
```diff
@@ -0,0 +1,66 @@
+export class Flow {
+  feeCode: string
+  displayName: string
+  certifyText: string
+}
+
+export class Business {
+     typeEnum: string
+     displayName: string
+     flows: Array<Flow>
```
cameron-freshworks: Looks like there is an extra tab here.

coops-ui/src/resources/business-config.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export const configJson = [{
+  typeEnum: 'BC',
+  displayName: 'Benefits Company',
+  flows: [
+    {
+      feeCode: 'OTADD',
```
cameron-freshworks: Did you want to use the fee code enum in place of these strings?

coops-ui/src/resources/business-config.ts
```diff
@@ -0,0 +1,66 @@
...
+      feeCode: 'OTCDR',
+      displayName: 'Change Of Directors',
+      certifyText: 'Note: It is an offence to make a false or misleading statement in' +
+           'respect of a material fact in a record submitted to the Corporate Registry for filing. ' +
+           'See section 127 of the Business Corporations Act.'
+    }
+  ]
+},
+{
+  typeEnum: 'CP',
```
cameron-freshworks: Same as above, i think we have an enum for entity types. Does it make sense to them in the config files?

coops-ui/src/views/AnnualReport.vue
```diff
@@ -163,6 +163,8 @@
                 :certifiedBy.sync="certifiedBy"
                 :currentDate="currentDate"
                 @valid="certifyFormValid=$event"
+                :entityDisplay="displayName()"
+                :message="certifyText('OTANN')"
```
cameron-freshworks: Maybe another spot we could use the filingCodes enum?

coops-ui/src/resources/business-config.ts
peter-freshworks: ah interesting...lint didn't pick this up, good catch!

coops-ui/src/resources/business-config.ts
peter-freshworks: you would think so, but keeping in mind this is meant to be an external resource and has no references to any of the front end code. I.E - This could also live in a plain text/json file, or in a database...think of it more as a table than application source code, as it has no dependency on any other source code (front or back end)

coops-ui/src/resources/business-config.ts
peter-freshworks: similar to my other comment :)

coops-ui/src/views/AnnualReport.vue
peter-freshworks: In this case however, yes it does make sense. Good catch!

severinbeauvais:
Final tweaks look good... Merging this.

PR #599, opened 2019-12-10
Annual Report and Change of Address Report fixes

leksmall:
*Issue #:* /bcgov/entity#2078

*Description of changes:*
Annual Report and Change of Address Report fixes

### DIFF
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -128,18 +128,27 @@ class Report:  # pylint: disable=too-few-public-methods
     def _get_template_filename(self):
         return '{}.html'.format(self._filing.filing_type)
 
-    def _get_template_data(self):
+    def _get_template_data(self):  # pylint: disable=too-many-branches
         filing = copy.deepcopy(self._filing.filing_json['filing'])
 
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/reports/report.py
```diff
@@ -128,18 +128,27 @@ def _get_environment():
...
+                if filing.get('annualReport', {}).get('deliveryAddress'):
+                    filing['registeredOfficeAddress'] = {
+                        'deliveryAddress': filing['annualReport']['deliveryAddress'],
+                        'mailingAddress': filing['annualReport']['mailingAddress']
+                    }
+                else:
+                    filing['registeredOfficeAddress'] = {
+                        'deliveryAddress': filing['annualReport']['offices']['registeredOffice']['deliveryAddress'],
+                        'mailingAddress': filing['annualReport']['offices']['registeredOffice']['mailingAddress']
+                    }
```
severinbeauvais: Does the same thing need to be done for Records Addresses?

leksmall: Not at this point

peter-freshworks: it will be done once the ticket is picked up for outputs, which should be soon since the designs are now done

PR #602, opened 2019-12-12
added IDs/classes for QA testing

severinbeauvais:
*Issue #:* /bcgov/entity#2042

*Description of changes:*
- added 'loading' indicators to filing cancel buttons (fix)
- moved MDI font from dev to prod dependencies (fix)
- updated preload font name so it works (fix)
- added ids to unique elements
- added classnames to repeated elements
- added ids to all dialogs
- attached dialogs to their components
- renamed some ids and classes for consistency
- fixed font color issue
- fixed return val error in hasTasks()
- improved COD edit/summary transitions
- added loading messages to COD and COA
- ensured all button texts have no extra line breaks (block vs inline elements)
- updated/fixed unit tests accordingly
- added Entity Status enum


coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -876,10 +881,11 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
...
    */
-  private showDeleteDirectorConfirmation (director): void {
-    this.showPopup = true
-    this.activeDirectorToDelete = director
-  }
+  // removed until release 2
+  // private showDeleteDirectorConfirmation (director): void {
+  //   this.showPopup = true
+  //   this.activeDirectorToDelete = director
+  // }
```
cameron-freshworks: Just to clarify, what is the state of the commented out code in the Directors component. Is it something that is in the works in the near future?

coops-ui/src/components/Dashboard/DirectorListSm.vue
```diff
@@ -1,57 +1,64 @@
...
+                                              {{ director.deliveryAddress.addressRegion }}
+                                              {{ director.deliveryAddress.postalCode }}</li>
+                    <li class="address-line4">{{ getCountryName(director.deliveryAddress.addressCountry) }}</li>
+                  </ul>
+                </v-list-item-subtitle>
+              </v-list-item-content>
+            </v-list-item>
+
+            <v-list-item class="mailing-address-list-item"
+              v-if="director.mailingAddress"
```
cameron-freshworks: This is a good change 👍

coops-ui/src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1,56 +1,77 @@
 <template>
```
cameron-freshworks: We might have some conflicts in this file with my most recent PR as well, though my changes were quite minimal. So make sure yours goes in first :)

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -1,47 +1,55 @@
...
+        <v-expansion-panel-header class="todo-item-toggle no-dropdown">
           <div class="list-item">
-            <div class="filing-type">
+            <div class="todo-label">
               <div class="list-item__title">{{item.title}}</div>
+
               <div class="bcorps-ar-subtitle"
-                 v-if="entityFilter(EntityTypes.BCORP) &&
-                 isConfirmEnabled(item.type, item.status)"
+                v-if="entityFilter(EntityTypes.BCORP) && isConfirmEnabled(item.type, item.status)"
```
cameron-freshworks: I like putting everything on one line as well, but sometimes i find myself breaking up the lines for consistency with other components when the properties get a little lengthy. Especially on those components that use a lot of small props built into the component. What is our rule of thumb, when to break it up into multiple lines? Use the limit set by Linting?

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -1,47 +1,55 @@
...
               >
                 <p>Verify your Office Address and Current Directors before filing your Annual Report.</p>
                 <v-checkbox
-                  id="confirm-checkbox"
                   class="todo-list-checkbox"
                   label="All information about the Office Addresses and Current Directors is correct."
                   :disabled=!item.enabled
-                  v-model=confirmCheckbox
+                  v-model="confirmCheckbox"
```
cameron-freshworks: Is it always correct to wrap these properties as a string? I know we can pass varying data types, in this case we're passing a boolean, so just curious as to why we are wrapping this?

coops-ui/src/components/EntityInfo.vue
```diff
@@ -1,11 +1,14 @@
...
-        <div class="entity-name mb-1">{{ entityName || 'Not Available' }}</div>
-        <v-chip class="entity-status" small label text-color="white" v-if="entityStatus"
+        <div class="mb-1" id="entity-legal-name">
+          <span>{{ entityName || 'Not Available' }}</span>
+        </div>
+        <v-chip id="entity-status" small label text-color="white"
+          v-if="entityStatus"
           :class="{
             'blue' : entityStatus === 'GOODSTANDING',
             'red' : entityStatus === 'PENDINGDISSOLUTION' | 'NOTINCOMPLIANCE',
```
cameron-freshworks: I know you didn't touch this code in this PR but possibly a good spot for us to refactor into some Constants or an entityStatus Enum in the future :)

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: I'm not certain about future plans for the Directors component, although I've overhead "custom appointment/cessation dates" and some of the commented-out code refers to that as well.
I don't know what else to say here. I don't want to delete the work done by another developer even though it's not currently in use. I commented out the method above because it was called from commented-out template code, and I didn't want to pretend that it is currently valid code.

coops-ui/src/components/EntityInfo.vue
```diff
@@ -108,11 +119,11 @@ export default class EntityInfo extends Vue {
...
-.entity-info {
+#entity-info {
   background: $BCgovInputBG;
 }
 
-// .entity-info.staff {
+// #entity-info.staff {
 //   background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='105' height='100'><text x='0' y='105' font-size='30' transform='rotate(-45 10,40)' opacity='0.1'>STAFF</text></svg>");
 //   background-repeat: repeat-x;
 // }
```
cameron-freshworks: Is this for a future feature as well?

coops-ui/src/components/Dashboard/TodoList.vue
severinbeauvais: Personally I don't like very long lines because I often have 2 windows side by side and I don't like scrolling horizontally to see everything :>
I also don't like _every single attribute_ on separate lines.
So, like you, I strive for consistency within a component. There are a few other things I consider, such as:
- if possible, move style attributes (such as "outlined") close to the tag, so it can be read as "outline button"
- if possible, move class close to the tag (some devs will say class should always be first, but I have lots of personal exceptions)
- if possible, move logic (v-if, v-show) close to the tag name, as it makes reviewing the code easier, especially when folding blocks
- if possible, move looping (v-for) attribute before other attributes that use the looping variables (`:class`, `:key`, `:id`, etc)

Also, you may have seen that I added spans around some standalone text. This is because most text is an inline element (vs block) and I don't want extra whitespace around it (even though whitespace is usually removed). Consider this template:
```
<v-btn>
  Click me
</v-btn>
```
which renders as:
```
<button>
  <span>
    " Click me "
  </span>
</button>
```
versus
```
<v-btn>Click me</v-btn>

<v-btn>
  <span>Click me too</span>
</v-btn>
```
which renders as:
```
<button>
  <span>
    "Click me"
  </span>
</button>

<button>
  <span>
    <span>Click me too</span>
  </span>
</button>
```
In the latter example, you can select the button text and it won't contain extra spaces.

coops-ui/src/components/Dashboard/TodoList.vue
severinbeauvais: I think there are exceptions but more attribute values should be in quotes. In this case I checked the other places we used v-model and they all had quotes. I changed this and tested it and it seemed to work fine.

coops-ui/src/components/EntityInfo.vue
severinbeauvais: Sorta. If you uncomment this then you get a nice effect when the user is staff. I did this when I was developing staff roles.

On other projects we always had an indication when a staff was logged in and doing something on behalf of a user, so while it's not currently a requirement for us, it may one day become one, and this is my contribution to it :)

coops-ui/src/components/EntityInfo.vue
severinbeauvais: Agreed and implemented in a separate commit below.

coops-ui/src/components/AnnualReport/Directors.vue
cameron-freshworks: Fair enough, thanks!

coops-ui/src/components/Dashboard/TodoList.vue
cameron-freshworks: wow, thanks Severin! That was a really informative and helpful comment, appreciate that. I think i will extract this little bit of information and will certainly reference it when i'm faced with another 'To line break or not to line break' dilemma. Thanks!

coops-ui/src/components/Dashboard/TodoList.vue
cameron-freshworks: I definitely agree that based on consistency quotes should be used, I will try and make sure i follow the same pattern as well, Thanks for the comment!

coops-ui/src/components/EntityInfo.vue
cameron-freshworks: Great! Nice job 👍

coops-ui/src/components/EntityInfo.vue
cameron-freshworks: Okie doke, sounds good!

coops-ui/src/components/EntityInfo.vue
```diff
@@ -1,18 +1,21 @@
...
-            'red' : entityStatus === 'PENDINGDISSOLUTION' | 'NOTINCOMPLIANCE',
+            'blue' : isGoodStanding,
+            'red' : isPendingDissolution || isNotInCompliance
           }">
-          <span v-if="entityStatus === 'GOODSTANDING'">In Good Standing</span>
-          <span v-else-if="entityStatus === 'PENDINGDISSOLUTION'">Pending Dissolution</span>
-          <span v-else-if="entityStatus === 'NOTINCOMPLIANCE'">Not in Compliance</span>
+          <span v-if="isGoodStanding">In Good Standing</span>
+          <span v-else-if="isPendingDissolution">Pending Dissolution</span>
+          <span v-else-if="isNotInCompliance">Not in Compliance</span>
```
cameron-freshworks: Nice job, looks good!

coops-ui/src/components/EntityInfo.vue
```diff
@@ -58,6 +69,7 @@
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator'
 import { mapState, mapGetters } from 'vuex'
+import { EntityStatus } from '@/enums'
```
cameron-freshworks: Should we add comments here for import consistency? ie // Enums , // Libraries ?  I started doing that before, not sure if was something you prefer though, thoughts on that?

coops-ui/src/components/dialogs/DashboardUnavailableDialog.vue
```diff
@@ -43,6 +42,9 @@ export default class DashboardUnavailableDialog extends Vue {
   // Prop to display the dialog.
   @Prop() private dialog: boolean
 
+  // Prop to provide attachment selector.
+  @Prop() private attach: string
```
cameron-freshworks: I don't fully understand the usage of this prop here and in the other dialogs. This is purely a question for my own knowledge, could i get the jist of why we pass this prop down? Is this to give the parent a reference to the dialogs real content, opposed to a just a reference to the existence of the child on the parent?

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -673,7 +672,7 @@ describe('Appoint New Director tests', () => {
     setTimeout(() => {
       expect(vm.$el.querySelectorAll('.v-messages')[0].textContent).toBe('')
       done()
-    }, 200)
+    }, 250)
```
cameron-freshworks: I know i've asked before and i'm probably a pain in the butt here, but what defines the time required in these? Will Async and await not do the trick in these cases?

coops-ui/src/components/EntityInfo.vue
severinbeauvais: Yes, I liked that you did that in other components that imported many things, but this component only has 3 imports and I thought the code was self-documenting :)

coops-ui/src/components/dialogs/DashboardUnavailableDialog.vue
severinbeauvais: Good question!

Although dialogs seem to live within a component, in fact the only thing there in the DOM is a stub. The actual dialog code/content is normally rendered dynamically right under `#app` (outside of the component). This is why it's hard to test dialogs, tooltips and similar.

But there exists the "attach" attribute, which causes the dialog content to now be dynamically rendered under its parent component. Now it can be selected easily! I did this in case QA wants to test any dialog.

coops-ui/tests/unit/Directors.spec.ts
severinbeauvais: We have not been able to get these working with async/await or flushPromises. I dunno why. (Guess: there's debouncing happening on the input?)

When these tests fail when I run them locally, increasing the timeouts usually works.

It bugs me but I had a quick look and didn't find a solution. You're welcome to try :)

coops-ui/src/components/EntityInfo.vue
cameron-freshworks: Makes sense, it certainly is!

coops-ui/src/components/dialogs/DashboardUnavailableDialog.vue
cameron-freshworks: Great explanation. I'm just as happy as QA with this discovery as well. This will only improve and make our unit testing easier. well done sir!

coops-ui/tests/unit/Directors.spec.ts
cameron-freshworks: Sounds good, working tests are whats important, so we can always revisit this when we have time!

PR #607, opened 2019-12-16
Cancel Payment functionality

leksmall:
*Issue #:* /bcgov/entity#1617, /bcgov/entity#2056, /bcgov/entity#2057, /bcgov/entity#1864

*Description of changes:*
Added/Modified UI components
Added Patch filing Endpoint to the legal API that integrates with Pay API cancel payment endpoint
Modified the payment token setter to  discard the locked filing check
Added UI tests
Added unit tests for the new patch end point
Modified the existing tests for payment token setter to exclude the lock check
Modified dashboard test to ignore the launch darkly warning


coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -5,6 +5,11 @@
       attach="#todo-list"
     />
 
+    <ConfirmDialog
+      ref="confirmCancelPaymentDialog"
+      attach="#todo-list" data-test-id="cancel-payment-confirm-dialog"
```
severinbeauvais: Please add a `data-test-id` to the other instance of this dialog (above) to be able to select one or the other.

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -13,6 +18,14 @@
       attach="#todo-list"
     />
 
+    <CancelPaymentErrorDialog
+      :dialog="cancelPaymentErrorDialog"
+      :errors="cancelPaymentErrors"
+      @okay="resetCancelPaymentErrors"
+      attach="#todo-list"
+      data-test-id="cancel-pay-error-dialog"
```
severinbeauvais: In this case we don't need `data-test-id` because there is a single instance of this dialog in this component so we can already uniquely select it.

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -131,14 +144,32 @@
...
-                  @click.native.stop="doResumePayment(item)"
-                >
-                  <span>Resume Payment</span>
-                </v-btn>
+                 <template v-else-if="isPending(item)">
+                  <v-btn class="btn-resume-payment"
+                    color="primary"
+                    :disabled="!item.enabled"
+                    @click.native.stop="doResumePayment(item)"
+                    data-test-id="btn-resume-payment">
```
severinbeauvais: We don't need `data-test-id` because we can already select this button via its class.
(Also, data-test-id might not be unique if there are several pending items.)

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -131,14 +144,32 @@
...
+                    <span>Resume Payment</span>
+                    <!-- Cancel Payment -->
+                  </v-btn>
+                   <v-menu offset-y left>
+                    <template v-slot:activator="{ on }">
+                      <v-btn color="primary"
+                        v-on="on" id="pending-item-menu-activator" :disabled="!item.enabled"
+                        class="actions__more-actions__btn px-0"
+                      @click.native.stop  data-test-id="btn-pending-filing-menu">
```
severinbeauvais: We don't need 'data-test-id' because we can already select this button via its id.

coops-ui/src/components/dialogs/CancelPaymentErrorDialog.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <v-dialog v-model="dialog" width="45rem" persistent :attach="attach">
```
severinbeauvais: Please add a `content-class` attribute to select this dialog.

Also please remove all `data-test-id` below where we already have an element id.

coops-ui/src/components/Dashboard/TodoList.vue
leksmall: As per John, it does not matter if a c omponent has id or not. The scripts will be looking at data-test-id and it should be there even if id is there so that the test script can be consistent

coops-ui/tests/unit/TodoList.spec.ts
```diff
@@ -1527,3 +1542,165 @@ describe('TodoList - Delete Draft', () => {
...
+
+    Vue.nextTick(async () => {
+      const button = wrapper.find('#pending-item-menu-activator')
+      await button.trigger('click')
+      const button1 = wrapper.find('#btn-cancel-payment')
+      await button1.trigger('click')
+      // verify confirmation popup is showing
+      expect(wrapper.vm.$refs.confirmCancelPaymentDialog).toBeTruthy()
+
+      await flushPromises()
```
severinbeauvais: This serves no purpose since you're not testing anything after it :)

coops-ui/src/components/dialogs/CancelPaymentErrorDialog.vue
leksmall: data-test-ids need another discussion with QA

coops-ui/src/components/Dashboard/TodoList.vue
severinbeauvais: OK, we will discuss this and decide what's best for everyone.

legal-api/src/legal_api/models/filing.py
```diff
@@ -110,8 +110,6 @@ def payment_token(self):
 
     @payment_token.setter
     def payment_token(self, token: int):
```
thorwolpert: as discussed, turn this back on and have the reset method alter the self._payment_token

leksmall: Done :)

coops-ui/src/components/dialogs/CancelPaymentErrorDialog.vue
leksmall: > I have made comments regarding "data-test-id" attributes but we need to discuss this with QA to see if my comments are valid. Otherwise this looks good to me.

Based on the discussion this morning data-test-id named attributes must be present for UI elements.

coops-ui/src/components/Dashboard/TodoList.vue
severinbeauvais: In addition to our discussion this morning, I chatted with Roland Stearns, the Senior Tester we have for a bit. He said to do no extra work while he reviews whether existing selectors are sufficient. In other words, just leave this here for now.

coops-ui/src/components/dialogs/CancelPaymentErrorDialog.vue
severinbeauvais: Please add a `content-class`, which is attached to the actual dialog.
Do not change what you've done for `data-test-id`.

coops-ui/src/components/Dashboard/TodoList.vue
severinbeauvais: FYI, adding `data-test-id` here can be used to select the stub div for the dialog (which contains nothing).
Instead, adding `content-class` to the v-dialog can be used to select the actual div that contains the dialog elements.
Based on this, I think adding data-test-id here is not useful.

severinbeauvais:
I'm waiting for final changes, or a comment that nothing will be changed, before accepting this review and merging this PR.

PR #618, opened 2019-12-18
#2167 - Coops Pending State Bug

cameron-freshworks:
*Issue #:* /bcgov/entity#2167

*Description of changes:*

* Fixed a front end bug where the if the state of a change of addresses filing for a COOPS was not changed from 'Paid' to 'Completed', the Pending Filing state would appear for that COOP. 

* UI now correctly verifies that the entity is a BCORP.

* This is only a fail safe, as the back end should handle the state and if the state of a Change of Address filing ( future effective ) is not filed and changed to completed, there may be a back end fix required as well. 

* Updated Unit tests and removed a duplicate test.


coops-ui/src/views/Dashboard.vue
```diff
@@ -245,7 +245,8 @@ export default {
      */
     checkPendingFilings (filings) {
       filings.forEach(filing => {
-        if (filing.status === FilingStatus.PAID) {
+        if (this.entityFilter(EntityTypes.BCORP) &&
+          filing.status === FilingStatus.PAID) {
           this.effectiveDate = filing.filingEffectiveDate
           this.coaPending = true
```
leksmall: Since the coaPending is set here should we check whether the filing type is change of address?

A filing can be in a paid status if the filer is down for some reason. It can be for any type of filing. In that scenario I am not sure whether caution icon will show up next to another filing type.

cameron-freshworks: You're 100% right, i will add that validation in as well and update the tests. Without there is a potential that a BCOMPs Annual report or Director change could trigger this state. Thanks!

coops-ui/src/views/Dashboard.vue
```diff
@@ -239,13 +239,15 @@ export default {
...
-     * Currently only BCOMPS have future effective filings.
+     * Currently only BCOMPs change of addresses filing are future effective.
      *
      * @param filings The array of filings in history
      */
     checkPendingFilings (filings) {
       filings.forEach(filing => {
-        if (filing.status === FilingStatus.PAID) {
+        if (this.entityFilter(EntityTypes.BCORP) &&
+          filing.name === 'Address Change' &&
```
severinbeauvais: I know these are assigned in FilingHistoryList.vue:loadData() -- ie, not your code -- but could you please go back and create an enum to both assign the name and compare it here?

coops-ui/src/views/Dashboard.vue
```diff
@@ -239,13 +239,15 @@ export default {
      * Searches the filings history for a 'paid' status.
      * Paid status indicates a filing that is paid but future effective.
      * Change the state of the UI when a filing is future effective.
-     * Currently only BCOMPS have future effective filings.
+     * Currently only BCOMPs change of addresses filing are future effective.
```
severinbeauvais: I recommend replacing all BCORP(s) with BCOMP(s) now, to avoid confusion in the code :)

coops-ui/src/views/Dashboard.vue
cameron-freshworks: Great idea, i considered this as well. I'm guilty of not being pro-active on this one :) Lets do it!

coops-ui/src/views/Dashboard.vue
cameron-freshworks: Good idea, was thinking this as well. When did we officially change from Bcorps to Bcomps.. or perhaps Bcorps was wrong all along 😆

cameron-freshworks:
> See my suggestions on improving existing code.
> Approved.

Updated! Might want to test it locally, as the name change affected a lot of files 😄

severinbeauvais:
> > See my suggestions on improving existing code.
> > Approved.
> 
> Updated! Might want to double check it locally, as the name change affected a lot of files 😄

I can find no remaining references to "bcorp" except in a comment, an interface name, a folder name, and some class names. You can get those next time.

severinbeauvais:
I've confirmed that linting is OK and all unit tests pass.

PR #624, opened 2019-12-20
DEVOPS - RC1

bsnopek-freshworks:
- Updated deployment configs and params to build all pods together
- Updated deployment-objects for inclusion of services, routes, configmaps, etc
- Updated to connect to EDB - although there is still an issue punching through load-balancer.


devops/openshift/_entity-filer.dc.json
```diff
@@ -4,18 +4,225 @@
     "metadata": {
         "annotations": {
             "description": "Deployment template for a entity-filer service.",
-            "tags": "${NAME}-${TAG_NAME}"
+            "tags": "${COMP_NAME"
```
kialj876: missed a bracket here

devops/openshift/_legal-api.dc.json
```diff
@@ -109,23 +455,22 @@
                         "type": "ConfigChange"
                     }
                 ],
-                "replicas": "${REPLICAS}",
+                "replicas": "${{REPLICA_MIN}}",
```
kialj876: Double brackets!

devops/openshift/_legal-api.dc.json
```diff
@@ -109,23 +455,22 @@
...
-                            "app": "${COMP_NAME}-${SUFFIX}",
-                            "deploymentconfig": "${COMP_NAME}-${SUFFIX}"
+                            "name": "${COMP_NAME}-${SUFFIX}"
                         }
                     },
                     "spec": {
                         "containers": [{
                             "name": "${COMP_NAME}",
-                            "image": "docker-registry.default.svc:5000/${IMAGE_NAMESPACE}/${COMP_NAME}:${SUFFIX}",
+                            "image": " ",
```
kialj876: Is this on purpose? Will it create the container without an image somehow?

devops/openshift/_entity-filer.dc.json
bsnopek-freshworks: Good catch ! Thanks

devops/openshift/_legal-api.dc.json
bsnopek-freshworks: Double braces allow for input of non-string values, the quotes will get replaced with the integer value. 

https://docs.openshift.com/container-platform/3.11/dev_guide/templates.html#writing-parameters

devops/openshift/_legal-api.dc.json
bsnopek-freshworks: This is on purpose yes - the groovy script that actually facilitates the deployment populates that value with properties derived from other variables.

rstens:
No other comments, but I would like a in person walk-through and discussion on this approach.

PR #628, opened 2020-01-01
added business rules for effective_date and directors delivery addresses, also added a replacement for datetime for a non-naive utcnow

thorwolpert:
Signed-off-by: thor wolpert <thor@wolpert.ca>

*Issue #:* /bcgov/entity#1580

*Description of changes:*
- added a datetime replacement that has a non-naive utcnow() (aka it has a tzinfo section)
- implemented business rules for
  - effective_date, not future effective or before any previous COD legal filing
  - delivery addresses for Directors, must be in BC & Canada


legal-api/src/legal_api/models/filing.py
```diff
@@ -292,6 +293,33 @@ def get_a_businesses_most_recent_filing_of_a_type(business_id: int, filing_type:
...
+
+        # As the JSON query is new for most, leaving the debug stmnt
+        # that dumps the query for easier debugging.
+        current_app.logger.debug(
+            str(filing.statement.compile(
+                dialect=dialect(),
+                compile_kwargs={'literal_binds': True}))
+        )
+
+        return filing.one_or_none()
```
katiemcgoff: In my initial test, I get a 500 error from this upon posting. I haven't dug further as to why, but the corpnum in dev is 1523.

thorwolpert: I think the DB you're using hasn't been migrated. One of the previous commits has the migration that creates the **offices** table. There's also a migration script that updates the data that Peter added.

kialj876: Yeah it's working for me. There's some failing postman tests due to the office table/schema change, but I'll update those now and add them to my PR. My PR includes fixes to the pytests that are failing due to it being 2020 so theres an extra task in the todos

katiemcgoff: Re-opening this - the error I'm getting, over dev db with CP0001523, is sqlalchemy.orm.exc.MultipleResultsFound on the filing.one_or_none. Checked the query from the debug statement in lile 315, and it returns 3 results, hence the error - one COD and two ARs. I haven't tested yet with other corps but this would at least break for this corp, in dev, if it's promoted.

PR #629, opened 2020-01-03
1147 colin updater

kialj876:
*Issue #:* /bcgov/entity#1147

*Description of changes:*
- legal:
    - added source col in filings table
        - auto populates on before_insert event based on user id
        - migration will set column value for all existing filings
        - tests updated
    - user model updated with find_by_id method
    - internal filings get method updated
- colin:
    - added ledger text for annual reports and also enters user id
    - updated failing pm test
- colin-updater:
    - updated code to work with current apis


jobs/update-colin-filings/update_colin_filings.py
```diff
@@ -86,18 +86,18 @@ def get_filings(app: Flask = None):
     return r.json()
 
 
-def send_filing(app: Flask = None, filing: dict = None):
+def send_filing(app: Flask = None, filing: dict = None, filing_id: str = None):
     """Post to colin-api with filing."""
-    clean_none(filing)
+    # clean_none(filing)
```
katiemcgoff: Is this supposed to be commented out?

legal-api/src/legal_api/models/user.py
```diff
@@ -40,6 +40,11 @@ class User(db.Model):
     iss = db.Column(db.String(1024))
     creation_date = db.Column(db.DateTime(timezone=True), default=datetime.utcnow)
 
+    @classmethod
+    def find_by_id(cls, submitter_id: int = None):
+        """Return a User if they exist and match the provided JWT."""
+        return cls.query.filter_by(id=submitter_id).one_or_none()
+
```
katiemcgoff: Your comment said it checks against JWT - is that true? I don't see that.

jobs/update-colin-filings/update_colin_filings.py
kialj876: Yep! Commenting it out because I don't think it's needed anymore, but I didn't want to delete the code in case we need it for a use case I'm not thinking of

legal-api/src/legal_api/models/user.py
kialj876: Oh oops that must have been a copy paste error. I'll update the comment!

PR #637, opened 2020-01-07
1448 directors warning

peter-freshworks:
*Issue #:* /bcgov/entity#1448

*Description of changes:*
Add generic warnings for out-of-compliance messages for directors. Text can change based on entity type and warning(s) encountered, so used the config-object for this. Warnings not pertaining to number of directors are out of scope.


coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -536,6 +561,9 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
   private activeDirectorToDelete = null
   private cessationDateTemp = null
   private isEditingDirector = false
+  private isCompliant = true
```
severinbeauvais: It looks like this isn't used for anything...?

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -981,6 +1019,11 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
     // close standalone cessation date picker and reset date
     this.cessationDateTemp = null
     this.activeIndexCustomCease = null
+    console.log(this.directors)
```
severinbeauvais: Please remove extraneous logs in final code.

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -981,6 +1019,11 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
...
+    if (this.directors.filter(x => x.cessationDate == null).length < 2) {
```
severinbeauvais: I don't understand what this code is doing... looking for directors that aren't ceased?

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -523,7 +546,9 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
   @Prop({ default: true })
   private componentEnabled: boolean
 
-  private directorEditInProgress:boolean = false;
+  private directorEditInProgress: boolean = false;
+
+  private lastValidDirector: number = -1
```
severinbeauvais: It looks like this isn't used for anything...?

coops-ui/src/components/dialogs/index.ts
```diff
@@ -8,7 +8,7 @@ import PaymentErrorDialog from './PaymentErrorDialog.vue'
 import ResumeErrorDialog from './ResumeErrorDialog.vue'
 import SaveErrorDialog from './SaveErrorDialog.vue'
 import CancelPaymentErrorDialog from './CancelPaymentErrorDialog.vue'
-
+import WarningPopover from './WarningPopover.vue'
```
severinbeauvais: Please restore blank line between code blocks (for easier reviewing :) ).

coops-ui/src/mixins/resource-lookup-mixin.ts
```diff
@@ -40,4 +40,40 @@ export default class ResourceLookupMixin extends Vue {
...
+    /**
+     * Method to validate directors on edit/cease and return any warning messages.
+     *
+     * @return the compliance message or null (if the configuration has been loaded).
+     */
+    directorWarning (directors: Array<any>): Object {
+      const configSection = this.configObject.flows.find(x => x.feeCode === 'OTCDR').warnings
+      let errors = []
+      // FUTURE: Too much code for this. Can be condensed and made more reusable.
+      if (configSection.bcResident != null) {
```
severinbeauvais: Would it be better to simply check is bcResident exists? Then you wouldn't have to declare a null property in the config file where it isn't needed (ie, it would be optional).

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -181,7 +182,14 @@
                 <header>
                   <h1 id="filing-header-review">Review: Director Change </h1>
                 </header>
-
+                <section v-if="complianceDialogMsg">
+                  <v-alert type="info" outlined
+                  icon="mdi-information"
+                  class="white-background"
+                  >
+                    <p style="font-size:16px; color:#495057">{{ complianceDialogMsg.msg }}</p>
```
severinbeauvais: Please define a new class for these styles.

coops-ui/src/components/dialogs/WarningPopover.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <v-menu v-if="dialogObj != null"
+    :close-on-content-click="false"
+    :nudge-width="200"
+    offset-y
+    max-width="18rem"
+  >
+    <template v-slot:activator="{ on }">
+      <span class="warning-text" v-on="on">
+        <v-icon style="vertical-align: bottom; color:#2196F3">mdi-information-outline</v-icon>{{ dialogObj.title }}
```
severinbeauvais: Please define a new class for these styles.

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -981,6 +1019,11 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
     // close standalone cessation date picker and reset date
     this.cessationDateTemp = null
     this.activeIndexCustomCease = null
+    console.log(this.directors)
+    if (this.directors.filter(x => x.cessationDate == null).length < 2) {
+      this.lastValidDirector = director.id
+    }
+    // Determine if there are enough directors to remain in compliance
```
severinbeauvais: Obsolete comment or incomplete code?

coops-ui/src/components/AnnualReport/Directors.vue
peter-freshworks: that's weird...I thought I removed this, along with the the logging and some other obsolete code referenced in comments...I shall look into this!

coops-ui/src/components/AnnualReport/Directors.vue
peter-freshworks: well, it shouldn't even be here, so my mistake...looking into it...possibly missed a "push"

coops-ui/src/components/AnnualReport/Directors.vue
peter-freshworks: not even used I should have deleted these 4 lines entirely

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -467,6 +476,20 @@
...
+          <v-alert
+          close-text="Close Alert"
+          dismissible
+          icon="mdi-information-outline"
+          class="white-background"
+          :id="'director-' + director.id + '-alert'"
+          v-if="complianceMsg && index == messageIndex"
+          v-once
+          >
+            <div class="complianceCard">
```
severinbeauvais: This isn't technically a "card" in Vuetify speak (contrast this with an actual "v-card" element). Please rename it.

coops-ui/src/components/dialogs/WarningPopover.vue
```diff
@@ -0,0 +1,44 @@
...
+         {{ dialogObj.msg }}
+        </v-list-item-content>
+      </v-list-item>
+    </v-card>
+  </v-menu>
+</template>
+
+<script lang="ts">
+import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
+import { AlertMessage } from '../../interfaces'
```
severinbeauvais: This would be better referenced as '@/interfaces'.

(@ = project-folder/src)

coops-ui/src/components/dialogs/WarningPopover.vue
```diff
@@ -0,0 +1,44 @@
...
+</script>
+
+<style lang="scss">
+.warning-text {
+  font-weight: bold;
+  font-size: 14px;
+  color: #2196F3;
+    .v-icon {
+    vertical-align: bottom !important;
+    color:#2196F3 !important;
```
severinbeauvais: Guideline: Use colour names when available (in this case, "Dodger Blue"). Also on line 38.

Ideal: Use CSS variables from a defined app theme (see theme.scss). This allows reuse as well as centralized app colours (ie, theme).

Recommendation: Add this colour to the theme file and use the CSS variable here. You'll need to add `@import '@/assets/styles/theme.scss';` at the top of this section.

coops-ui/src/interfaces/informationAlert-interfaces.ts
```diff
@@ -0,0 +1,7 @@
+import Vue from 'vue'
+
+export interface AlertMessage extends Vue {
```
severinbeauvais: This doesn't extend Vue, so you can just write `export interface AlertMessage {` (and stop importing Vue). I know you just copied what was there before but it's wrong too ... feel free to update the other interfaces.

This file contains a single interface, so I'd be tempted to name it singular instead of plural.

Why not name this file alertMessage-interface, according to its content?

(Sorry, I'm picky today).

coops-ui/src/mixins/resource-lookup-mixin.ts
```diff
@@ -40,4 +40,40 @@ export default class ResourceLookupMixin extends Vue {
...
       return ''
     }
+
+    /**
+     * Method to validate directors on edit/cease and return any warning messages.
+     *
+     * @return the compliance message or null (if the configuration has been loaded).
+     */
+    directorWarning (directors: Array<any>): AlertMessage {
+      const configSection = this.configObject.flows.find(x => x.feeCode === 'OTCDR').warnings
```
severinbeauvais: Please use filing code enum instead of OTCDR.

coops-ui/src/resources/business-config.ts
```diff
@@ -33,7 +33,16 @@ export const configJson = [{
...
            'respect of a material fact in a record submitted to the Corporate Registry for filing. ' +
-           'See section 127 of the Business Corporations Act.'
+           'See section 127 of the Business Corporations Act.',
+      warnings: {
+        minDirectors: {
+          count: 1,
+          title: 'One Director Required',
+          message: 'A minimum of one director is required, to be in compliance with the ' +
+            'Business Corporations Act (Section 120). You can continue your filing, but you must ' +
+            'become compliant with the  Business Corporations Act. as soon as possible.'
```
severinbeauvais: Extra space and period in this line.

coops-ui/src/resources/business-config.ts
```diff
@@ -60,7 +69,32 @@ export const configJson = [{
       displayName: 'Change Of Directors',
       certifyText: 'Note: It is an offence to make a false or misleading statement in ' +
            'respect of a material fact in a record submitted to the Corporate Registry for filing. ' +
-           'See section 78 of the Cooperative Association Act.'
+           'See section 78 of the Cooperative Association Act.',
```
severinbeauvais: Looks like "section" is capitalized almost everywhere else. Please fix globally.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -844,4 +853,9 @@ h2 {
     margin-left: 0.5rem;
   }
 }
+
+.complianceDialogMsg {
+  font-size:16px;
```
severinbeauvais: Please change size to "1rem" (which in our case is 16px).

For fonts especially, use rem units to allow the browser to resize elements consistently.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -844,4 +853,9 @@ h2 {
...
+  color:#495057
```
severinbeauvais: Please use CSS variable for this colour instead.

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -1589,4 +1644,9 @@ ul {
 .director-list-item {
   padding: 1.25rem;
 }
+
+.complianceCard {
+  font-size:14px;
```
severinbeauvais: Please specify font size in rems.

coops-ui/src/assets/styles/base.scss
```diff
@@ -266,3 +266,9 @@ ul.list {
 .vue-affix {
   width: 282px;
 }
+
+.icon-blue {
+  .v-icon {
+    color: #2196F3 !important;
```
severinbeauvais: See the comment I made about this colour in WarningPopover.vue.

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -155,18 +162,18 @@ describe('Directors as a COOP', () => {
   it('displays the list of directors', () => {
     const directorListUI = vm.$el.querySelectorAll('.director-list-item')
 
-    expect(directorListUI[1].textContent).toContain('Peter')
-    expect(directorListUI[1].textContent).toContain('Griffin')
+    expect(directorListUI[0].textContent).toContain('Peter')
```
severinbeauvais: Why did you change this unit test? Does it correctly verify that the director sorting is functioning?

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -191,39 +198,39 @@ describe('Directors as a COOP', () => {
...
 
     // Verify the updated text field value
     expect(vm.$el.querySelectorAll('.edit-director__first-name input')[0].value).toBe('Steve')
 
     // Verify that only Change Name component is shown
-    expect(vm.editFormShowHide.showAddress).toEqual(false)
+    expect(vm.editFormShowHide.showAddress).toEqual(true)
     expect(vm.editFormShowHide.showName).toEqual(true)
-    expect(vm.editFormShowHide.showDates).toEqual(false)
+    expect(vm.editFormShowHide.showDates).toEqual(true)
```
severinbeauvais: So now all 3 components are showing? That does not seem correct. At a minimum, the comment on line 212 should be updated, but I don't understand why this test would be so fundamentally different from before?

(If these fixes are not due to your code then I recommend you don't make these changes and instead create a ticket to fix these tests.)

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -191,39 +198,39 @@ describe('Directors as a COOP', () => {
...
     setValue(vm, '.edit-director__first-name input', 'Steve')
 
     // Verify the name is updated
-    expect(vm.directors[1].officer.firstName).toBe('Steve')
+    expect(vm.directors[0].officer.firstName).toBe('Steve')
 
     // Cancel Director edit and verify the name is back to its base name
     await vm.$el.querySelectorAll('.cancel-edit-btn')[0].click()
-    expect(vm.directors[1].officer.firstName).toBe('Joe')
+    expect(vm.directors[0].officer.firstName).toBe('Peter')
```
severinbeauvais: Same comment as above for lines 223, 229 and 233.

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -505,13 +536,13 @@ describe('Directors as a BCOMP', () => {
...
-    expect(directorListUI[0].textContent).toContain('Glasgow')
-    expect(directorListUI[0].textContent).toContain('1234')
-    expect(directorListUI[0].querySelector('.same-address').textContent).toContain('Same as Delivery')
+    expect(directorListUI[0].textContent).toContain('Griffin')
+    expect(directorListUI[0].textContent).toContain('mailing_address city')
+    expect(directorListUI[0].textContent).toContain('4321')
+    expect(directorListUI[1].textContent).toContain('Joe')
+    expect(directorListUI[1].textContent).toContain('Glasgow')
+    expect(directorListUI[1].textContent).toContain('1234')
+    expect(directorListUI[1].querySelector('.same-address').textContent).toContain('Same as Delivery')
```
severinbeauvais: Same comment as above re: why change these?

coops-ui/src/components/dialogs/WarningPopover.vue
peter-freshworks: sorry, vscode auto-inserted this

coops-ui/tests/unit/Directors.spec.ts
peter-freshworks: this was done before the new sorting ticket was merged, I will have to rebase and see if this is still valid

coops-ui/src/assets/styles/base.scss
```diff
@@ -266,3 +266,9 @@ ul.list {
 .vue-affix {
   width: 282px;
 }
+
+.icon-blue {
```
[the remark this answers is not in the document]
peter-freshworks: For whatever reason, no matter where I put this, it wasn't reflecting on the ui (unless I put it in base.scss)...I'll try some more!

coops-ui/src/interfaces/informationAlert-interfaces.ts
peter-freshworks: I've gone with plural because, conceivably, there could be more informational alert types

coops-ui/src/interfaces/informationAlert-interfaces.ts
```diff
@@ -0,0 +1,5 @@
+export interface AlertMessage {
+  title: String
+  msg: String
+  additionalItems: Array<String>
```
severinbeauvais: Use `string` primitive for properties/variables.

You should only use `String` objects to reference the object itself, for example to create a string ("new String()").

See this page https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String under heading "Distinction between string primitives and String objects".

coops-ui/src/main.ts
```diff
@@ -28,6 +28,24 @@ Vue.use(Affix)
 
 const vuetify = new Vuetify({ iconfont: 'mdi' })
 
+function resetAuth () {
```
severinbeauvais: What's this? It's not documented as a change in this PR's description or the ticket description.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -348,7 +356,8 @@ export default {
       totalFee: 0,
       // EntityTypes Enum
       EntityTypes,
-      FilingCodes
+      FilingCodes,
+      complianceDialogMsg: null
```
severinbeauvais: This looks like complianceDialogMsg should be an "EntityTypes Enum" (per comment on line 357).

Please move this up with the other properties or add a blank line above it to start a new properties section.

coops-ui/src/components/dialogs/WarningPopover.vue
```diff
@@ -0,0 +1,45 @@
...
+}
+</script>
+
+<style lang="scss">
+@import '@/assets/styles/theme.scss';
+.warning-text {
+  font-weight: bold;
+  font-size: 14px;
+  color: $BCgovIconBlue;
+    .v-icon {
```
severinbeauvais: This line should be at the same indent level as the line above it. Also, we try to add a blank line between style blocks, to make it easier to see that a new selector is declared.

coops-ui/src/components/dialogs/WarningPopover.vue
```diff
@@ -0,0 +1,45 @@
...
+  @Prop({ default: null })
+  private dialogObj: AlertMessage
+}
+</script>
+
+<style lang="scss">
+@import '@/assets/styles/theme.scss';
+.warning-text {
+  font-weight: bold;
+  font-size: 14px;
```
severinbeauvais: Please use rem units for font size.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -181,7 +182,14 @@
                 <header>
                   <h1 id="filing-header-review">Review: Director Change </h1>
                 </header>
-
+                <section v-if="complianceDialogMsg">
+                  <v-alert type="info" outlined
+                  icon="mdi-information"
```
severinbeauvais: Please indent lines 187-188.

coops-ui/src/views/StandaloneDirectorsFiling.vue
```diff
@@ -181,7 +182,14 @@
...
+                  class="white-background"
+                  >
+                    <p class="complianceDialogMsg">{{ complianceDialogMsg.msg }}</p>
+                </v-alert>
+                </section>
```
severinbeauvais: Something isn't right here: `</v-alert>` and `</section>` shouldn't be at the same indent level. Please fix.

coops-ui/tests/unit/Directors.spec.ts
```diff
@@ -468,6 +475,30 @@ describe('Directors as a BCOMP', () => {
...
+    Vue.nextTick(() => {
+    // click second director's cease button
+      click(vm, '#director-2-cease-btn')
+        Vue.nextTick(() => {
+          expect(vm.complianceMsg.msg).toContain('A minimum of one director is required')
+          expect(vm.$el.querySelector('.complianceCard')).not.toBeNull()
+          expect(vm.$el.querySelector('.warning-text')).not.toBeNull()
+          // un-cease director
+          click(vm, '#director-2-cease-btn')
+          Vue.nextTick(() => {
```
severinbeauvais: When you use "Vue.nextTick()" it's like using setTimeout with a callback function. This callback function (or in your case, arrow function) is called in a future slice of time... after your test has completed!

Your test has completed because it ran code to register a callback to run "in the next tick". Registering this callback succeeded.

The point is, the callback hasn't been run by the time your test completes.

To handle this, the second parameter of the "it()" should be "done" instead of "()". And then, when your test is totally complete (eg, after line 496) then you call `done()`.

There are other example of similar tests. Search for "done()".

In some cases, you can "await" for something to happen (like a click). For examples of that, search for tests with "async" in them.

coops-ui/src/main.ts
peter-freshworks: Yikes, I left some debug code in...good catch, Thanks!

coops-ui/src/assets/styles/base.scss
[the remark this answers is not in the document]
peter-freshworks: whoops, I've been using them as a check-list

coops-ui/src/mixins/resource-lookup-mixin.ts
```diff
@@ -1,6 +1,7 @@
 import { Component, Vue } from 'vue-property-decorator'
 import { mapState } from 'vuex'
-
+import { AlertMessage } from '@/interfaces'
```
cameron-freshworks: I believe we added an `IF` onto the end of our interfaces, ie `AlertMessageIF`. Not sure if you want to just to keep things consistent.

PR #638, opened 2020-01-07
Bugfix - Order of display of directors information

leksmall:
*Issue #:* /bcgov/entity#1734

*Description of changes:*
1. Legal API - Added sort order to the director query that returns the list of active directors
2. UI - Changes to retain the order of directors returned by the API and to add new directors at the 
    top of the list

### DIFF
--- a/coops-ui/src/components/AnnualReport/Directors.vue
+++ b/coops-ui/src/components/AnnualReport/Directors.vue
@@ -176,7 +176,7 @@
           <span>Appointed/Elected</span>
         </v-subheader>
         <li class="director-list-item"
-          v-for="(director, index) in orderBy(directors, 'id', -1)"
+          v-for="(director, index) in directors"
           :id="'director-' + director.id"
           :class="{ 'remove' : !isActive(director) || !isActionable(director)}"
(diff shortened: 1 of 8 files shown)


coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -1056,7 +1056,7 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
    */
   private saveEditDirector (index, id): void {
     // get current director
-    let director = this.directors[id - 1]
+    let director = this.directors.filter(el => el.id === id)[0]
```
severinbeauvais: A better choice for finding a single array element is https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find.

Also, you should check that you found something.

severinbeauvais: Actually, can't you just use 'index' to access the current director?

leksmall: I think Index can be used

PR #640, opened 2020-01-07
fixed director multiple forms issue

severinbeauvais:
*Issue #:* /bcgov/entity#2271

*Description of changes:*
- only render 1 set of edit forms when editing a director
- fixed syntax for interfaces properties
- fixed unit tests

### DIFF
--- a/coops-ui/src/interfaces/address-interfaces.ts
+++ b/coops-ui/src/interfaces/address-interfaces.ts
@@ -1,29 +1,29 @@
 import Vue from 'vue'
 
 export interface BaseAddressType extends Vue {
-  $refs: any
+  $refs: any;
 }
 
 // Base address Interface to match the address.json schema and it's optional fields.
 export interface AddressIF extends Vue {
-  actions?: string[],
-  addressCity: string
-  addressCountry: string
-  addressRegion: string
-  deliveryInstructions?: string
-  postalCode: string
-  streetAddress: string
-  streetAddressAdditional?: string
+  actions?: string[];
+  addressCity: string;
+  addressCountry: string;
+  addressRegion: string;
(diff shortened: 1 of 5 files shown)


coops-ui/src/interfaces/address-interfaces.ts
```diff
@@ -1,29 +1,29 @@
 import Vue from 'vue'
 
 export interface BaseAddressType extends Vue {
-  $refs: any
+  $refs: any;
```
cameron-freshworks: This is interesting! Are you able to elaborate on why the semi-colons in this file?

severinbeauvais: This is the normal way that properties (indeed, any JS statement) is terminated. However, because this is actually TS, there is some flexibility in the syntax. But I prefer terminating my statements explicitly to avoid any surprises!

cameron-freshworks: Fair enough and i couldn't agree more. Personally i would love to see the semi colons everywhere :)  I guess the question should have been more along the lines " Why here in the interfaces but not elsewhere? "

severinbeauvais: Well, for one, one of the interfaces used semi-colons and the others didn't, so I made it consistent. But you're right, a lot (most?) of our other code doesn't use semi-colons for properties. So this could go either way. This particular change seemed like a good idea at the time, hah.

cameron-freshworks: Works for me, i was just curious!

severinbeauvais:
I have 2 approvals so I'm merging this. OKed by Katie.

PR #641, opened 2020-01-07
Bugfix - IE11 - cannot delete draft

leksmall:
*Issue #:* /bcgov/entity#1853, /bcgov/entity#2037, /bcgov/entity#1805

*Description of changes:*
Added min-width to the list item so that the delete draft button is visible

### DIFF
--- a/coops-ui/src/components/AnnualReport/Directors.vue
+++ b/coops-ui/src/components/AnnualReport/Directors.vue
@@ -1548,6 +1548,10 @@ ul {
 .actions__more-actions {
   padding: 0;
 
+  .v-list-item {
+    min-width: 160px;
+  }
+
   .v-list-item__title {
     font-size: 0.875rem;
   }
--- a/coops-ui/src/components/Dashboard/TodoList.vue
+++ b/coops-ui/src/components/Dashboard/TodoList.vue
@@ -708,7 +708,7 @@ export default {
   padding: 0;
 
   .v-list-item {
-    min-width: 120px;
+    min-width: 140px;
   }
 
   .v-list-item__title {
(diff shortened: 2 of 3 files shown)


severinbeauvais:
I noticed this bug occurs in other places. Can you fix those too please (and update the ticket accordingly)?
1. Update business profile drop-down
2. Directors change drop-down
3. Cancel payment (I cannot see this yet so I am assuming)

leksmall:
> I noticed this bug occurs in other places. Can you fix those too please (and update the ticket accordingly)?
> 
> 1. Update business profile drop-down
> 2. Directors change drop-down
> 3. Cancel payment (I cannot see this yet so I am assuming)

Cancel Payment and Delete Draft uses the same style. So it works. The other component fixes are added to this PR now. Will link the tickets for them in this PR

coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -1548,6 +1548,10 @@ ul {
 .actions__more-actions {
   padding: 0;
 
+  .v-list-item {
+    min-width: 160px;
```
severinbeauvais: Looks like the menuitem could use some vertical centering also.
drop down

coops-ui/src/components/Dashboard/TodoList.vue
```diff
@@ -707,6 +707,10 @@ export default {
 .actions__more-actions {
   padding: 0;
 
+  .v-list-item {
+    min-width: 140px;
```
severinbeauvais: Too wide. Try 112px.

coops-ui/src/components/AnnualReport/Directors.vue
severinbeauvais: Please also check v centering on others.

coops-ui/src/components/Dashboard/TodoList.vue
leksmall: This is to take care of Cancel Payment

PR #661, opened 2020-01-13
Director sort order - Moving sort logic to UI

leksmall:
*Issue #:* /bcgov/entity#1734

*Description of changes:*
Removed the sort logic from legal API
Added the sorting logic in the UI
Updated tests


coops-ui/src/components/AnnualReport/Directors.vue
```diff
@@ -820,6 +820,12 @@ export default class Directors extends Mixins(DateMixin, ExternalMixin, EntityFi
 
             var directors = response.data.directors
 
+            const fieldSorter = (fields) => (a, b) => fields.map(o => {
+              return a['officer'][o] > b['officer'][o] ? 1 : a['officer'][o] < b['officer'][o] ? -1 : 0
+            }).reduce((p, n) => p || n, 0)
+
+            directors = directors.sort(fieldSorter(['lastName', 'firstName', 'middleName']))
+
```
cameron-freshworks: We may only use the function twice so far, but would it make sense to move it into a mixin of sorts to prevent duplicating it?

coops-ui/src/App.vue
```diff
@@ -313,7 +313,14 @@ export default {
 
     storeDirectors (response) {
       if (response && response.data && response.data.directors) {
-        const directors = response.data.directors
+        const directorsList = response.data.directors
+
+        const fieldSorter = (fields) => (a, b) => fields.map(o => {
+          return a['officer'][o] > b['officer'][o] ? 1 : a['officer'][o] < b['officer'][o] ? -1 : 0
+        }).reduce((p, n) => p || n, 0)
```
cameron-freshworks: Complex :) Maybe just a short comment describing the jist of the function!

coops-ui/src/components/AnnualReport/Directors.vue
leksmall: Sure. Will do that

leksmall: Moved the method to a mixin and added some documentation

coops-ui/src/mixins/director-mixin.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Mixin that provides some utilities for director data.
+ */
+@Component({})
+export default class DirectorMixin extends Vue {
+  /**
+   * Sorter function based on ES6 map reduce to sort the list of directors in the json.
+   * @param fields The list of fields based on which director list needs to be sorted.
+   * @return The sort function
+   */
+  fieldSorter (fields: Array<string>): any {
```
severinbeauvais: See if you can return a type different than "any" (for future strict TS).

Also, my linter is warning me that 'a' and 'b' implicitly have "any" type (which may also cause issues with future strict TS).

leksmall:
> The functionality looks fine so I approve this, but see if you can improve typings to prevent future work when we enable strict TS.

Sure. Will do that in a separate PR

PR #662, opened 2020-01-13
#2331 Updated Design of UI for FilingHistory List & Address Component

cameron-freshworks:
*Issue #:* /bcgov/entity#2331

*Description of changes:*
* Updated Filing history UI for future effective dates. 
https://user-images.githubusercontent.com/26335434/71854968-93bab000-3094-11ea-8b42-126229b19d82.PNG

* Fixed UI for Address component.
https://images.zenhubusercontent.com/5d49ce359547713e57d6cd4b/f8557ee5-b404-4ad1-a553-503eff0a94eb

* Fixed Some functionality for Records address when it is the same as Registered Address. 

* Updated Unit testing.

### DIFF
--- a/coops-ui/src/components/Dashboard/FilingHistoryList.vue
+++ b/coops-ui/src/components/Dashboard/FilingHistoryList.vue
@@ -19,28 +19,28 @@
                 <v-col cols="6" class="v-col-padding">
                   <div class="list-item__title mb-1">{{item.name}}</div>
                 </v-col>
-                <v-col cols="6" class="v-col-padding" v-if="isCoaFutureEffective(item.name, item.status)">
-                  <v-scale-transition>
-                    <v-tooltip
-                      top
-                      content-class="pending-tooltip"
-                    >
-                      <template v-slot:activator="{ on }">
-                        <div id="pending-alert" class="list-item__subtitle" v-on="on">
-                          <v-icon color="yellow" small>mdi-alert</v-icon>
-                          PAID AND PENDING
-                        </div>
-                      </template>
-                      <span>
+              </v-row>
+              <div class="list-item__subtitle">
+                <v-scale-transition v-if="isCoaFutureEffective(item.name, item.status)">
+                  <v-tooltip
+                    top
+                    content-class="pending-tooltip"
+                  >
+                    <template v-slot:activator="{ on }">
+                      <div id="pending-alert" class="list-item__subtitle" v-on="on">
+                        <span>
+                          FILED AND PENDING (filed by {{item.filingAuthor}} on {{item.filingDate}})
+                        </span>
+                        <v-icon color="yellow" small>mdi-alert</v-icon>
+                      </div>
+                    </template>
+                    <span>
                       The updated office addresses will be legally effective on {{ item.filingEffectiveDate }},
                       12:01 AM (Pacific Time). No other filings are allowed until then.
                     </span>
-                    </v-tooltip>
-                  </v-scale-transition>
-                </v-col>
-              </v-row>
-              <div class="list-item__subtitle">
-                <span>FILED AND PAID (filed by {{item.filingAuthor}} on {{item.filingDate}})</span>
+                  </v-tooltip>
+                </v-scale-transition>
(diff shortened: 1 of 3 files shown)


coops-ui/src/components/common/OfficeAddresses.vue
```diff
@@ -286,7 +286,7 @@ export default class OfficeAddresses extends Mixins(CommonMixin, EntityFilterMix
   private inheritRecDeliveryAddress: boolean = true
 
   // State of the checkbox for determining whether the Record address is the same as the Registered address
-  private inheritRegisteredAddress: boolean = true
+  private inheritRegisteredAddress: boolean = this.isSame(this.registeredAddress, this.recordsAddress)
```
severinbeauvais: This probably doesn't work as you expect because this assignment is done when the component is created (I *think*) and therefore this.registeredAddress and this.recordsAddress aren't set yet.

I might be wrong but I believe I saw this situation in the BaseAddress component. If I'm right then this code looks like it does something but actually it doesn't, so it should be changed.

If you really care about the value of inheritRegisteredAddress then you may need other code (eg, watcher) to set it (unless that already exists -- I didn't check).

Is there a unit test that checks for the situation where the address are the same, and different? If not, please add.

coops-ui/src/components/common/OfficeAddresses.vue
```diff
@@ -103,7 +103,7 @@
             />
           </div>
 
-          <div v-if="!isSame(registeredAddress, recordsAddress) || !inheritRegisteredAddress">
```
severinbeauvais: Now this code is reactive, whereas the new code on line 289 isn't. (See my other comment.)

coops-ui/src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -19,28 +19,28 @@
...
-                      <template v-slot:activator="{ on }">
-                        <div id="pending-alert" class="list-item__subtitle" v-on="on">
-                          <v-icon color="yellow" small>mdi-alert</v-icon>
-                          PAID AND PENDING
-                        </div>
-                      </template>
-                      <span>
+              </v-row>
+              <div class="list-item__subtitle">
+                <v-scale-transition v-if="isCoaFutureEffective(item.name, item.status)">
```
severinbeauvais: Can you really put a v-if in a transition? I thought transitions were only wrappers for elements that changed visibility or existence.

(Sorry, can't check online docs as they're currently broken.)

coops-ui/src/components/common/OfficeAddresses.vue
severinbeauvais: Also, in those unit tests, set the prop after you create the component (don't pass the props in the creation statement) to simulate what happens irl.

cameron-freshworks: I'll have to dig into it a little further. After modifying this it was working as expected but i can certainly dig a little deeper into this!  Re Unit Tests: Absolutely i can definitely update!

coops-ui/src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: It works as expected but i will 100% dig into it further and look at the options!

coops-ui/src/components/common/OfficeAddresses.vue
cameron-freshworks: This was a good topic, thanks for bringing it up!    So after digging into it, it is ok to evaluate this property with the props passed into it ( it works as expected ) BUT it was lacking being reactive ( As you mentioned ). Initially i didn't think it needed to be reactive because the props values will always be present when the component mounts however i wasn't accounting for the draft filings ( which haven't modified the base state of these addresses yet ). So inside of the addresses watcher, we re-evaluate the state of the checkbox when we're passing draft data into the offices component.       RE: Unit tests - There is tests covering when they are the same and when they are different 👍

coops-ui/src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Thanks for looking into it and letting me know!

PR #663, opened 2020-01-13
2262 data reset tool in Openshift

katiemcgoff:
*Issue #:* /bcgov/entity#2262

*Description of changes:*
Data reset tool deployed in openshift - build and pipeline, plus dev deployment.
This is currently build from my branch in Openshift if you want to see it in action.
ALSO INCLUDES a bug fix for adding director addresses.

### DIFF
--- a/data-reset-tool/pre-hook-update-db.sh
+++ b/data-reset-tool/pre-hook-update-db.sh
@@ -1,0 +1,1 @@
+# THIS IS EMPTY - but present so that deployment config is consistent with other python applications.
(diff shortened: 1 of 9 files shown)


data-reset-tool/pre-hook-update-db.sh
```diff
@@ -0,0 +1 @@
+# THIS IS EMPTY - but present so that deployment config is consistent with other python applications.
```
kialj876: I think we should delete this file since it's not being used and is a bit confusing if you don't know that

PR #665, opened 2020-01-15
#1898 incorporation schema

cameron-freshworks:
*Issue #:* /bcgov/entity#1898

*Description of changes:*

* Implemented a schema for an incorporations filing, with regards to what we know at this point.

* Implemented a sub schema for potential name request data ( Optional for now, but included for a future rough guideline, as per designs ).

* Implemented example data and unit testing for incorporations schema.

* Removed duplicate import of  'Corp change of address'.

The tests run, the pylint checks out and the schema validates via some online tools, but my first crack at the back end schemas so be critical in review :) 👍 


schemas/src/registry_schemas/schemas/incorporation.json
```diff
@@ -0,0 +1,76 @@
...
+                "certifiedBy": {
+                    "type": "string"
+                }
+            }
+        },
+        "incorporation": {
+            "type": "object",
+            "required": [
+                "nameRequest",
+                "businessContact"
```
leksmall: Is offices an optional element?

cameron-freshworks: Good question, it won't be optional i don't believe, but i'm glad you asked because in our current sprint we won't have anything with regards to adding the office addresses component and data into the POST, so i left it as optional until we can implement that piece. Would it make sense to leave it out until we have it, or just enforce it now?

leksmall: I am ok with adding it later :).

schemas/tests/unit/test_incorporation.py
```diff
@@ -0,0 +1,136 @@
+# Copyright © 2019 Province of British Columbia
```
severinbeauvais: 2020 now 😁

cameron-freshworks: Time flies!

schemas/src/registry_schemas/schemas/incorporation.json
```diff
@@ -0,0 +1,50 @@
...
+    "required": [
+        "incorporation"
+    ],
+    "type": "object",
+    "title": "Incorporation Filing",
+    "properties": {
+        "incorporation": {
+            "type": "object",
+            "required": [
+                "nameRequest",
```
katiemcgoff: I'm late to the game on this, but can we require nameRequest in the case of numbered companies?

schemas/src/registry_schemas/schemas/name_request.json
```diff
@@ -0,0 +1,45 @@
...
+        "legalType": {
+            "type": "string",
+            "title": "The Legal Type of the business.",
+            "default": "BC",
+            "enum": [
+                "BC",
+                "CP"
+            ]
+        },
+        "requestType": {
```
katiemcgoff: Can you add some description for some of these fields? I'm unclear how this differs from legalType.

schemas/src/registry_schemas/schemas/incorporation.json
cameron-freshworks: never late! Absolutely, thats a great point. As far as the flow is concerned, the only way we will know the incorporating businesses legaltype is via the NameRequest portion of the application. So inside the name request schema ( currently ) legaltype is the only thing required at this point. As far we as know ( aka - a chat with Linda about it ) even numbered companies will need to still indicate a type through this NR flow.  Does it make sense to leave it as required, or possibly have legaltype of requirement of another section perhaps?

schemas/src/registry_schemas/schemas/name_request.json
cameron-freshworks: Absolutely, but to be honest i don't really know what it means :laugh. I can certainly take a crack at giving them a better definition though :).         
                                                               
I just pulled these properties off of the design. For this specific property, the value given ( as per design mockup ) was `New Business`.  Attached is a link to the mock up.
 
https://preview.uxpin.com/523b0c07ed451c8c3a417a9b758e194424fc1ead#/pages/121388482/simulate/sitemap

schemas/src/registry_schemas/schemas/incorporation.json
katiemcgoff: I suppose it depends whether the numbered co's come in with an NR (you can do an NR for a numbered company in the current systems), or not. If the assumption in our current functionality is that they have an NR, ie: we haven't moved on to numbered co's, then we can leave it as is I suppose and change it if needed when we get to incorporating numbered co's. So I guess I'm fine with it for now. The schema is ever-evolving :)

schemas/src/registry_schemas/schemas/name_request.json
```diff
@@ -0,0 +1,45 @@
...
+            "type": "string"
+        },
+        "status": {
+            "type": "string"
+        },
+        "expDate": {
+            "type": "string",
+            "format": "date"
```
katiemcgoff: In the NR data, expiration date is datetime, although the time portion is always the same. I think though if we are getting NR data from an API call we will probably get it as datetime.

schemas/src/registry_schemas/schemas/name_request.json
```diff
@@ -0,0 +1,45 @@
...
+        },
+        "condition": {
```
katiemcgoff: This should be "consent" - the condition is wording in the NR decision text only (among other messaging in the text field) but this would be more around consent status. IF an entity requires consent from another organization to use a name, this is where it's tracked, and marked as received by staff when that happens. I'm not seeing where we're actually marking that in the new Names database (this team's last project) but it should come through as a single-character flag. You could leave it as unknown string for now but this might change.

schemas/src/registry_schemas/schemas/name_request.json
cameron-freshworks: sounds great, i will update!

schemas/src/registry_schemas/schemas/name_request.json
cameron-freshworks: Ok sounds good, thanks for the heads up :) Like you said, the Schema is ever evolving!

schemas/src/registry_schemas/schemas/name_request.json
```diff
@@ -0,0 +1,45 @@
...
+            "type": "string"
+        },
+        "submittedBy": {
```
katiemcgoff: Submitter info (aka applicant info) will need to be full contact-style details, ie: first name, last name, address fields, etc. I just spoke to Scott and he envisions being able to populate the Business Contact Info from these fields in the NR. So *if* we want to save this data and not have to retrieve the NR from API call to get this info, we'll want all those fields in this schema. 

Here's a link to the model in the namex project that this data will come from: https://github.com/bcgov/namex/blob/master/api/namex/models/applicant.py

schemas/src/registry_schemas/schemas/incorporation.json
```diff
@@ -0,0 +1,50 @@
...
+                        "$ref": "https://bcrs.gov.bc.ca/.well_known/schemas/office"
+                    },
+                    "recordsOffice": {
+                        "$ref": "https://bcrs.gov.bc.ca/.well_known/schemas/office"
+                    },
+                    "required": [
+                        "registeredOffice"
+                    ]
+                },
+                "businessContact" : {
```
katiemcgoff: might want to make this a sub-schema for use in businessContact and NR submitter/applicant. See notes below.

schemas/src/registry_schemas/schemas/name_request.json
cameron-freshworks: Awesome! This is a great thought, there is certainly some questions still around the process a little. Will discuss in person :)

schemas/src/registry_schemas/schemas/incorporation.json
cameron-freshworks: Good idea, we can do that for sure, thanks!

schemas/src/registry_schemas/schemas/name_request.json
```diff
@@ -0,0 +1,52 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://bcrs.gov.bc.ca/.well_known/schemas/name_request",
+    "type": "object",
+    "title": "Name Request Schema",
+    "properties": {
+        "nrNumber": {
+            "type": "string"
```
katiemcgoff: We can lock this down to a 10-character string, since that's what it is. Also starts with "NR", ie: "NR 1234567".

cameron-freshworks: ok great, will do!

schemas/src/registry_schemas/schemas/filing.json
```diff
@@ -143,12 +143,28 @@
...
                 },
                 {
-                    "$ref": "https://bcrs.gov.bc.ca/.well_known/schemas/business"
+                    "if": {
+                        "properties": {
+                            "filingType": {
+                                "not": {
+                                  "const": "incorporation"
+                                }
+                            }
```
katiemcgoff: Are we not able to nest this condition a bit more to get to filing.header.name, rather than adding the new property filing.filingType? Adding another field here, which basically duplicates the same thing as in header, feels like too much.

Also (if you keep this new property) we should call it the same as header, so "incorporationApplication" rather than "incorporation".

schemas/src/registry_schemas/schemas/incorporation.json
```diff
@@ -0,0 +1,37 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://bcrs.gov.bc.ca/.well_known/schemas/incorporation",
+    "required": [
+        "incorporation"
+    ],
+    "type": "object",
+    "title": "Incorporation Filing",
+    "properties": {
+        "incorporation": {
```
katiemcgoff: This should be incorporationApplication, to match header.name options. That has been the approach so far and enables searching for json properties based on filing.name in our code.

cameron-freshworks: ok great, will update!

schemas/src/registry_schemas/schemas/filing.json
cameron-freshworks: I looked into the options here, as i would also prefer not having an additional property. I will keep looking into it and find a way! 🤞

schemas/src/registry_schemas/schemas/contact_point.json
```diff
@@ -0,0 +1,23 @@
...
+          },
+          "phone": {
+              "type": "string",
+              "format": "phone"
+          },
+          "extension": {
+              "type": "number"
+          }
+      }
+}
```
katiemcgoff: Nice use of schema.org (JSON LD). Although technically it's "telephone" and not "phone" according to that schema :)

schemas/src/registry_schemas/schemas/filing.json
cameron-freshworks: Do you have any suggestions of how to approach this without using an additional property, i'm struggling finding a way to reference a specific property from another local reference?

katiemcgoff: I had thought it was along the lines of:
```
"if": {
   "properties": {
      "header": {
          "properties": {
               "name": {
                   "not": {
                      "const": "incorporation"                             
```

However I can see no actual evidence of this. Let me just use a schema validator to check.

cameron-freshworks: I've been trying variations of this, i'll keep at it 👍

schemas/src/registry_schemas/schemas/incorporation.json
katiemcgoff: Yep just did a simple check of the syntax and it works! FYI you can use https://www.jsonschemavalidator.net to check syntax with simplified examples.

schemas/src/registry_schemas/schemas/filing.json
katiemcgoff: Yep just did a simple check of the syntax and it works! FYI you can use https://www.jsonschemavalidator.net to check syntax with simplified examples, that's what I use.

katiemcgoff: > Yep just did a simple check of the syntax and it works! FYI you can use https://www.jsonschemavalidator.net to check syntax with simplified examples, that's what I use.

(using "incorporationApplication", not "incorporation" as the constant).

cameron-freshworks: You're awesome, i was overthinking this. Has been updated!

PR #668, opened 2020-01-17
Fix for CRON job hang

peter-freshworks:
*Issue #:* /bcgov/entity#2161

*Description of changes:*
After some troubleshooting and debugging, this seems to be fixed. Added libraries to requirements file

### DIFF
--- a/jobs/future-effective-filings/requirements/prod.txt
+++ b/jobs/future-effective-filings/requirements/prod.txt
@@ -7,5 +7,6 @@ flask-jwt-oidc>=0.1.5
 python-dotenv
 requests
 sentry-sdk[flask]
+python-dateutil
 
 git+https://github.com/bcgov/lear.git#egg=registry_schemas&subdirectory=schemas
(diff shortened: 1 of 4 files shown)


jobs/future-effective-filings/requirements/prod.txt
```diff
@@ -7,5 +7,9 @@ flask-jwt-oidc>=0.1.5
 python-dotenv
 requests
 sentry-sdk[flask]
-
+python-dateutil
+asyncio-nats-client
+asyncio-nats-streaming
 git+https://github.com/bcgov/lear.git#egg=registry_schemas&subdirectory=schemas
+git+https://github.com/bcgov/lear.git#egg=legal_api&subdirectory=legal-api
```
katiemcgoff: I'd remove legal_api from requirements since it's not actually used; otherwise confusing. Also I don't think registry_schema is used? Definitely entity_queue_common is.

PR #670, opened 2020-01-20
1894 save incorporation filing

peter-freshworks:
*Issue #:* /bcgov/entity#194

*Description of changes:*
Looking for input on this method of creating/saving an incorporation filing. POST/PUT via businesses, GET via Filings (as normal) 


legal-api/src/legal_api/models/business.py
```diff
@@ -222,7 +222,7 @@ def validate_identifier(identifier: str) -> bool:
         # TODO This is not correct for entity types that are not Coops
         if identifier[:-7] not in ('CP', 'XCP'):
             return False
-
+        """
```
katiemcgoff: I think instead of removing the functionality of validate_identifier, we could just not call it if the filing type is incorporation. That way we still get the validation when we need it.

legal-api/src/legal_api/resources/business/business.py
```diff
@@ -41,3 +45,65 @@ def get(identifier):
...
+            
+            if business:
+                return  {'message': f'Incorporation filing for {business_id} already exists'}, \
+                HTTPStatus.BAD_REQUEST
+            # Create an empty business record, to be updated by the filer
+            business = Business()
+            business.identifier = temp_corp_num
+            business.save()
+
+        # Ensure the business identifier matches the NR in the filing
```
katiemcgoff: I think this should go into the validator.

legal-api/src/legal_api/models/business.py
peter-freshworks: Added NR for now

legal-api/src/legal_api/resources/business/business.py
peter-freshworks: I added the soft validator to the Validator, and kept anything hitting the DB in this method. The validator will surely get more complex as more rules are discovered.

legal-api/src/legal_api/resources/business/business.py
```diff
@@ -41,3 +45,66 @@ def get(identifier):
...
+        if not filing:
+            filing = Filing()
+            # filing.filing_type = 'incorporationApplication'
+            filing.business_id = business.id
+        elif len(filing) > 1:
+            return {'message': 'more than one incorporation filing found for corp'}, HTTPStatus.BAD_REQUEST
+        else:
+            filing = filing[0]
+        filing.filing_json = incorporation_body
+        filing.save()
```
leksmall: As data is getting  persisted to multiple tables shouldnt this happen in a transactional way? ie commit() should be called only after both are added to the session

legal-api/src/legal_api/resources/business/business.py
```diff
@@ -41,3 +45,66 @@ def get(identifier):
...
+    @staticmethod
+    def _save_incorporation_filing(incorporation_body, client_request, business_id=None):
+        """Create or update an incorporation filing. """
+        # Check that there is a JSON filing
+        if not incorporation_body:
+            return  {'message': f'No filing json data in body of post for incorporation'}, \
+                HTTPStatus.BAD_REQUEST
+        # Validate Json
+
+        temp_corp_num = incorporation_body['filing']['incorporationApplication']['nameRequest']['nrNumber']
```
leksmall: Is this nrNumber applicable for numbered company? Or this will change when we support numbered company?

peter-freshworks: it is *likely* to change...it may be more a matter of semantics (which is why I named the variable temp corp num)

legal-api/src/legal_api/resources/business/business.py
peter-freshworks: Good point, I will make the change

PR #674, opened 2020-01-21
fixed logic that caused looping

severinbeauvais:
*Issue #:* /bcgov/entity#2377

*Description of changes:*
- set missing id so logic drops through and doesn't cause infinite looping
- minor cleanup + added helpful comments for future development/debugging

### DIFF
--- a/coops-ui/src/views/Dashboard.vue
+++ b/coops-ui/src/views/Dashboard.vue
@@ -180,13 +180,14 @@ export default {
       clearTimeout(this.refreshTimer)
 
       let filingId = null
+      // NB: use unary plus operator to cast string to number
       if (this.$route !== undefined) filingId = +this.$route.query.filing_id // if missing, this is NaN
 
       // only consider refreshing the dashboard if we came from a filing
       if (!filingId) return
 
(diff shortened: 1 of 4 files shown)


coops-ui/src/views/Dashboard.vue
```diff
@@ -196,18 +197,18 @@ export default {
...
+    // checks whether this filing's status has changed
+    // respawns itself approx. every 1 second for up to 10 iterations
     checkFilingStatus (filingId) {
-      // check whether this filing's status has changed - recursive, runs approx. every 1 second for up to 10 seconds
-
-      this.checkFilingStatusCount++
-
       // stop this cycle after 10 iterations
-      if (this.checkFilingStatusCount > 10) {
+      if (++this.checkFilingStatusCount >= 10) {
```
cameron-freshworks: This is interesting. How does this work?

severinbeauvais: In a nutshell, the property is incremented BEFORE it's used in the expression. Compare with post-increment where the property is incremented AFTER it's used in the expression. In this case I felt that it cleaned up the code nicely and made sense.

https://www.google.com/search?q=javascript+pre+increment

PR #680, opened 2020-01-23
enable majority directors rule for coops, fix majority calculation

peter-freshworks:
*Issue #:* /bcgov/entity#1448

*Description of changes:*
Enable Canadian resident and BC resident rules for Coop (originally thought to be out-of-scope for 1448)

### DIFF
--- a/coops-ui/src/mixins/resource-lookup-mixin.ts
+++ b/coops-ui/src/mixins/resource-lookup-mixin.ts
@@ -48,35 +48,44 @@ export default class ResourceLookupMixin extends Vue {
      * @return the compliance message or null (if the configuration has been loaded).
      */
     directorWarning (directors: Array<any>): AlertMessageIF {
-      const configSection = this.configObject.flows.find(x => x.feeCode === FilingCodes.DIRECTOR_CHANGE_OT).warnings
-      let errors = []
       // FUTURE: Too much code for this. Can be condensed and made more reusable.
-      if (configSection.bcResident) {
-        if (directors.filter(x => (x.deliveryAddress.addressRegion !== 'BC' ||
-        (x.mailingAddress && x.mailingAddress.addressRegion !== 'BC'))).length === directors.length) {
-          errors.push({ 'title': configSection.bcResident.title, 'msg': configSection.bcResident.message })
+      if (directors && directors.length) {
+        const configSection = this.configObject.flows.find(x => x.feeCode === FilingCodes.DIRECTOR_CHANGE_OT).warnings
+        let errors = []
+        // If this entity has a BC Residency requirement for directors, one of the
+        // directors specified needs to have both their mailing and delivery address within British Columbia
+        if (configSection.bcResident) {
+          if (directors.filter(x => (x.deliveryAddress.addressRegion !== 'BC' ||
+                                    (x.mailingAddress &&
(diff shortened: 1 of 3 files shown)


coops-ui/src/mixins/resource-lookup-mixin.ts
```diff
@@ -59,8 +59,11 @@ export default class ResourceLookupMixin extends Vue {
       }
 
       if (configSection.canadianResident) {
-        if (directors.filter(x => (x.deliveryAddress.addressCountry !== 'CA' ||
-        (x.mailingAddress && x.mailingAddress.addressRegion !== 'BC'))).length === directors.length) {
+        const count = directors.length
+        const notCanadian = directors.filter(x => (x.deliveryAddress.addressCountry !== 'CA' ||
+        (x.mailingAddress && x.mailingAddress.addressCountry !== 'CA'))).length
+
+        if (notCanadian / count > 0.5) {
```
cameron-freshworks: Interesting!  Business rules on full display here :)

severinbeauvais: Is there any chance we'll get here with 0 directors?

coops-ui/src/mixins/resource-lookup-mixin.ts
```diff
@@ -59,8 +59,11 @@ export default class ResourceLookupMixin extends Vue {
       }
 
       if (configSection.canadianResident) {
-        if (directors.filter(x => (x.deliveryAddress.addressCountry !== 'CA' ||
-        (x.mailingAddress && x.mailingAddress.addressRegion !== 'BC'))).length === directors.length) {
+        const count = directors.length
+        const notCanadian = directors.filter(x => (x.deliveryAddress.addressCountry !== 'CA' ||
+        (x.mailingAddress && x.mailingAddress.addressCountry !== 'CA'))).length
```
severinbeauvais: The requirements doc doesn't specify WHICH address(es) to use to determine whether a director had an address outside Canada. Is it where they live? Is it where they do business?

Please confirm this with a BA and then add a code comment indicating the decision so that reviewers/future devs know what the exact business rule is implemented here.

peter-freshworks: It should be both mailing and delivery (I've been told). I will add comments explaining that.

severinbeauvais: Both or either?

severinbeauvais:
Both or either?
________________________________

coops-ui/src/mixins/resource-lookup-mixin.ts
peter-freshworks: None of their addresses should be outside of Canada

coops-ui/src/mixins/resource-lookup-mixin.ts
severinbeauvais: I see that a check was added for the directors object to exist and for its length to not be null/zero. Therefore, count should never be zero and this code is now safe.

coops-ui/src/mixins/resource-lookup-mixin.ts
severinbeauvais: Thanks for the comments in the code indicating the business rule that is implemented.

PR #682, opened 2020-01-24
2410 cross reference table colin event

kialj876:
*Issue #:* /bcgov/entity#2410

*Description of changes:*
- added table for colin event ids sothat we can map multiple filings from colin to single filings in lear (i.e. an AR with a COA and COD within it)


legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -484,25 +485,32 @@ def get(status=None):
...
+            filing = Filing.find_by_id(filing_id)
+            if not filing:
+                return {'message': f'{filing_id} no filings found'}, HTTPStatus.NOT_FOUND
+            try:
+                colin_event_id_obj = ColinEventId()
+                colin_event_id_obj.colin_event_id = colin_id
+                filing.colin_event_ids.append(colin_event_id_obj)
+                filing.save()
+            except BusinessException as err:
+                print(err)
```
katiemcgoff: no print :)

PR #683, opened 2020-01-25
added return URL to login redirect

severinbeauvais:
*Issue #:* /bcgov/entity#2433

*Description of changes:*
- deleted obsolete SBC Header props + unneeded code
- added env variable for Signin Path
- added Return URL to signin redirect
- updated startup error handling
- also fixed usage of Vue2Filters (no need for ExternalMixin)
- also fixed "no console" build warnings
- also added () to click handlers


coops-ui/src/views/AnnualReport.vue
```diff
@@ -206,15 +206,15 @@
           v-if="isAnnualReportEditable"
           :disabled="!isSaveButtonEnabled || busySaving"
           :loading="saving"
-          @click="onClickSave"
+          @click="onClickSave()"
```
cameron-freshworks: Ive noticed the added parenthesis throughout though i don't fully understand the difference in this context. Essentially if feels like the same thing with or without them, so i'm just curious? Without parentheses we're referencing the method and when we have them we're invoking the method itself, thats about it for my understanding of it :)

coops-ui/src/App.vue
```diff
@@ -121,7 +115,8 @@ export default {
         businessId = this.getBusinessId()
         this.updateCurrentDate()
       } catch (error) {
-        console.error(error) // eslint-disable-line no-console
+        // eslint-disable-next-line no-console
+        console.error(error)
```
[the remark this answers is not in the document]
severinbeauvais: Either the error text from a 'throw new Error' statement or any other error the framework might generate during an exception. I think sometimes you get a stack track too.
I expect this kind of text -- even when our app generates it -- is too technical for a normal user so we log the error to the console and display a higher-level message to the user.

coops-ui/src/views/AnnualReport.vue
severinbeauvais: Both ways seem to work but I prefer to include the parentheses as it makes it more explicit that we're calling a method.

PR #719, opened 2020-02-05
#1725 incorporations api validators

cameron-freshworks:
*Issue #:* /bcgov/entity#1725

*Description of changes:*

* Implemented Validators on Legal API for Incorporation Office Addresses

* Implemented Unit Testing


cameron-freshworks:
I don't have much idea about the test that is failing except that it is/was also failing in master. I will look into it!

katiemcgoff:
> I don't have much idea about the test that is failing except that it is/was also failing in master. I will look into it!
> *Edit : has something to do with feature flags?

Not a problem, Thor is resolving in his latest work. It's something that fails locally for me but not Thor, for example.

schemas.deprecated/src/registry_schemas/example_data/schema_data.py
```diff
@@ -323,7 +323,8 @@
 
 INCORPORATION = {
         'nameRequest': {
-            'legalType': 'BC'
+            'legalType': 'BC',
+            'nrNumber': 'NR1234567'
         },
```
katiemcgoff: The schemas have recently been moved to their own repo business-schemas, so this change will need to move there. It can be its own PR (since different repo). 
https://github.com/bcgov/business-schemas

If you're working on this in your local branch again, you'll also want to change your requirements file to the new repo, but it's already changed in master so you can just rebase and re-install and you should be fine.

cameron-freshworks: ohh right, i will update this and submit a new PR to that repo. Totally forget this all moved recently. Thanks for the heads up!

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -30,6 +32,44 @@ def validate(business, incorporation_json):
...
+                    raise LookupError
+            except LookupError:
+                err_path = '/filing/incorporationApplication/offices/%s/%s/addressCountry' % (
+                    item, k
+                )
+                msg.append({'error': "Address Country must be 'CA'.",
+                            'path': err_path})
     if msg:
         return Error(HTTPStatus.BAD_REQUEST, msg)
+
```
katiemcgoff: I think it would be better to return just the list of messages. The actual message code from this response doesn't get used in the calling validate() function, only the messages, so might as well just return the list of messages ("msg") and adjust accordingly in lines 35-37. 

Your choice - it was jarring, though, to see the resturned Error object from this function though, so it might confuse others also.

cameron-freshworks: Sounds good, will update!

coops-ui/package-lock.json
```diff
@@ -0,0 +1,15337 @@
+{
```
katiemcgoff: This package-lock.json and the other one at lear root should not be part of this PR as far as I can tell (also coops-ui folder has moved since this was checked in).

PR #721, opened 2020-02-06
2140 entity filer

peter-freshworks:
*Issue #:* /bcgov/entity#2140

*Description of changes:*
Initial commit for incorporation application entity filer - should fetch the next available corp number, update the business (with the little information available currently in the filing). Dependent on COLIN API  

### DIFF
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -223,7 +223,7 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes
         except ValueError:
             return False
         # TODO This is not correct for entity types that are not Coops
-        if identifier[:-7] not in ('CP', 'XCP', 'NR', 'BC'):
+        if identifier[:-7] not in ('CP', 'XCP', 'NR'):
             return False
 
         return True
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -223,7 +223,7 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes
         except ValueError:
             return False
         # TODO This is not correct for entity types that are not Coops
-        if identifier[:-7] not in ('CP', 'XCP', 'NR'):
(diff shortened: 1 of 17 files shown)


legal-api/src/legal_api/models/business.py
```diff
@@ -223,7 +223,7 @@ def validate_identifier(identifier: str) -> bool:
         except ValueError:
             return False
         # TODO This is not correct for entity types that are not Coops
-        if identifier[:-7] not in ('CP', 'XCP'):
+        if identifier[:-7] not in ('CP', 'XCP', 'NR', 'BC'):
```
katiemcgoff: I moved the NR check up to line 213, since this breaks (doesn't allow valid NR value) when the NR is in the correct format "NR 1234567" with space. So you can remove "NR" from here and just leave "BC".

katiemcgoff:
There are flake8 and pylint errors :)

queue_services/entity-filer/src/entity_filer/filing_processors/incorporation_filing.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def insert_business_info(corp_num: str, business: Business, business_info: Dict):
+    """Format and update the business entity from incorporation filing."""
+    if corp_num and business and business_info:
+        business.identifier = corp_num
+        # TODO: Other properties contained in the NR
+    else:
+        return None
+    return business
```
katiemcgoff: I think this function would be better named "update_" rather than "insert_" business_info, since that's what it's doing - the business record already exists.

peter-freshworks: I was thinking "insert" since, up to this point, there is no actual info in the record, but I can change that to Update

peter-freshworks: should be resolved now

legal-api/src/legal_api/models/business.py
peter-freshworks: should be resolved now

PR #724, opened 2020-02-10
Added one-time script that was used for fixing older filings without offices.

katiemcgoff:
*Issue #:* /bcgov/entity#2537

*Description of changes:*
Added one-time script that was used for fixing older filings without offices.

### DIFF
--- a/lear-db/test_data/fix_missing_offices_json.py
+++ b/lear-db/test_data/fix_missing_offices_json.py
@@ -1,0 +1,68 @@
+# Copyright © 2019 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


lear-db/test_data/fix_missing_offices_json.py
```diff
@@ -0,0 +1,68 @@
...
+import copy
+
+load_dotenv(find_dotenv())
+
+FLASK_APP = Flask(__name__)
+FLASK_APP.config.from_object(get_named_config('production'))
+db.init_app(FLASK_APP)
+
+
+def do_stuff():
```
cameron-freshworks: I like the name!

lear-db/test_data/fix_missing_offices_json.py
```diff
@@ -0,0 +1,68 @@
...
+
+        deliv_addr = filing_json['filing']['changeOfAddress']['deliveryAddress']
+        mail_addr  = filing_json['filing']['changeOfAddress']['mailingAddress']
+
+        filing_json['filing']['changeOfAddress']['offices'] = {
+            'registeredOffice': {
+                'deliveryAddress': deliv_addr,
+                'mailingAddress': mail_addr,
+            }
+        }
```
cameron-freshworks: Just to clarify for myself, we don't need to do a check or wrap anything related to Records office because all of that data is POST Bcomps which have the wrapper already?

katiemcgoff: Correct - this is only for specific filings (found outside of this script, via sql) that are missing the office element in the json. So coops only, no bcomps. That's the commented-out line at the top with the list of filing ids. It's a "utility" (hack).

PR #730, opened 2020-02-11
fix for empty string in AGM Date

severinbeauvais:
*Issue #:* /bcgov/entity#2569

*Description of changes:*
- a "cleanup" commit was omitted from the initial PR and is now included here:
    - better initialization of filing codes
    - added unit test (+ cleanup)
- explicity set AGM Date to null because API doesn't handle empty string

### DIFF
--- a/coops-ui/src/views/AnnualReport.vue
+++ b/coops-ui/src/views/AnnualReport.vue
@@ -513,15 +513,11 @@ export default {
                 this.$refs.directorsList.setDraftDate(annualReport.annualGeneralMeetingDate)
               }
               if (this.entityFilter(EntityTypes.COOP)) {
-                // set the new AGM date in the AGM Date component
-                // (this value is empty if user did not enter an AGM date)
-                this.newAgmDate = annualReport.annualGeneralMeetingDate
-                // set the new No AGM flag in the AGM Date component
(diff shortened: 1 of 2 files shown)


coops-ui/src/views/AnnualReport.vue
```diff
@@ -513,15 +513,11 @@ export default {
                 this.$refs.directorsList.setDraftDate(annualReport.annualGeneralMeetingDate)
               }
               if (this.entityFilter(EntityTypes.COOP)) {
-                // set the new AGM date in the AGM Date component
-                // (this value is empty if user did not enter an AGM date)
-                this.newAgmDate = annualReport.annualGeneralMeetingDate
-                // set the new No AGM flag in the AGM Date component
-                this.newNoAgm = annualReport.didNotHoldAgm
+                // set the new AGM date in the AGM Date component (may be null or empty)
+                this.newAgmDate = annualReport.annualGeneralMeetingDate || ''
```
katiemcgoff: should this be empty string or null?

severinbeauvais: Either one is fine. The AGMDate component takes this value in as a prop as follows:
```
  @Prop({ default: null })
  private newAgmDate: string | null
```
After that it's treated as truthy or falsy, where "" is falsy.

PR #733, opened 2020-02-12
staff correction UI (Part 1) + dashboard plumbing (disabled)

severinbeauvais:
*Issue #:* /bcgov/entity#2607

*Description of changes:*
- updates to Dashboard and its components (but disabled)
- created Correction view (WIP)
- create DetailComment component
- cleaned up loading by filing type
- cleaned up filing names/types/titles
- added Correction type
- initial correction object for API
- added required rule to Certified By
- updated component names in HTML to kebab case
- changed currentDate prop from Certify and SummaryCertify to store variable
- added label to Certify date
- misc cleanup

### DIFF
--- a/coops-ui/src/components/AnnualReport/AGMDate.vue
+++ b/coops-ui/src/components/AnnualReport/AGMDate.vue
@@ -82,7 +82,7 @@ import { FormType } from '@/interfaces'
     ...mapGetters(['lastFilingDate'])
   }
 })
-export default class AGMDate extends Mixins(DateMixin) {
+export default class AgmDate extends Mixins(DateMixin) {
   // annotate form to fix "Property X does not exist on type Y" error
   $refs!: {
     form: FormType
@@ -164,7 +164,7 @@ export default class AGMDate extends Mixins(DateMixin) {
   }
 
   /**
-   * Lifecycle hook.
+   * Called when component is mounted.
    */
   private mounted (): void {
     // initialize date picker but not text field
--- a/coops-ui/src/views/Dashboard.vue
+++ b/coops-ui/src/views/Dashboard.vue
@@ -1,6 +1,6 @@
 <template>
   <div id="dashboard">
-    <CoaWarningDialog
+    <coa-warning-dialog
       :dialog="coaWarningDialog"
       @toggle="toggleCoaWarning"
       @proceed="goToStandaloneAddresses"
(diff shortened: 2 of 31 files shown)


coops-ui/src/components/AnnualReport/AGMDate.vue
```diff
@@ -82,7 +82,7 @@ import { FormType } from '@/interfaces'
     ...mapGetters(['lastFilingDate'])
   }
 })
-export default class AGMDate extends Mixins(DateMixin) {
+export default class AgmDate extends Mixins(DateMixin) {
```
severinbeauvais: FYI - I changed this so we could use the recommended markup for this component (`<agm-date ...>`). If I had left it as AGMDate then Vue would have expected "a-g-m-date" :/

severinbeauvais: PS I did not change the filename because git gets confused by that.

coops-ui/tests/unit/SummaryCertify.spec.ts
```diff
@@ -25,28 +23,34 @@ const defaultDate = '2019-01-01'
...
-      'currentDate': currentDate,
-      'isCertified': isCertified
+      certifiedBy,
+      isCertified
     } })
 }
 
 describe('SummaryCertified', () => {
-  let vm
+  // let vm
```
severinbeauvais: I'll delete this in my next PR.

coops-ui/src/views/Dashboard.vue
```diff
@@ -244,11 +245,9 @@ export default {
      *
      * @param filings The array of filings in history
      */
-    checkPendingFilings (filings) {
-      filings.forEach(filing => {
-        if (this.entityFilter(EntityTypes.BCOMP) &&
-        filing.name === FilingTypes.ADDRESS_CHANGE &&
-        filing.status === FilingStatus.PAID) {
+    checkPendingFilings (filings: Array<any>) {
+      filings.some(filing => {
```
severinbeauvais: Oops. This was an optimization to stop processing when we found an element, but the function needs to return true/false, which I forgot to add.

Because there is no return value, this will just process every filing (same as before). However I will add a return value in my next PR.

coops-ui/src/components/AnnualReport/AGMDate.vue
thorwolpert: got it, thanks and I like the consistent naming. nice to keep lifting the code forward.

PR #748, opened 2020-02-26
Added correction filing processing to entity filer.

katiemcgoff:
*Issue #:* /bcgov/entity#2637

*Description of changes:*
Added correction filing processing to entity filer.
Includes two changes to legal api to support corrections in Entity Filer.

### DIFF
--- a/legal-api/src/legal_api/models/filing.py
+++ b/legal-api/src/legal_api/models/filing.py
@@ -50,6 +50,7 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
         ERROR = 'ERROR'
         PAID = 'PAID'
         PENDING = 'PENDING'
+        PENDING_CORRECTION = 'PENDING_CORRECTION'
 
     class Source(Enum):
         """Render an Enum of the Filing Sources."""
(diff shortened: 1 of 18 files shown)


queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
```diff
@@ -0,0 +1,52 @@
...
+    correction_filing.comments.append(
+        Comment(
+            comment=filing['correction']['comment'],
+            staff_id=correction_filing.submitter_id
+        )
+    )
+
+    # set correction filing to PENDING, for manual intervention
+    # - include flag so that listener in Filing model does not change state automatically to COMPLETE
+    correction_filing._status = Filing.Status.PENDING.value  # pylint: disable=protected-access
```
thorwolpert: Consider a new state for corrections, as PENDING for other filings means awaiting payment.

katiemcgoff: Per discussion with Thor and then Cam/Sev re. UI, changing this status to a new status PENDING_CORRECTION and adjusting the rest of the UI to handle (to-do list, mainly).

legal-api/src/legal_api/models/filing.py
```diff
@@ -229,6 +229,13 @@ def _raise_default_lock_exception():
             status_code=HTTPStatus.FORBIDDEN
         )
 
+    @property
+    def is_corrected(self):
+        """Has this filing been corrected."""
+        if self.parent_filing:
```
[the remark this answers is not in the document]
katiemcgoff: It's always the correction filing as far as we know right now. But you are right we should check type. I will adjust.

PR #766, opened 2020-03-19
code changes to update/get for new db director tables

kialj876:

*Issue #:* /bcgov/entity#2820

*Description of changes:*
- legal changes for getting director info from party_role/party table and tests
- filer changes for updating party+role/party tables, also req update
- notebook changes to include commands for getting/updating the party_role/party tables in director section (check my import of party_role/party)


kialj876:
> At some point we need to add in datetime into the GET as an optional param.
> We would either write the SQL or use the continuum plugin to get the version that matches the dates.
> Do you want to consider this now, or as a new issue once this is done?

I think we should do that as a separate issue and include changing the schema with that etc. We should really be sending back a datetime from the frontend and just keeping it as a datetime at all times I think (right now the appointment/cessation dates are set with dates not datetimes so searching via a date makes sense and if we change that I think we should change it all the way through)

legal-api/src/legal_api/models/party.py
```diff
@@ -107,6 +107,16 @@ def valid_party_type_data(self):
                 return False
         return True
 
+    @classmethod
+    def find_by_name(cls, first_name: str, last_name: str, organization_name: str):
```
thorwolpert: If you add
from __future__ import annotations
then you should be able to type the return:
def find_by_name(cls, first_name: str, last_name: str, organization_name: str) -> Party:
Since we're targeting Py3.7+ we should start adding this.

legal-api/src/legal_api/resources/business/business_directors.py
```diff
@@ -55,10 +55,22 @@ def get(identifier, director_id=None):
                 del director_json['mailingAddress']
             res.append(director_json)
 
+        # party role code to replace director
+        party_list = []
+        active_directors = PartyRole.get_active_directors(business.id, end_date)
+        print(active_directors)
```
thorwolpert: should remove this, or put it into a debug stmnt

legal-api/src/legal_api/resources/business/business_directors.py
```diff
@@ -55,10 +55,22 @@ def get(identifier, director_id=None):
...
+        for director in active_directors:
+            print(director)
```
thorwolpert: same as above

legal-api/src/legal_api/resources/business/business_directors.py
```diff
@@ -55,10 +55,22 @@ def get(identifier, director_id=None):
...
+            director_json = director.json
+            print(director_json)
```
thorwolpert: same as above

legal-api/src/legal_api/resources/business/business_directors.py
```diff
@@ -55,10 +55,22 @@ def get(identifier, director_id=None):
...
+            if business.legal_type == 'CP':
+                del director_json['mailingAddress']
+            party_list.append(director_json)
+
         return jsonify(directors=res)
 
     @staticmethod
-    def _get_director(business, director_id=None):
+    def _get_director(business, director_id=None, party_id=None):
```
thorwolpert: I think this needs to get reworked to find either by director_id or party_id, then handle the return of nothing found, etc.

queue_services/entity-filer/src/entity_filer/filing_processors/__init__.py
```diff
@@ -48,20 +48,72 @@ def update_address(address: Address, new_info: dict):
     return address
 
 
-def update_director(director: Director, new_info: dict):
+def create_director(director_info: dict):
```
thorwolpert: Should this create the director in both old & new structures until after data migration and migrating the code to using parties?

legal-api/src/legal_api/resources/business/business_directors.py
kialj876: That would be nicer, but since this isn't going to last more than a couple days in our code base I don't think it's worth changing. I've already branched off to make changes for after this PR and in that one this method goes back to just the 1 director_id and searches the party/party_role tables. If it's important though I'll obvs change this

queue_services/entity-filer/src/entity_filer/filing_processors/__init__.py
kialj876: The director is also created as before in the processor. Since I'll be taking out the director table in my next PR and removing that code anyways I didn't think it was worth it to refactor adding the director into this method :P

PR #767, opened 2020-03-24
2820 db refactor people 4

kialj876:
*Issue #:* /bcgov/entity#2820

*Description of changes:*
- see ticket for dets
- removed director table and all director code
- replaced with party/party_role in all cases


kialj876:
> Looks good to me.
> 
> 'Completing Party' is one of the roles in the incorporation flow. That may have to be added to the RoleTypes enum in PartyRole Model

Right - that should be part of Camerons legal-api ticket for incorporation I think, this is just the refactor

PR #770, opened 2020-03-26
#2673 legal api step2

cameron-freshworks:
*Issue #:* /bcgov/entity#2673

*Description of changes:*
* Updated Schema version dependency.
* Implemented Validators for Step 2 People and Roles data.
* Pytests for additional validators and properties.

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/incorporation_application.py
+++ b/legal-api/src/legal_api/services/filings/validations/incorporation_application.py
@@ -90,30 +90,41 @@ def validate_roles(incorporation_json) -> Error:
     for item in parties_array:
         role_array = item['roles']
 
-        if role_array.count('Completing Party') == 1:
+        if 'Completing Party' in role_array:
             completing_party_count += 1
 
-        if role_array.count('Incorporator') == 1:
+        if 'Incorporator' in role_array:
             incorporator_count += 1
 
-        if role_array.count('Director') == 1:
+        if 'Director' in role_array:
             director_count += 1
 
     if completing_party_count == 0:
         err_path = '/filing/incorporationApplication/parties/roles'
         msg.append({'error': 'Must have a minimum of one completing party', 'path': err_path})
 
-    if completing_party_count >= 2:
+    if completing_party_count > 1:
         err_path = '/filing/incorporationApplication/parties/roles'
         msg.append({'error': 'Must have a maximum of one completing party', 'path': err_path})
 
-    if incorporator_count < 1:
-        err_path = '/filing/incorporationApplication/parties/roles'
-        msg.append({'error': 'Must have a minimum of one Incorporator', 'path': err_path})
+    if incorporation_json['filing']['incorporationApplication']['nameRequest']['legalType'] == 'BC':
+        if incorporator_count < 1:
+            err_path = '/filing/incorporationApplication/parties/roles'
+            msg.append({'error': 'Must have a minimum of one Incorporator', 'path': err_path})
 
-    if director_count < 1:
-        err_path = '/filing/incorporationApplication/parties/roles'
-        msg.append({'error': 'Must have a minimum of one Director', 'path': err_path})
+        if director_count < 1:
+            err_path = '/filing/incorporationApplication/parties/roles'
+            msg.append({'error': 'Must have a minimum of one Director', 'path': err_path})
+
+    if incorporation_json['filing']['incorporationApplication']['nameRequest']['legalType'] == 'CP':
+        if incorporator_count < 3:
+            err_path = '/filing/incorporationApplication/parties/roles'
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -69,3 +77,67 @@ def validate_offices(incorporation_json) -> Error:
...
+            incorporator_count += 1
+
+        if role_array.count('Director') == 1:
+            director_count += 1
+
+    if completing_party_count == 0:
+        err_path = '/filing/incorporationApplication/parties/roles'
+        msg.append({'error': 'Must have a minimum of one completing party', 'path': err_path})
+
+    if completing_party_count >= 2:
```
lekshmimallika-aot: Cant it be >1?

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -69,3 +77,67 @@ def validate_offices(incorporation_json) -> Error:
...
+    parties_array = incorporation_json['filing']['incorporationApplication']['parties']
+    msg = []
+    completing_party_count = 0
+    incorporator_count = 0
+    director_count = 0
+
+    for item in parties_array:
+        role_array = item['roles']
+
+        if role_array.count('Completing Party') == 1:
```
lekshmimallika-aot: If the purpose is to check whether the role array has completing party it can be done like
if 'Completing Party' in role_array:

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -69,3 +77,67 @@ def validate_offices(incorporation_json) -> Error:
...
+
+    if completing_party_count == 0:
+        err_path = '/filing/incorporationApplication/parties/roles'
+        msg.append({'error': 'Must have a minimum of one completing party', 'path': err_path})
+
+    if completing_party_count >= 2:
+        err_path = '/filing/incorporationApplication/parties/roles'
+        msg.append({'error': 'Must have a maximum of one completing party', 'path': err_path})
+
+    if incorporator_count < 1:
```
lekshmimallika-aot: These rules are applicable for BCOMP. For CP it is 3 I believe. The director count for CP is also for 3 I think. Even though we are not implementing CP now, it would be better to have that check around and add a TODO comment or so for CP

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -69,3 +77,67 @@ def validate_offices(incorporation_json) -> Error:
...
+
+def validate_parties_mailing_address(incorporation_json) -> Error:
+    """Validate the person data of the incorporation filing."""
+    parties_array = incorporation_json['filing']['incorporationApplication']['parties']
+    msg = []
+
+    for item in parties_array:
+        for k, v in item['mailingAddress'].items():
+            if v is None:
+                err_path = '/filing/incorporationApplication/parties/mailingAddress/%s/%s/' % (
```
lekshmimallika-aot: If we have multiple parties with some invalid mailing address should we provide an identifier like party name or index or something to identify which fields are invalid for which party

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: Absolutely :) Will update!

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: oh nice, i like that much better! Will update for all roles 👍

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: Yes, your are 100% correct. I was thinking about that but wasn't sure if i should implement something now if we aren't going to use it. I will wrap these conditions in a check tho for BCOMP and add a TODO for CP :) Thanks!

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: Great idea, will look at this!

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: Edit* Just added the checks for both types in 👍

cameron-freshworks:
Have 2 approvals and comments have been addressed. Merging!

PR #774, opened 2020-03-30
2903 update config, error handling, name request proxy endpoint

ochiu:
*Issue #:* /bcgov/entity#2903

*Description of changes:*
- Updates to allow for proxying of name requests calls from legal-api to namex-api
- Added config for getting a JWT for namex-api as it lives in a different keycloak realm
- Name request proxy endpoint


ochiu:
@peter-freshworks Thanks! updated the copyright text. Not sure if this one requires a test as it is a proxy call to namex api? Do you know of an example that exists? Maybe I can base it off of that if you know of any.

legal-api/tests/unit/api/test_name_requests.py
```diff
@@ -0,0 +1,36 @@
...
+# limitations under the License.
+
+"""Tests to assure the name requests end-point.
+
+Test-Suite to ensure that the /nameRequests endpoint is working as expected.
+"""
+from http import HTTPStatus
+
+
+def test_name_requests_success(client):
```
thorwolpert: This should get wrapped with an 'integration", you'll need to create one in tests/pytest_marks.py and add it to __init__.py

legal-api/tests/unit/api/test_name_requests.py
```diff
@@ -0,0 +1,36 @@
...
+    """Assert that a name request can be received."""
+    rv = client.get('/api/v1/nameRequests/NR 3252362')
+
+    assert rv.status_code == HTTPStatus.OK
+    assert 'nrNum' in rv.json
+    assert rv.json['nrNum'] == 'NR 3252362'
+
+
+def test_name_requests_not_found(client):
```
thorwolpert: same here as above ^^

legal-api/tests/unit/api/test_name_requests.py
ochiu: Thanks! updated.

legal-api/tests/unit/api/test_name_requests.py
ochiu: Done!

PR #777, opened 2020-03-31
filer for applying party info in incorporation - also fixed incorp

kialj876:

*Issue #:* /bcgov/entity#2821

*Description of changes:*
- added new role type in legal api
- fixed incorp on filer (was failing)
- added applying party roles to the db in incorp filings
- refactored updating directors a bit to treat directors and parties similarly (as they are the same thing)
- updated tests
- updated reqs
- updated param file in k8s to include colin url (will also add in openshift)


queue_services/entity-filer/requirements.txt
```diff
@@ -8,17 +8,17 @@ attrs==19.1.0
...
-pyrsistent==0.15.7
+pyrsistent==0.16.0
 python-dotenv==0.12.0
-sentry-sdk==0.14.2
+sentry-sdk==0.14.3
 six==1.14.0
 urllib3==1.25.8
 zipp==3.1.0
-git+https://github.com/bcgov/business-schemas.git@2.5.5#egg=registry_schemas
+git+https://github.com/bcgov/business-schemas.git@2.5.8#egg=registry_schemas
```
lekshmimallika-aot: Should this be 2.5.9 now

queue_services/entity-filer/src/entity_filer/filing_processors/change_of_directors.py
```diff
@@ -57,7 +57,13 @@ def process(business: Business, filing: Dict):  # pylint: disable=too-many-branc
         if 'appointed' in new_director['actions']:
 
             # add new diretor party role to the business
-            new_director_role = create_director(director_info=new_director)
+            party = create_party(party_info=new_director)
+            role = {
+                'roleType': 'director',
```
lekshmimallika-aot: Should this be 'Director' to match the key in JSON_ROLE_CONVERTER. Not sure whether get by key is case sensitive

thorwolpert: it should probably be the Enum value?

kialj876: Oh no actually it needs to be 'Director' -> that's the value the FE decided to use for this roletype and in the util I convert it into the enum value (that's why I made the JSON_ROLE_CONVERTER)

queue_services/entity-filer/requirements.txt
kialj876: Oh good catch I'll update that too

PR #783, opened 2020-04-07
bug fix for 2932

peter-freshworks:
*Issue #:* /bcgov/entity#2932

*Description of changes:*
Default to next anniversary year if no annual reports exist


peter-freshworks:
> Looks good. May be good to have a test case that covers this scenario

added a unit test :)

PR #785, opened 2020-04-14
1940 country bug

peter-freshworks:
*Issue #:* /bcgov/entity#1940

*Description of changes:*
Retrieve full country name for report display

### DIFF
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -17,6 +17,7 @@ from datetime import datetime
 from http import HTTPStatus
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/reports/report.py
```diff
@@ -61,6 +62,15 @@ def _get_report_filename(self):
     def _get_filing_description(self):
         return self._get_primary_filing()['title']
 
+    def _format_directors(self, directors):
+        for director in directors:
+            try:
```
lekshmimallika-aot, line 67: As per the last discussion all formatting will be moved to template and I think this should move to the respective template

PR #786, opened 2020-04-16
3102 - Incorporate NR todo item

ochiu:
*Issue #:* /bcgov/entity#3102

*Description of changes:*
- Implemented namex service for re-usable code when querying for NR data
- Updated name request proxy endpoint to use namex service
- Added function and logic to validate NR data
- Updated filings endpoint to return empty filings array when business not found
- Updated tasks endpoint to return empty tasks array when business not found
- Added in NR number check for tasks endpoint and return todo item if no draft filing exists
- Added in error handling when NR is invalid to return 403
- Added in a function for specifically creating a todo item for incorporating NR
- Added unit tests to cover the different NR data validations

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_tasks.py
+++ b/legal-api/src/legal_api/resources/business/business_tasks.py
@@ -15,6 +15,7 @@
 
 Provides all the search and retrieval from the business filings datastore.
 """
+
 from datetime import datetime
 from http import HTTPStatus
 
@@ -23,6 +24,7 @@ from flask import jsonify
 from flask_restplus import Resource, cors
 
 from legal_api.models import Business, Filing
+from legal_api.services import namex
 from legal_api.services.filings import validations
 from legal_api.utils.util import cors_preflight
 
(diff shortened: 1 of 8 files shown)


legal-api/src/legal_api/resources/business/business_tasks.py
```diff
@@ -39,11 +41,31 @@ class TaskListResource(Resource):
     def get(identifier):
         """Return a JSON object with meta information about the Service."""
         business = Business.find_by_identifier(identifier)
+        is_nr = identifier.startswith('NR')
```
peter-freshworks, line 44: curious, could we use the namex service here instead to determine if this is a business coming from NR? I don't think we can guarantee new companies' identifiers will start with 'NR' (I know we have this validation elsewhere, so maybe we keep it this way for now?)

ochiu: I think this will change once we solidify how we are handling identifiers, but currently the UI feeds in NR prefixed string. So we will probably address this in a future PR.

PR #790, opened 2020-04-20
fixes for 2672

peter-freshworks:
*Issue #:* /bcgov/entity#2672

*Description of changes:*
In adding additional checks for roles and share classes in postman tests, I noticed a bug with the response for incorporation filing. This ticket addresses both the missing test cases as well as the coalescing of parties based on roles (as COLIN stores a party with multiple roles as multiple parties) 


colin-api/src/colin_api/models/corp_party.py
```diff
@@ -104,8 +106,27 @@ def _build_parties_list(cls, cursor, event_id: int = None):
...
+
+        for k, v in parties_dict.items():  # pylint: disable=unused-variable;
+            party = v[0]
+            roles = []
+            for i in v:
+                role = i.role_type
+                roles.append({'roleType': role_dict[role]})
+            party.roles = roles
+            grouped_list.append(party)
+        return grouped_list
```
kialj876, line 129: It's pretty hard to read this because of the variable names used here (x , k, v don't give me any idea of what you're looping through etc.). Would help a lot if you added in a couple short comments to describe what's going on

colin-api/src/colin_api/models/corp_party.py
```diff
@@ -104,8 +106,27 @@ def _build_parties_list(cls, cursor, event_id: int = None):
                     party.cessation_date = None
 
                 party_list.append(party)
+        return cls.group_parties(party_list)
 
-        return party_list
+    @classmethod
+    def group_parties(cls, parties):
```
kialj876, line 112: can you give parties a type pls

colin-api/src/colin_api/models/corp_party.py
peter-freshworks: should be better now  (the k,v are short for key, value, and the x just represents the captured element in the lambda)...most other languages I've worked with seem to encourage this type of convention for inline/lambda functions

colin-api/src/colin_api/models/corp_party.py
peter-freshworks: done!

colin-api/src/colin_api/models/corp_party.py
kialj876: Oh I see that makes sense. Thanks for adding the documentation

kialj876:
> Tests all pass, but I believe it's returning the wrong format for incorporation (doesn't match our schema). The appointment/cessation dates should be within the roles correct?

The cessation/appointment date are now in the roles, but they are also still in the level above in the json - please remove those. Also the return json from the post is missing the completing party role -> looks like a small bug where it's not returning all the roles for each party

PR #795, opened 2020-04-22
initial commit of refactor.md

peter-freshworks:
*Issue #:* /bcgov/entity#2544

Write up describing next steps, with a few questions open for input

### DIFF
--- a/data-reset-tool/Refactor.md
+++ b/data-reset-tool/Refactor.md
@@ -1,0 +1,54 @@
+
+## Objective
+
+The goal is to decouple the logic in the data reset tool from the models in legal API. This will result in a simpler, easier to maintain tool with no dependencies on the code in legal-api. We do not want to change the endpoints or the reset tool's API structure too much to reduce the amount of refactoring that has to be done in the nightwatch scripts.   
+
+## Motivation
+
+The models and the database structure in LEAR change frequently. If a change goes in without being reflected in the data reset tool, there could either be missed data or catastrophic failures. Decoupling the data reset tool from the models in legal-api would result in more reliability and predicatability within the data reset tool. By using postgres' built in bulk copy methods we should also see a performance improvement over the current design.   
+
+## Design Proposal
+
+The proposal is to gut the dependencies on the legal-api models from the current data-reset tool code and replace them with native postgres bulk import and export mechanisms (copy and copy_expert). Postgres has the ability to map columnar data in CSV format to tables. This would change the current code structure in the following ways:
+
+* The JSONConverter class would either no longer be needed, or would be changed to not call methods of the entities. A more agnostic approach may be to use the GET methods in legal API to retrieve entity data.
+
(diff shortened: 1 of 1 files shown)


data-reset-tool/Refactor.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Design Proposal
+
+The proposal is to gut the dependencies on the legal-api models from the current data-reset tool code and replace them with native postgres bulk import and export mechanisms (copy and copy_expert). Postgres has the ability to map columnar data in CSV format to tables. This would change the current code structure in the following ways:
+
+* The JSONConverter class would either no longer be needed, or would be changed to not call methods of the entities. A more agnostic approach may be to use the GET methods in legal API to retrieve entity data.
+
+* The folder structure where the test data is held (e2e/lear-data) would change to have a folder per corp num. Each folder would have a csv file per table (ie CP0000001/directors.csv) since bulk copy cannot work with multiple tables.
```
riyazuddinsyed, line 16: Is there anyway just to restore a group of businesses from the back up avoiding the pain of managing and maintaining the  CSV files

peter-freshworks: From a db backup? Of course, but that would remove the ability to set custom scenarios. Or did you mean reduce the number of CSV files? I.e - have multiple businesses in a business.csv instead of a business.csv per corp

data-reset-tool/Refactor.md
```diff
@@ -0,0 +1,54 @@
...
+
+* The removal of the legal-api.models dependency
```
kialj876, line 18: I think it's important to keep using the model (it prevents our tests inserting garbage data into the db and if we change the database then we'll have to update the code inserting into the database without the model anyways).
1) Could we instead just add an endpoint to the legal-api that allows you to insert the data using the model? This would allow us to get rid of the data-reset-tool altogether and force devs to update the methods on any refactor. 
2) The data_loader accomplishes loading the database with business info, addresses, directors in a much more condensed way - could we alter/combine that with the data reset tool to accomplish both things?
-- We could accomplish either of these very quickly if we weren't using excel/csv files to populate our data with the reset tool. Are we only using those so that QA can make test cases easier? We could make the json templates for them to make it simple and easy to add (similar to the json/param files we use for creating objects in openshift). If we were using json templates QA could put their test json data with the example data in the schema repo too, which would then get updated by anyone who needed to change the schema keeping everything in sync.
Something along these lines is probably what I would want to do, thoughts?

peter-freshworks: Using the loader would limit QA to whatever is already in COLIN, and if records were modified between runs then they no longer have the benefit of having a guaranteed known state. I don't hate the idea of using JSON templates, but I can see how working with JSON might be a bit less friendly VS. csv files. I also don't mind just adding this to legal-api, that seems like a way to kill two birds with one stone. One of the main issues with the data reset tool is that we want to decouple the model from the tool

kialj876: Yeah if we were going to combine the loader we would have to add in functionality for it to read in json/csv files, but we could reuse all the methods for inserting data into the db, and since it's code we have to update anyways it would stay up to date. Probably easier to just add something into the legal-api though

PR #798, opened 2020-04-28
add that a JWT and read authorization is required for GET

thorwolpert:
*Issue #:* n/a

*Description of changes:*
add that a JWT and read authorization is required for GET

### DIFF
--- a/legal-api/src/legal_api/resources/business/business.py
+++ b/legal-api/src/legal_api/resources/business/business.py
@@ -38,8 +38,15 @@ class BusinessResource(Resource):
 
     @staticmethod
     @cors.crossdomain(origin='*')
+    @jwt.requires_auth
     def get(identifier):
         """Return a JSON object with meta information about the Service."""
+        # check authorization
+        if not authorized(identifier, jwt, action=['view']):
+            return jsonify({'message':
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/resources/business/business.py
```diff
@@ -38,8 +38,15 @@ class BusinessResource(Resource):
 
     @staticmethod
     @cors.crossdomain(origin='*')
+    @jwt.requires_auth
     def get(identifier):
         """Return a JSON object with meta information about the Service."""
+        # check authorization
+        if not authorized(identifier, jwt, action=['view']):
+            return jsonify({'message':
+                            f'You are not authorized to submit a filing for {identifier}.'}), \
```
lekshmimallika-aot, line 47: Is the message 'submit a filing' ok for a get endpoint with view authorization?

thorwolpert: Good point, switched that to:
f'You are not authorized to view business {identifier}.'

thorwolpert: Good point, switched that to:
f'You are not authorized to view business {identifier}.'

PR #799, opened 2020-04-28
1891 handle multiple names

kialj876:
*Issue #:* /bcgov/entity#1891

*Description of changes:*
- added changes for allowing multiple parties with the same exact name across different corps
- tests added to secure this
- added md file to explain the process for data fixing the parties that were possibly copied over incorrectly (as result of them having the exact same names) when the directors table was deleted

### DIFF
--- a/legal-api/src/legal_api/models/party.py
+++ b/legal-api/src/legal_api/models/party.py
@@ -100,11 +100,6 @@ class Party(db.Model):  # pylint: disable=too-many-instance-attributes
     @property
     def name(self) -> str:
         """Return the full name of the party for comparison."""
-        print(self.first_name)
-        print(self.middle_initial)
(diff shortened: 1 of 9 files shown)


legal-api/src/legal_api/models/party.py
```diff
@@ -97,6 +97,20 @@ def json(self) -> dict:
 
         return member
 
+    @property
+    def name(self) -> str:
+        """Return the full name of the party for comparison."""
+        print(self.first_name)
+        print(self.middle_initial)
```
lekshmimallika-aot, line 104: Are these print statements needed?

kialj876: oops I'll delete those!

PR #802, opened 2020-04-29
3506 - integration tests collection with pre-request script for staff token

ochiu:
*Issue #:* /bcgov/entity#3506

*Description of changes:*
- Created file structure for integration tests (/tests/postman since it will span different components)
- Added postman collection with pre-request script that gets a staff token (script is based on pay-api collection script that does the same thing)

### DIFF
--- a/tests/postman/lear-integration.postman_collection.json
+++ b/tests/postman/lear-integration.postman_collection.json
@@ -1,0 +1,83 @@
+{
+	"info": {
+		"_postman_id": "ef97304d-d447-4c55-9de8-a672a2c6f01f",
+		"name": "lear-integration",
+		"description": "LEAR integration tests collection.",
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+	},
+	"item": [],
+	"auth": {
+		"type": "oauth2",
+		"oauth2": [
+			{
+				"key": "accessToken",
+				"value": "{{accessToken}}",
+				"type": "string"
+			},
+			{
+				"key": "addTokenTo",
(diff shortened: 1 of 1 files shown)


severinbeauvais:
> * Created file structure for integration tests (/tests/postman since it will span different components)

It may span different components, or components may need their own. Not sure. (Eg, need auth to call COLIN API?)

I think John mentioned that a new folder should be created for tests specific to incorp apps. So I added him as a reviewer for his input :)

severinbeauvais:
How would a reviewer know that this code is correct, and/or test it?

ochiu:
> How would a reviewer know that this code is correct, and/or test it?

The collection at the current basic state required 3 environment variables
1) staff-token-url
2) staff-service-account-id
3) staff-service-account-secret

I then added a GET request to the collection (just did localhost as the url). It will error since it isn't really a valid url, but the postman console should show the request with the authorization headers as well as the pre-request script executing and outputting the access token.

JohnamLane:
> > * Created file structure for integration tests (/tests/postman since it will span different components)
> 
> It may span different components, or components may need their own. Not sure. (Eg, need auth to call COLIN API?)
> 
> I think John mentioned that a new folder should be created for tests specific to incorp apps. So I added him as a reviewer for his input :)

There's no perfect answer for where to locate a test suite that spans multiple services. Most important would be getting everyone on the same page. The browser tests are in entity since they cover all services, but that repo lacks the GitHub actions that lear has and may be moving soon. On the OpenShift side, pipelines can invoke other pipelines across namespaces, so no issues there.

PR #805, opened 2020-05-01
Payment related changes for incorporation

lekshmimallika-aot:
*Issue #:* /bcgov/entity#3527,  /bcgov/entity#3540

*Description of changes:*
Changed the filing type code for Incorporation
Added the futureEffective flag to filing type for Incorporation
Removed the method of payment from the create invoice payload

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -351,7 +351,7 @@ class ListFilingResource(Resource):
             # for any legal type, set effective date as set in json; otherwise leave as default
             filing.effective_date = \
                 datetime.datetime.fromisoformat(filing.filing_json['filing']['header']['effectiveDate']) \
-                if filing.filing_json['filing']['header'].get('effectiveDate', None) else None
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -411,6 +411,11 @@ def _get_filing_types(filing_json: dict):
                     'priority': False if filing_type == 'annualReport' else priority_flag,
                     'waiveFees': filing_json['filing']['header'].get('waiveFees', False)
                 })
+            elif k == 'incorporationApplication':
+                filing_types.append({
+                    'filingTypeCode': Filing.FILINGS[k].get('code'),
+                    'futureEffective': ListFilingResource._is_future_effective_filing(filing_json)
```
kialj876, line 417: Should the future effective flag be added to the other filing types too? Also should incorporationApplication include the priority + waiveFees options too?

lekshmimallika-aot: By default the value is false. We need to add that only if it has a fee related to future effective.
Also, we dont have a priority + waive fee option for incorporation. I think we can add that if that feature is supported

PR #807, opened 2020-05-05
3536 - BCOL Payment error messages

ochiu:
*Issue #:* /bcgov/entity#3536

*Description of changes:*
- added payment_error_type column to filings to capture the the type of payment error used to query error details from pay-api
- added logic to return payment error type on the header so dashboard can fetch the error and display as needed
- added logic to clear out payment error type on successful payment

### DIFF
--- a/legal-api/migrations/versions/54a3fc54a2cf_.py
+++ b/legal-api/migrations/versions/54a3fc54a2cf_.py
@@ -1,0 +1,24 @@
+"""empty message
+
+Revision ID: 54a3fc54a2cf
+Revises: 442829649a2a
+Create Date: 2020-05-05 07:49:31.147648
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '54a3fc54a2cf'
+down_revision = '442829649a2a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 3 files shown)


ochiu:
> Looks good! One unit test on a filing that fails and checks that the error is set would be good though. And then also if it checks that it successfully clears the error after payment being successful. Not sure if any tests like this already exist or not

I was looking into this, I am not sure how to emulate the pay api response to cause this error though as we had to modify data in the database to even generate an error. I did test this with Sumesh on my local build doing payments to ensure the error is set and returned as well as cleared on a successful payment.

kialj876:
> > Looks good! One unit test on a filing that fails and checks that the error is set would be good though. And then also if it checks that it successfully clears the error after payment being successful. Not sure if any tests like this already exist or not
> 
> I was looking into this, I am not sure how to emulate the pay api response to cause this error as we had to modify data in the database to even generate an error. I did test this with Sumesh on my local build doing payments to ensure the error is set and returned as well as cleared on a successful payment.



> > Looks good! One unit test on a filing that fails and checks that the error is set would be good though. And then also if it checks that it successfully clears the error after payment being successful. Not sure if any tests like this already exist or not
> 
> I was looking into this, I am not sure how to emulate the pay api response to cause this error as we had to modify data in the database to even generate an error. I did test this with Sumesh on my local build doing payments to ensure the error is set and returned as well as cleared on a successful payment.

Ah okay well looks good to me

PR #808, opened 2020-05-05
Implemented Future Effective Validators

cameron-freshworks:
*Issue #:* /bcgov/entity#2131

*Description of changes:*
* Implemented validators an effective date for future effective incorporation filings.
* Updated unit testing.

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/incorporation_application.py
+++ b/legal-api/src/legal_api/services/filings/validations/incorporation_application.py
@@ -14,12 +14,13 @@
 """Validation for the Incorporation filing."""
 from http import HTTPStatus  # pylint: disable=wrong-import-order
 from typing import Dict
+from datetime import timedelta
 
 import pycountry
-
 from flask_babel import _ as babel  # noqa: N813, I004, I001, I003
-from legal_api.errors import Error
 
+from legal_api.errors import Error
+from legal_api.utils.datetime import datetime as dt
 
 def validate(incorporation_json: Dict):
     """Validate the Incorporation filing."""
@@ -40,6 +41,10 @@ def validate(incorporation_json: Dict):
         msg.append(err)
 
     err = validate_share_structure(incorporation_json)
+    if err:
+        msg.append(err)
+
+    err = validate_incorporation_effective_date(incorporation_json)
     if err:
         msg.append(err)
 
@@ -181,6 +186,45 @@ def validate_share_structure(incorporation_json) -> Error:
                                      % (series['name'], item['name']),
                             'path': '%s/maxNumberOfShares' % err_path
                         })
+
+    if msg:
+        return msg
+
+    return None
+
+
+def validate_incorporation_effective_date(incorporation_json) -> Error:
+    """Return an error or warning message based on the effective date validation rules.
+
+    Rules:
+        - The effective date must be the correct format.
+        - The effective date must be a minimum of 2 minutes in the future.
+        - The effective date must be a maximum of 10 days in the future.
+    """
+    # Setup
+    msg = []
+    now = dt.utcnow()
+    now_plus_2_minutes = now + timedelta(minutes=2)
+    now_plus_10_days = now + timedelta(days=10)
+
+    try:
+        filing_effective_date = incorporation_json['filing']['header']['effectiveDate']
+    except KeyError:
+        return msg
+
+    try:
+        effective_date = dt.fromisoformat(filing_effective_date)
+    except ValueError:
+        msg.append({'error': babel('%s is an invalid ISO format for effective_date.') % filing_effective_date})
+        return msg
+
+    if effective_date < now_plus_2_minutes:
+        msg.append({'error': babel('Invalid Datetime: %s, effective date must be a minimum of 2 minutes ahead.')
+                             % filing_effective_date})
+
+    if effective_date > now_plus_10_days:
+        msg.append({'error': babel('Invalid Datetime: %s, effective date must be a maximum of 10 days ahead.')
+                             % filing_effective_date})
 
     if msg:
         return msg
--- a/legal-api/src/legal_api/services/filings/validations/incorporation_application.py
+++ b/legal-api/src/legal_api/services/filings/validations/incorporation_application.py
@@ -221,11 +221,11 @@ def validate_incorporation_effective_date(incorporation_json) -> Error:
 
     if effective_date < now_plus_2_minutes:
         msg.append({'error': babel('Invalid Datetime: %s, effective date must be a minimum of 2 minutes ahead.')
-                    % filing_effective_date})
+                             % filing_effective_date})
 
     if effective_date > now_plus_10_days:
         msg.append({'error': babel('Invalid Datetime: %s, effective date must be a maximum of 10 days ahead.')
-                    % filing_effective_date})
+                             % filing_effective_date})
 
     if msg:
         return msg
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -186,3 +192,40 @@ def validate_share_structure(incorporation_json) -> Error:
...
+
+    Rules:
+        - The effective date must be the correct format.
+        - The effective date must be a minimum of 2 minutes in the future.
+        - The effective date must be a maximum of 10 days in the future.
+    """
+    # Setup
+    msg = []
+    now = dt.utcnow()
+    now_plus_2_minutes = now + timedelta(minutes=2)
```
lekshmimallika-aot, line 208: Is this working fine if you choose exactly 2 mins in the UI?

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -186,3 +192,40 @@ def validate_share_structure(incorporation_json) -> Error:
...
+    now_plus_10_days = now + timedelta(days=10)
+
+    try:
+        filing_effective_date = incorporation_json['filing']['header']['effectiveDate']
```
lekshmimallika-aot, line 212: Are we passing effective date in case of immediate filing? Is the validation passing for immediate filing

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: The front end won't let you File And Pay if it is any sooner than 2 minutes, yea.

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
lekshmimallika-aot: Also I think the legal API returns the effective date in the filing header when filing is resumed ( after save). It would be a good check to see whether these validations are passing in the second submit

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: For example, if you select exactly 2 minutes ahead, but take 5 minutes to file. You will have to update that time before filing and paying.

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: So, correct me if i'm wrong, but if it is immediate, effective date is being passed as `null` or `None` to the backend. This check should return the validations with no errors if that is the case. It is the second test scenario. ( At least those were my intentions :) )

cameron-freshworks: oh interesting! How could i go about testing that..

legal-api/tests/unit/api/test_business_incorporation.py
```diff
@@ -66,51 +75,54 @@ def test_post_new_incorporation(session, client, jwt):
...
-    assert filing_id == rv.json['filing']['header']['filingId']
+        assert HTTPStatus.OK == rv.status_code
+        assert filing_id == rv.json['filing']['header']['filingId']
 
 
 def test_put_draft_incorporation_filing(session, client, jwt):
     """Assert that an incorporation filing can be put (updated) to filings endpoint."""
-    nr_number = 'NR 1234567'
     filing_json = copy.deepcopy(INCORPORATION_FILING_TEMPLATE)
     filing_json['filing']['incorporationApplication']['nameRequest']['nrNumber'] = nr_number
```
kialj876, line 127: How is nr_number getting assigned now? It looks like this should cause an error to me, am I missing something?

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
lekshmimallika-aot: utcnow will always be ahead of the submission time by milliseconds or so due to network latency and stuff right? I think there is achance that exact 2 mins from the UI may cause a validation failure. Could you please test it out from UI

cameron-freshworks: Good point for sure, i know this is an extreme edge case ( someone paying $100 to have it effective in 2 mins from now 😆 ), so should we adjust the time and give a buffer on the back end. It could potentially pass front end validation and then fail the validator during the request.

The front end will not make the request if it is < 2 mins, but you're right, we should take into account the latency, would adding an extra minute be acceptable?

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -186,3 +192,40 @@ def validate_share_structure(incorporation_json) -> Error:
...
+    except KeyError:
+        return msg
+
+    try:
+        effective_date = dt.fromisoformat(filing_effective_date)
+    except ValueError:
+        msg.append({'error': babel('%s is an invalid ISO format for effective_date.') % filing_effective_date})
+        return msg
+
+    if effective_date < now_plus_2_minutes:
```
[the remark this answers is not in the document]
cameron-freshworks: The front end actually stores and makes the request with `effectiveDate` it in UTC('IsoString').
 Just from working with DateTimes earlier, we knew that since the backend was handling it in this format, it was simpler to avoid the time conversions. We just convert convert the DateTime when presenting it in the UI.

legal-api/tests/unit/api/test_business_incorporation.py
cameron-freshworks: I defined `nr_number` as the `indentifier` at the beginning of the test setup, and assigned that identifier to the filing nrNumber in each test. Basically just defined the NR once opposed to defining it in each test. 

see line 31 for the definition

an example of how i assigned it in each test.
`f['filing']['incorporationApplication']['nameRequest']['nrNumber'] = identifier`

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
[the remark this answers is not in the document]
kialj876: awesome

legal-api/tests/unit/api/test_business_incorporation.py
kialj876: Oh cool! Nice

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
lekshmimallika-aot: Also, this particular one need not block the PR as such and the fix (if any) can be a separate PR

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
lekshmimallika-aot: Try 'save and resume' for an immediate filing. In the second submission(dont change anything just click on file and pay ) if it passes, it is fine

cameron-freshworks: kk, will do thanks!

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
cameron-freshworks: Absolutely, will look into this, thanks Lekshmi!

PR #811, opened 2020-05-08
3639 - update set effective date logic to take into account local time offsets

ochiu:
*Issue #:* /bcgov/entity#3639

*Description of changes:*
- updated set effective date logic to take into account local time offset


ochiu:
I confirmed a date of 2020-05-08 returned the following date time `2020-05-09T07:00:00+00:00`.

PR #820, opened 2020-05-20
3508 run colin updater

kialj876:
*Issue #:* /bcgov/entity#3508

*Description of changes:*
- added to the job pipeline jenkinsfile
    - deletes prev job if existing
    - creates new job with k8s json template
    - verifies job ran and is successful (this also makes it wait until the job is finished)
- renamed job stages appropriately

### DIFF
--- a/Jenkinsfiles/run-job-pipeline.groovy
+++ b/Jenkinsfiles/run-job-pipeline.groovy
@@ -69,19 +69,18 @@ stage("Verify ${JOB} success") {
     script {
         openshift.withCluster() {
             openshift.withProject("${NAMESPACE}-${TAG_NAME}") {
-                    // below is code for verifying dataloader ran -> repurpose for colin-updater
-                // def data_loader = openshift.selector('pod', [ "job-name":"data-loader" ])
-                // data_loader.untilEach {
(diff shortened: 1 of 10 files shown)


Jenkinsfiles/run-job-pipeline.groovy
```diff
@@ -41,48 +41,46 @@ stage("NI: Run ${JOB}") {
...
-                           // job doesn't exist yet (hopefully)
-                    //     echo "${e.getMessage()}"
-                    //     data_load_output = sh (
-                    //     script: """oc process -f data-loader.yml -p ENV_TAG=test | oc create -f -""",
-                    //         returnStdout: true).trim()
-                    // }
+                    try {
+                        delete_job = sh (
+                            script: """
+                            oc project ${NAMESPACE}-${TAG_NAME}; \
```
severinbeauvais, line 47: Where does this shell run actually?
And I gather it has the oc tools installed already.

kialj876: yes. It's just running within the Jenkins pod that spins up when the pipeline is run

PR #827, opened 2020-05-26
changed corpType to TMP, placed NR# in name if present, updated error handling now that deletes are now available

thorwolpert:
*Issue #:* /bcgov/entity#3779

*Description of changes:*
- change corpType on affiliation to TMP, until we have lookups for all filing types
- add NR# to the name field, if an NR is provided
- add error handling for failed affiliations

### DIFF
--- a/legal-api/src/legal_api/resources/business/business.py
+++ b/legal-api/src/legal_api/resources/business/business.py
@@ -15,6 +15,7 @@
 
 Provides all the search and retrieval from the business entity datastore.
 """
+from contextlib import suppress
 from http import HTTPStatus
 
 from flask import jsonify, request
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/resources/business/business.py
```diff
@@ -77,5 +78,14 @@ def post():
         if not isinstance(bootstrap, RegistrationBootstrap):
             return {'error': babel('Unable to create Incorporation Filing.')}, HTTPStatus.SERVICE_UNAVAILABLE
 
-        RegistrationBootstrapService.register_bootstrap(filing_account_id, bootstrap)
+        try:
+            business_name = json_input['filing']['incorporationApplication']['nameRequest']['nrNumber']
```
severinbeauvais, line 82: The namex object name this prop "nrNum".

legal-api/src/legal_api/services/bootstrap.py
```diff
@@ -118,7 +123,7 @@ def create_affiliation(cls, account: int, business_registration: str, business_n
 
         # Create an entity record
         entity_data = json.dumps({'businessIdentifier': business_registration,
-                                  'corpTypeCode': 'NR',  # to be deprecated
+                                  'corpTypeCode': 'TMP',  # to be deprecated
```
severinbeauvais, line 126: Is comment now obsolete?

legal-api/src/legal_api/resources/business/business.py
severinbeauvais: Should we change our code so it's the same?

lekshmimallika-aot: Right now, we pass a only three properties from the name request ( nrNumber, legalName and legalType). If we make a schema change, we have to make the corresponding changes in all other components as well like filer, reports etc,

severinbeauvais: Thor said we're going with "nrNumber" everywhere.

PR #828, opened 2020-05-27
3766 - add folio number to invoice payload

ochiu:
*Issue #:* /bcgov/entity#3766

*Description of changes:*
- add folio number to invoice payload

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -479,7 +479,7 @@ class ListFilingResource(Resource):
         return filing_types
 
     @staticmethod
-    def _create_invoice(business: Business,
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -479,7 +479,7 @@ def _get_filing_types(filing_json: dict):
         return filing_types
 
     @staticmethod
-    def _create_invoice(business: Business,
+    def _create_invoice(business: Business,  # pylint: disable=too-many-locals
```
ochiu, line 482: I wanted to keep with the pattern within this function, so I disabled this pylint error.

ochiu:
SonarCloud analysis seems to be failing on some code not part of this PR. Let me know if it needs to be fixed, not familiar with that piece of code.

PR #838, opened 2020-06-02
3598-documents-list

ochiu:
*Issue #:* /bcgov/entity#3598

*Description of changes:*
- add documents meta to filings json
- add document meta service to generate array of documents

### DIFF
--- a/legal-api/src/legal_api/services/document_meta.py
+++ b/legal-api/src/legal_api/services/document_meta.py
@@ -1,0 +1,154 @@
+# Copyright © 2020 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This provides the service for filings documents meta data."""
+
+from enum import Enum
+
+from legal_api.models import Filing
+from legal_api.utils.legislation_datetime import LegislationDatetime
+
+
+class DocumentMetaService():
+    """Provides service for document meta data."""
+
+    class DocumentType(Enum):
+        """Define an enum of document types."""
+
+        REPORT = 'REPORT'
+
+    class ReportType(Enum):
+        """Define an enum of report types."""
+
+        CERTIFICATE = 'certificate'
+        NOTICE_OF_ARTICLES = 'noa'
+
+    def get_documents(self, filing: dict):
+        """Return an array of document meta for a filing."""
+        business_identifier = filing['filing']['business']['identifier']
+        filing_id = filing['filing']['header']['filingId']
+        filing_date = filing['filing']['header']['date']
+        filing_status = filing['filing']['header']['status']
+        filing_type = filing['filing']['header']['name']
+        paper_only = filing['filing']['header']['availableOnPaperOnly']
+        documents = []
+
+        if filing_status not in (Filing.Status.COMPLETED.value, Filing.Status.PAID.value) or paper_only:
+            return []
+
+        if filing_type == 'annualReport':
+            documents = self.get_annual_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'changeOfAddress':
+            documents = self.get_coa_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'changeOfDirectors':
+            documents = self.get_cod_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'changeOfName':
+            documents = self.get_con_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'specialResolution':
+            documents = self.get_special_resolution_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'voluntaryDissolution':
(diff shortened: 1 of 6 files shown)


ochiu:
Please don't merge, found some issues testing IAs

legal-api/src/legal_api/services/document_meta.py
```diff
@@ -0,0 +1,154 @@
...
+    class DocumentType(Enum):
+        """Define an enum of document types."""
+
+        REPORT = 'REPORT'
+
+    class ReportType(Enum):
+        """Define an enum of report types."""
+
+        CERTIFICATE = 'certificate'
+        NOTICE_OF_ARTICLES = 'noa'
```
severinbeauvais, line 35: Please add the other 2 report types.

Also I wonder whether we should be more specific about the entity type (BC, CP, etc), for future?

legal-api/src/legal_api/services/document_meta.py
```diff
@@ -0,0 +1,154 @@
...
+from legal_api.utils.legislation_datetime import LegislationDatetime
+
+
+class DocumentMetaService():
+    """Provides service for document meta data."""
+
+    class DocumentType(Enum):
+        """Define an enum of document types."""
+
+        REPORT = 'REPORT'
```
severinbeauvais, line 29: We now have 3 words for the same thing (output, document, report). Is it possible to choose just one?

legal-api/src/legal_api/services/document_meta.py
```diff
@@ -0,0 +1,154 @@
...
+        elif filing_type == 'changeOfDirectors':
+            documents = self.get_cod_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'changeOfName':
+            documents = self.get_con_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'specialResolution':
+            documents = self.get_special_resolution_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'voluntaryDissolution':
+            documents = self.get_voluntary_dissolution_report(filing_id, business_identifier, filing_date)
+        elif filing_type == 'incorporationApplication':
+            documents = self.get_incorporation_application_report(filing)
```
severinbeauvais, line 63: Any thought on whether each of these gets should also return a link to the receipt output?

legal-api/src/legal_api/services/document_meta.py
ochiu: We can, the ticket mentioned to exclude receipts as they were from a different api so I just followed that.

legal-api/src/legal_api/services/document_meta.py
ochiu: We can, the ticket mentioned to exclude receipts as they were from a different api so I just followed that.

legal-api/src/legal_api/services/document_meta.py
ochiu: These are for capturing the required ?type query param for IA is why I added these.

@lekshmimallika-aot are there any types I missed?

legal-api/src/legal_api/services/document_meta.py
severinbeauvais: Right, OK for now.

legal-api/src/legal_api/services/document_meta.py
severinbeauvais: In future, when we support other incorp entity types, we may different types of certificates or notices. Can we update the code then without affecting the UI? (I don't want to add an unnecessary dependency if we can preempt it now with unique type names.)

ochiu: I think you can use the same type such as ```noa``` and ```certificate``` as it would be based on the filing itself and we can derive a Notice of Article based on the entity type when we return the output.

legal-api/src/legal_api/services/document_meta.py
ochiu: @severinbeauvais so I wanted to use documents as I wanted this document service to potentially handle more than outputs/reports as needed and it populates the documents array in the filing json.

The word report is what is also mainly used in the API, so I am not sure what is best to refactor here.

@lekshmimallika-aot any thoughts?

legal-api/src/legal_api/services/document_meta.py
```diff
@@ -0,0 +1,169 @@
...
+            elif filing_type == 'specialResolution':
+                documents = self.get_special_resolution_report(filing_id, business_identifier, filing_date)
+            elif filing_type == 'voluntaryDissolution':
+                documents = self.get_voluntary_dissolution_report(filing_id, business_identifier, filing_date)
+
+        return documents
+
+    def get_annual_report(self, filing_id, business_identifier, filing_date):
+        """Return an annual report document meta object."""
+        return [self.create_report_object(filing_id, 'Annual Report',
```
lekshmimallika-aot, line 72: Since all these are calling the same method with different arguments, is it worth to have a map kind of structure with the filing type as key and  an array of objects as value. Each of that object will have the file name, query param, is  draft version available as properties so that any future reports will have to keep adding the required data to the array and not touch the framework code? We can do it post mvp also. Just thinking whether that would be a nice thing to do.

legal-api/src/legal_api/services/document_meta.py
lekshmimallika-aot: Looks fine to me

legal-api/src/legal_api/services/document_meta.py
lekshmimallika-aot: I understand that there is phase 2 for this ticket to provide the links. If my above comment makes sense, it can be done at that point.

ochiu: @lekshmimallika-aot I will take a look at this in another PR if that is okay. I agree we can probably do something a bit better here.

PR #839, opened 2020-06-03
Bug 3748

peter-freshworks:
*Issue #:* /bcgov/entity#3748

*Description of changes:*
Fix for file-and-pay failure on numbered companies. Also includes fix for bad print statement in future-effective filings job
-incorp tests will fail unless corresponding pr is merged in schemas

### DIFF
--- a/jobs/future-effective-filings/file_future_effective.py
+++ b/jobs/future-effective-filings/file_future_effective.py
@@ -88,7 +88,9 @@ def get_filings(app: Flask = None):
 
 async def run(loop, application: Flask = None):
     """Run the methods for applying future effective filings."""
-    application = create_app()
+    if application is None:
+        application = create_app()
+
     queue_service = ServiceWorker(
         loop=loop,
         nats_connection_options=default_nats_options,
@@ -134,9 +136,10 @@ async def run(loop, application: Flask = None):
             application.logger.error(err)
 
 if __name__ == '__main__':
+    application = create_app()
     try:
         event_loop = asyncio.get_event_loop()
-        event_loop.run_until_complete(run(event_loop))
+        event_loop.run_until_complete(run(event_loop, application))
     except Exception as err:  # pylint: disable=broad-except; Catching all errors from the frameworks
-        print('problem in running the service: %s', err, stack_info=True, exc_info=True)
+        application.logger.error(err)
 
--- a/jobs/future-effective-filings/file_future_effective.py
+++ b/jobs/future-effective-filings/file_future_effective.py
@@ -141,5 +141,4 @@ if __name__ == '__main__':
         event_loop = asyncio.get_event_loop()
         event_loop.run_until_complete(run(event_loop, application))
     except Exception as err:  # pylint: disable=broad-except; Catching all errors from the frameworks
-        application.logger.error(err)
-
+        sentry_sdk.capture_exception(err)
--- a/jobs/future-effective-filings/file_future_effective.py
+++ b/jobs/future-effective-filings/file_future_effective.py
@@ -141,4 +141,6 @@ if __name__ == '__main__':
         event_loop = asyncio.get_event_loop()
         event_loop.run_until_complete(run(event_loop, application))
     except Exception as err:  # pylint: disable=broad-except; Catching all errors from the frameworks
-        sentry_sdk.capture_exception(err)
+        application.logger.error(err)
+        raise err
+
--- a/legal-api/requirements/bcregistry-libraries.txt
+++ b/legal-api/requirements/bcregistry-libraries.txt
@@ -1,1 +1,1 @@
-git+https://github.com/bcgov/business-schemas.git@2.5.13#egg=registry_schemas
+git+https://github.com/bcgov/business-schemas.git#egg=registry_schemas
--- a/legal-api/requirements/bcregistry-libraries.txt
+++ b/legal-api/requirements/bcregistry-libraries.txt
@@ -1,1 +1,1 @@
-git+https://github.com/bcgov/business-schemas.git#egg=registry_schemas
+git+https://github.com/bcgov/business-schemas.git@2.5.13#egg=registry_schemas
--- a/legal-api/requirements/bcregistry-libraries.txt
+++ b/legal-api/requirements/bcregistry-libraries.txt
@@ -1,1 +1,1 @@
-git+https://github.com/bcgov/business-schemas.git@2.5.12#egg=registry_schemas
+git+https://github.com/bcgov/business-schemas.git#egg=registry_schemas
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -506,7 +506,7 @@ class ListFilingResource(Resource):
             try:
                 business.legal_name = filing.json['filing']['incorporationApplication']['nameRequest']['legalName']
             except KeyError:
-                business.legal_name =  business.identifier
+                business.legal_name = business.identifier
 
         else:
             mailing_address = business.mailing_address.one_or_none()
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -501,9 +501,12 @@ class ListFilingResource(Resource):
         if filing.filing_type == Filing.FILINGS['incorporationApplication'].get('name'):
             mailing_address = Address.create_address(
                 filing.json['filing']['incorporationApplication']['offices']['registeredOffice']['mailingAddress'])
-            corp_type = filing.json['filing']['incorporationApplication']['nameRequest']['legalType']
-            business.legal_name = filing.json['filing']['incorporationApplication']['nameRequest'].get('legalName') \
-                or business.identifier
+            corp_type = filing.json['filing']['business'].get('legalType', 'BC')
+
+            try:
+                business.legal_name = filing.json['filing']['incorporationApplication']['nameRequest']['legalName']
+            except KeyError:
+                business.legal_name =  business.identifier
 
         else:
             mailing_address = business.mailing_address.one_or_none()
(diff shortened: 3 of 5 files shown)


legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -116,7 +116,7 @@ def validate_roles(incorporation_json) -> Error:
         err_path = '/filing/incorporationApplication/parties/roles'
         msg.append({'error': 'Must have a maximum of one completing party', 'path': err_path})
 
-    if incorporation_json['filing']['incorporationApplication']['nameRequest']['legalType'] == 'BC':
```
severinbeauvais, line 119: FWIW, this field was called `requestTypeCd`.

peter-freshworks: really? I haven't seen anything in the db with that key...is the front end passing that to the back end at any point?

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -116,7 +116,7 @@ def validate_roles(incorporation_json) -> Error:
...
+    if incorporation_json['filing']['business'].get('legalType','BC') == 'BC':
```
kialj876, line 119: I think this line is incorrect according to the schema. You won't have a business section in an incorp filing -> legalType will be in the namerequest section

legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -501,9 +501,12 @@ def _create_invoice(business: Business,  # pylint: disable=too-many-locals
         if filing.filing_type == Filing.FILINGS['incorporationApplication'].get('name'):
             mailing_address = Address.create_address(
                 filing.json['filing']['incorporationApplication']['offices']['registeredOffice']['mailingAddress'])
-            corp_type = filing.json['filing']['incorporationApplication']['nameRequest']['legalType']
-            business.legal_name = filing.json['filing']['incorporationApplication']['nameRequest'].get('legalName') \
-                or business.identifier
+            corp_type = filing.json['filing']['business'].get('legalType', 'BC')
```
kialj876, line 504: If the filing.json is an incorp filing then you won't have a business section according to the schema, correct? This should fail the pytests if they're using the example data from the schema repo

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
severinbeauvais: Nope. That's the name of the field in the NR. But I don't see anything wrong with getting the legal type from the business object. (It should have been set there when the IA was first created.)

In this screenshot, I notice a possibly new field, "entity_type_cd". I didn't see this in previous NRs we queried. (Also the type is CR -- corporation -- which is a current limitation of the namex that we query. But the business record should say it's a BC.)

image

legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -501,9 +501,12 @@ def _create_invoice(business: Business,  # pylint: disable=too-many-locals
...
+
+            try:
+                business.legal_name = filing.json['filing']['incorporationApplication']['nameRequest']['legalName']
+            except KeyError:
+                business.legal_name =  business.identifier
```
kialj876, line 509: What's the issue you're trying to fix here? I don't think this will change anything, but it looks like there could be a problem where the 'business.identifier' doesn't exist yet for an FED incorp? When a future effective numbered company is filed there's no legal name and no business yet

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
severinbeauvais: Yes we do.

image

severinbeauvais: But a Numbered Company IA filing does not have an incorporationApplication object.

image

kialj876: Interesting okay well then we should change the logic in the schema: right now it says if not incorp then business is required. Now we're depending on the business to be there in an incorp so it should be required for incorp as well

severinbeauvais: Does it make sense to update the schema in THIS ticket?

legal-api/src/legal_api/resources/business/business_filings.py
kialj876: Okay I just saw your schema PR -> will the business.identifier exist already or does it need to be filing.json['filing']['business']['identifier']?

peter-freshworks: I believe the business Identifier in this case will come from the bootstrap entity (it is a union type way up in the initial caller...from what I recall)

legal-api/src/legal_api/resources/business/business_filings.py
peter-freshworks: maybe we should update the schema...hmmm...but the sample data has been changed and we can be guaranteed of a 'business' node since the introduction of the bootstrap

legal-api/requirements/bcregistry-libraries.txt
```diff
@@ -1 +1 @@
-git+https://github.com/bcgov/business-schemas.git@2.5.12#egg=registry_schemas
+git+https://github.com/bcgov/business-schemas.git#egg=registry_schemas
```
severinbeauvais, line 1: Are you sure that you don't want to control the version of the schema you use?

kialj876: yeah we will want this controlled I think

kialj876: If you want to bring in the schema changes then you can do a gh release in the schema repo for .13 and then point to version .13 here

thorwolpert: Your requirements.txt will need the correct version.

peter-freshworks: oh we don't want the most recent? Ok will update

severinbeauvais: Currently, you probably want the most recent (13).
But we may want the API to use a set version of the schema (eg, 12), for example if the API is not yet ready to handle the latest schema changes. Or maybe for a hotfix.

The UIs do the same thing for the sbc-common-components. It allows the UI to control when it upgrades to the latest library (instead of automatically picking up the latest changes, which may be breaking!).

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -116,7 +116,7 @@ def validate_roles(incorporation_json) -> Error:
         err_path = '/filing/incorporationApplication/parties/roles'
         msg.append({'error': 'Must have a maximum of one completing party', 'path': err_path})
 
-    if incorporation_json['filing']['incorporationApplication']['nameRequest']['legalType'] == 'BC':
+    if incorporation_json['filing']['business'].get('legalType', 'BC') == 'BC':
```
thorwolpert, line 119: Based on Linda's comment, this looks incorrect.
"My understanding is temporarily we are basing this on the button the person selects on the business registry page. Once names request is built, the legal type will be determined through Names Request"

peter-freshworks: should the entire check be omitted then? It sounds like we can't guarantee the legalType is in the incorporation JSON currently

severinbeauvais: @thorwolpert Yes, we can get entity type from the NR (but not currently as it is CR).

Wouldn't `filing.business.entityType` always be set from when the IA is created?

I'm not sure when that "button" will be implemented, but even so (and especially for numbered companies), that code will know what entity type to set, right?

legal-api/requirements/bcregistry-libraries.txt
thorwolpert: The version doesn't need to be here. It does need to be in the requirements.txt file.

jobs/future-effective-filings/file_future_effective.py
```diff
@@ -134,9 +136,10 @@ async def run(loop, application: Flask = None):
...
 
 if __name__ == '__main__':
+    application = create_app()
     try:
         event_loop = asyncio.get_event_loop()
-        event_loop.run_until_complete(run(event_loop))
+        event_loop.run_until_complete(run(event_loop, application))
     except Exception as err:  # pylint: disable=broad-except; Catching all errors from the frameworks
-        print('problem in running the service: %s', err, stack_info=True, exc_info=True)
+        application.logger.error(err)
```
thorwolpert, line 144: If this error requires human intervention to fix, it should get put on to Sentry

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
thorwolpert: According to the req folks, it is set by the names flow.
For now, it could be left out until names gets working for us.
And for MVP we only handle this one legal type anyways, AND BC doesn't let us know it's NOT a Benefit Company. So it doesn't do anything we actually want / need.
+1 on just removing the check @peter-freshworks as you. suggested

jobs/future-effective-filings/file_future_effective.py
peter-freshworks: I thought this did that, this has now been changed...hopefully better now

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
peter-freshworks: check has been removed with FUTURE tag

legal-api/requirements/bcregistry-libraries.txt
peter-freshworks: should be fixed now

peter-freshworks: updated

severinbeauvais: Why does the diff still show that you deleted "@2.5.12" ?

Is this is a generated file, so it doesn't matter?

peter-freshworks: well, if the version doesn't need to be here I figured you probably want the most recent version if rebuilding the requirements locally. I had it @2.5.13 but if the version isn't needed in this file I'll drop it as 2.5.12 would be incorrect

PR #870, opened 2020-06-16
email template tweaks

severinbeauvais:
*Issue #:* /bcgov/entity#4060

*Description of changes:*

These are the template changes to satisfy the subject ticket. There will be additional code changes to populate some variables used by the templates.

- updated company name
- updated incorporation number
- updated business number
- deleted obsolete file (and its reference)


queue_services/entity-emailer/src/entity_emailer/email_processors/__init__.py
```diff
@@ -41,7 +41,6 @@ def substitute_template_parts(template_code):
         'header',
         'initiative-notice',
         'logo',
-        'partners',
```
kialj876, line 44: thank you!

queue_services/entity-emailer/src/entity_emailer/email_templates/common/business-info.html
```diff
@@ -2,8 +2,8 @@
 <table class="business-info-table" cellspacing="10" role="presentation">
   <tr>
     <td class="name">Company Name:</td>
-    {% if incorporationApplication.nameRequest.nrNumber %}
-    <td class="value">{{ incorporationApplication.nameRequest.legalName }}</td>
+    {% if business.legalName %}
```
lekshmimallika-aot, line 5: Is this business-info used only in the email after filer completes?

severinbeauvais: Yes, it is used for:
1. email "we have received your incorporation application"
2. email "you have successfully registered a business with BC Business Registry"

In the first case it's after filing. In the second case it's after completion.

severinbeauvais: In the first case, for a numbered company, there is no Legal Name.

In the second case, for a numbered company, it should be "1234567 B.C. LTD.".

lekshmimallika-aot: Ya , I got that part. Does this work for a named company which is just paid?

severinbeauvais: I'll look into this. Don't merge this PR in the meantime.

severinbeauvais: Looks like the code above won't work for a just-filed named company (since the filing business object doesn't contains legal name). I will fix this.

severinbeauvais: I restored some of the previous code (which worked), but left the check for business.legalName as that should be populated for completed filings (in near-future code).

PR #874, opened 2020-06-16
Report fixes

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4083

*Description of changes:*
Fix to display the filing id in the footer
Updated unit tests


legal-api/src/legal_api/reports/report.py
```diff
@@ -247,6 +247,7 @@ def _format_address(address):
 
     def _format_incorporation_data(self, filing, report_type):
         filing['header']['reportType'] = report_type
+        filing['header']['filingId'] = self._filing.id
```
severinbeauvais, line 250: So filing id is otherwise populated when I save a draft? Will this new code overwrite that?

lekshmimallika-aot: This code will ensure that filing id is always present in the header of the json that gets passed to the report template

PR #875, opened 2020-06-16
strip out incorrect address type coming from front end

peter-freshworks:
*Issue #:* /bcgov/entity#4016

*Description of changes:*
remove the word 'Address' in the address key if it exists. It comes from the front end into the filing JSON for formatting reasons, but it is not a valid address type. This is the simplest way to sanitise it.

### DIFF
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/__init__.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/__init__.py
@@ -34,6 +34,14 @@ def create_address(address_info: Dict, address_type: str) -> Address:
     """Create an address."""
     if not address_info:
         return Address()
+
+    db_address_type = address_type
+
+    if db_address_type == 'mailingAddress':
+        db_address_type == 'mailing'
+    elif db_address_type == 'deliveryAddress':
+        db_address_type = 'delivery'
+
     address = Address(street=address_info.get('streetAddress'),
                       street_additional=address_info.get('streetAddressAdditional'),
                       city=address_info.get('addressCity'),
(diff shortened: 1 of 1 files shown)


queue_services/entity-filer/src/entity_filer/filing_processors/__init__.py
```diff
@@ -41,7 +41,7 @@ def create_address(address_info: Dict, address_type: str) -> Address:
                       country=pycountry.countries.search_fuzzy(address_info.get('addressCountry'))[0].alpha_2,
                       postal_code=address_info.get('postalCode'),
                       delivery_instructions=address_info.get('deliveryInstructions'),
-                      address_type=address_type
+                      address_type=address_type.replace('Address', '')
```
severinbeauvais, line 44: Would it be more clear to have a switch or if statements, kinda like:

if (address_type == "addressMailing") then
  address_type = "mailing"
else if (address_type == "addressDelivery") then
  address_type = "delivery"

severinbeauvais: Also, is case correct?

severinbeauvais: Also, can this be unit tested?

peter-freshworks: 1) yes, I thought about it, but it sort of defeats the purpose of iterating over dictionary keys in the caller...but I think I will change it since you've mentioned it.

2) This is how it comes in from the filing, and the schema should enforce that, but it doesn't hurt to make it case-insensitive

3) It shouldn't need a new test case as we already test IA using this format

severinbeauvais: 1. or add a comment explaining why / what you're doing

severinbeauvais: 2. I thought it was "addressMailing"?

severinbeauvais: 3. ok

peter-freshworks: 1) done, 2) no, it was mailingAddress, 3) good

PR #885, opened 2020-06-24
updated comment

severinbeauvais:
*Issue #:* N/A

*Description of changes:*
After testing this in Dev, I confirm that the time comparison logic is correct.

### DIFF
--- a/jobs/future-effective-filings/file_future_effective.py
+++ b/jobs/future-effective-filings/file_future_effective.py
@@ -125,7 +125,7 @@ async def run(loop, application: Flask = None):
             for filing in filings:
                 filing_id = filing['filing']['header']['filingId']
                 effective_date = filing['filing']['header']['effectiveDate']
-                # TODO Use UTC time?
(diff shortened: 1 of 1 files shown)


jobs/future-effective-filings/file_future_effective.py
```diff
@@ -125,7 +125,7 @@ async def run(loop, application: Flask = None):
             for filing in filings:
                 filing_id = filing['filing']['header']['filingId']
                 effective_date = filing['filing']['header']['effectiveDate']
-                # TODO Use UTC time?
+                # NB: effective_date and now are both UTC
```
thorwolpert, line 128: should this just be deleted?

severinbeauvais: I thought the comment had value but I'm OK with just deleting it.

Do any other reviewers have a strong opinion about this?

ochiu: Either is fine with me :)

PR #886, opened 2020-06-24
fixes suggested by DBAs

peter-freshworks:
*Issue #:* /bcgov/entity#4035

*Description of changes:*

### DIFF
--- a/colin-api/src/colin_api/models/filing.py
+++ b/colin-api/src/colin_api/models/filing.py
@@ -49,7 +49,7 @@ class Filing:
             'OTCON': 'continuedOut'
         },
         # implemented BCOMP filings
-        'BC': {
+        'BEN': {
             'BEINC': 'incorporationApplication',
             'ANNBC': 'annualReport',
             'NOCDR': 'changeOfDirectors',
@@ -59,7 +59,7 @@ class Filing:
 
     USERS = {
         'CP': 'COOPER',
-        'BC': 'BCOMPS'
+        'BEN': 'BCOMPS'
     }
 
     # dicts containing data
@@ -811,9 +811,9 @@ class Filing:
 
             if (legal_type!='CP'):
                # Future: May need a different way of determining legal type
-               legal_type = 'BC'
+               legal_type = 'BEN'
 
-            user_id = Filing.USERS[legal_type] if legal_type in ('CP', 'BC') else None
+            user_id = Filing.USERS[legal_type] if legal_type in ('CP', 'BEN') else None
             cursor = con.cursor()
 
             # create new event record, return event ID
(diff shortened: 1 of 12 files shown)


colin-api/src/colin_api/models/filing.py
```diff
@@ -893,6 +910,7 @@ def add_filing(cls, con, filing):  # pylint: disable=too-many-locals,too-many-st
                 filing_type_cd = 'OTINC'
                 if legal_type == 'BC':
                     filing_type_cd = 'BEINC'
+                    corp_num = corp_num.replace('BC', '00')[-7:]
```
kialj876, line 913: I thought you were going to take out the replace?

colin-api/src/colin_api/models/filing.py
```diff
@@ -796,6 +808,11 @@ def add_filing(cls, con, filing):  # pylint: disable=too-many-locals,too-many-st
         try:
             corp_num = filing.get_corp_num()
             legal_type = corp_num[:2]
+
+            if (legal_type!='CP'):
+               # Future: May need a different way of determining legal type
+               legal_type = 'BC'
```
kialj876, line 814: Shouldn't this be 'BEN'?

kialj876: When I was testing this with David I had to update this in a few places

peter-freshworks: D'oh

peter-freshworks:
> Hey peter - looks pretty good to me now except incorporation is failing for me in postman. Looks like the get next corp num return type is different (str where it used to be a list I think?) which will cause the filer to fail so it needs to be fixed, but even after fixing that locally the incorp still fails for me. Soemthing to do with find_by_identifier is giving an error. Do you get the same thing?

should work now

queue_services/entity-filer/src/entity_filer/filing_processors/incorporation_filing.py
```diff
@@ -36,7 +36,7 @@ def get_next_corp_num(business_type: str):
         return None
 
     if resp.status_code == 200:
-        new_corpnum = resp.json()['corpNum'][0]
+        new_corpnum = resp.json()['corpNum']
```
kialj876, line 39: did you test that this works properly now that it's a string instead of an int?

kialj876:
> I still get an error on incorporation:
> 2020-06-25 09:06:35,090 - colin_api - ERROR in business:business.py:230 - find_by_identifier:
> 2020-06-25 09:06:35,091 - colin_api - ERROR in filing:filing.py:130 - post:
> 2020-06-25 09:06:35,116 - colin_api - ERROR in filing:filing.py:137 - post:
> 2020-06-25 09:06:35,140 - werkzeug - INFO in _internal:_internal.py:122 - _log: 127.0.0.1 - - [25/Jun/2020 09:06:35] "POST /api/v1/businesses/BC0870705/filings/incorporationApplication HTTP/1.1" 500 -

same one as yesterday using the postman inorp test over oracle-dev

queue_services/entity-filer/src/entity_filer/filing_processors/incorporation_filing.py
kialj876: this breaks when I run the test for it

PR #892, opened 2020-06-26
handle future effective dating NR

ochiu:
*Issue #:* /bcgov/entity#3883

*Description of changes:*
- added functions for updating a name request in namex service
- added logic to update to name request expiration date on a completed filing, if the effective date exceeds the name request expiration date, the expiration date is extended to the future effective date (plus one day to allow for time jobs to process)
- added an update nr test

### DIFF
--- a/legal-api/src/legal_api/services/namex.py
+++ b/legal-api/src/legal_api/services/namex.py
@@ -16,12 +16,16 @@
 from datetime import datetime
 from enum import Enum
 
+import datedelta
+import pytz
 import requests
 from flask import current_app
 
 
 class NameXService():
     """Provides services to use the namex-api."""
+
+    DATE_FORMAT = '%a, %d %b %Y %H:%M:%S %Z'
 
     class State(Enum):
         """Name request states."""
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/services/namex.py
```diff
@@ -64,6 +68,46 @@ def query_nr_number(identifier: str):
 
         return nr_response
 
+    @staticmethod
+    def update_nr(nr_json):
```
severinbeauvais, line 72: Does this method have to be declared as async?

ochiu: I don't believe so, I ran some tests against this for various filings and it seems to work fine. I also haven't really used async for python so I am not familiar what requirements in this case should warrant its usage.

severinbeauvais: Per discussion with Thor, this method simply blocks on its POST call (and runs in its own thread), so it doesn't need to be async (and it is not awaited upon anyway)... In short, this code is fine. I'm resolving this conversation.

PR #900, opened 2020-06-30
BN trigger changes

lekshmimallika-aot:
*Issue #:* /bcgov/entity#N/A

*Description of changes:*

### DIFF
--- a/jobs/update-legal-filings/update_legal_filings.py
+++ b/jobs/update-legal-filings/update_legal_filings.py
@@ -24,6 +24,7 @@ import sentry_sdk  # noqa: I001; pylint: disable=ungrouped-imports; conflicts wi
 from colin_api.models.filing import Filing
 from entity_queue_common.service import QueueServiceManager
 from legal_api.services.bootstrap import AccountService
+from legal_api.services.queue import QueueService
 from sentry_sdk import capture_message
 from sentry_sdk.integrations.logging import LoggingIntegration  # noqa: I001
 from flask import Flask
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/services/queue.py
```diff
@@ -120,17 +120,25 @@ async def close(self):
...
         """Publish the json payload to the Queue Service."""
         try:
-            self.loop.run_until_complete(self.async_publish_json(payload))
+            self.loop.run_until_complete(self.async_publish_json(payload, self.subject))
         except Exception as err:
             self.logger.error('Error: %s', err)
             raise err
 
-    async def async_publish_json(self, payload=None):
+    async def publish_json_to_subject(self, payload=None, subject=None):
```
lekshmimallika-aot, line 128: Added a new method here because the self.subject is the filer one and it is better not to disturb that. Also the existing publish_json calls run_until_complete which throws an error  in this case as the event loop is already running from the update_legal_filings.py. We just have to use an async/await here

kialj876: Nicely done!

jobs/update-legal-filings/update_legal_filings.py
```diff
@@ -24,6 +24,7 @@
 from colin_api.models.filing import Filing
 from entity_queue_common.service import QueueServiceManager
```
kialj876, line 25: Can we remove the common queueServiceManger and entity_queue_common from the requirements/repos and requirements.txt? I don't think it's used now that you've switched to the legal_api's queue service

lekshmimallika-aot: Sure. Will do

PR #904, opened 2020-07-09
added Name Translation to IA/NOA outputs

severinbeauvais:
*Issue #:* /bcgov/entity#4104

*Description of changes:*
- added name translation section to IA report template
- ~~computed list of translations for alteration filings~~ _future ticket_
- set list of translations for IA filings

### DIFF
--- a/legal-api/report-templates/incorporationApplication.html
+++ b/legal-api/report-templates/incorporationApplication.html
@@ -90,6 +90,13 @@
                   <span class="section-sub-title">{{incorporationApplication.nameRequest.legalName}}</span>
                   {% endif %}
                </p>
+               {% if listOfTranslations|length > 0 %}
+               <p>
+                  <div class="title">Translation(s) of Company Name:</div>
+                  {% for translation in listOfTranslations %}
+                  <div>{{translation}}</div>
+                  {% endfor %}
+               </p>
             </div>
          </div>
          <div class="separator"></div>
--- a/legal-api/report-templates/incorporationApplication.html
+++ b/legal-api/report-templates/incorporationApplication.html
@@ -97,6 +97,7 @@
                   <div>{{translation}}</div>
                   {% endfor %}
                </p>
+               {% endif %}
             </div>
          </div>
          <div class="separator"></div>
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -170,13 +170,6 @@ class Report:  # pylint: disable=too-few-public-methods
             except KeyError:
                 pass
 
-            # set translation list for COA/COD/Alteration filing
-            try:
-                # FUTURE: get current name translations from database
-                filing['listOfTranslations'] = filing['business'].get('nameTranslations', [])
-            except KeyError:
-                pass
-
             self._set_dates(filing)
 
         self._set_meta_info(filing)
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -170,9 +170,10 @@ class Report:  # pylint: disable=too-few-public-methods
             except KeyError:
                 pass
 
-            # set translation list from the alteration filing
+            # set translation list for COA/COD/Alteration filing
             try:
-                self._set_translations(filing)
+                # FUTURE: get current name translations from database
+                filing['listOfTranslations'] = filing['business'].get('nameTranslations', [])
             except KeyError:
                 pass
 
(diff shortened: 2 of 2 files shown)


legal-api/src/legal_api/reports/report.py
```diff
@@ -216,6 +223,24 @@ def _format_directors(self, directors):
                 pass
         return directors
 
+    def _set_translations(self, filing):
+        # initial (existing) translations, if any
+        nameTranslations = filing['business'].get('nameTranslations', [])
```
severinbeauvais, line 228: @lekshmimallika-aot I'm assuming there will be a list of existing name translations in the business object. Is this actually the plan? If not, where do I get the initial list of name translations for an alteration?

severinbeauvais:
@lekshmimallika-aot And... how do I test all this? Thanks in advance! :)

legal-api/src/legal_api/reports/report.py
```diff
@@ -216,6 +223,24 @@ def _format_directors(self, directors):
                 pass
         return directors
 
+    def _set_translations(self, filing):
+        # initial (existing) translations, if any
+        name_translations = filing['business'].get('nameTranslations', [])
```
severinbeauvais, line 228: @lekshmimallika-aot I'm assuming there will be a list of existing name translations in the business object. Is this actually the plan? If not, where do I get the initial list of name translations for an alteration?

lekshmimallika-aot: For which report is this? NOA ?

legal-api/src/legal_api/reports/report.py
```diff
@@ -262,6 +287,8 @@ def _format_incorporation_data(self, filing, report_type):
         self._format_address(filing['incorporationApplication']['offices']['recordsOffice']['deliveryAddress'])
         self._format_address(filing['incorporationApplication']['offices']['recordsOffice']['mailingAddress'])
         self._format_directors(filing['incorporationApplication']['parties'])
+        # create helper list for translations
+        filing['listOfTranslations'] = filing['incorporationApplication']['nameTranslations']
```
lekshmimallika-aot, line 291: Is this method under _set_addresses?

legal-api/src/legal_api/reports/report.py
severinbeauvais: Technically, this is for "not IA" filings -- so yes, for Notice of Alteration.

lekshmimallika-aot: NOA is available only after filing is complete. So you can directly get it from the business object and set it to the json. Its not just for name translations. director addresses etc wont be captured in the schema for alteration yet they have to be present in the output. The same logic applies here

severinbeauvais: That makes sense, thanks.

Do you know if the business object is complete and up to date, or should it be queried from the db? For example, if this is a COA/COD filing, will its business object contain the current name translations?

I figure this code will be used for NOA report for an alteration filing. Does this make sense?

severinbeauvais: > Do you know if the business object is complete and up to date, or should it be queried from the db? For example, if this is a COA/COD filing, will its business object contain the current name translations?

This part will need to be updated - see /bcgov/entity#4299.

severinbeauvais:
I think this is as far as I can go with this ticket. The remaining work to populate the name translations will be done in /bcgov/entity#4299.

legal-api/src/legal_api/reports/report.py
thorwolpert: > Technically, this is for "not IA" filings -- so yes, for Notice of Alteration report.

seems like you guys are talking about 2 different things here. A NOA (Notice of Articles) is available after the filing is complete. A Notice of Alteration is a Filing and is available upon submission.

severinbeauvais: Oops, typo. I meant Notice of Articles... fixed in comment above. Thanks for catching that.

I think we are trying to call the Alteration filing just that -- not "Notice of..." -- to reduce confusion :)

legal-api/src/legal_api/reports/report.py
```diff
@@ -169,6 +169,14 @@ def _get_template_data(self, report_type=None):  # pylint: disable=too-many-bran
                 self._set_directors(filing)
             except KeyError:
                 pass
+
+            # set translation list for COA/COD/Alteration filing
+            try:
+                # FUTURE: get current name translations from database
+                filing['listOfTranslations'] = filing['business'].get('nameTranslations', [])
```
lekshmimallika-aot, line 176: Could you please remove this block? Only the changes for IA needs to be present in this PR

severinbeauvais: Is this not needed for (near-future) Notice Of Articles outputs for (BCOMP) COA, COD and Alteration filings?

This other ticket specifies a NOA for these other filings: /bcgov/entity#4107

severinbeauvais: I will remove this code as requested, as it looks this code will be reworked as part of 4299 (ie, to fetch additional output data that is not in the filing JSON).

legal-api/report-templates/incorporationApplication.html
```diff
@@ -90,6 +90,13 @@
                   <span class="section-sub-title">{{incorporationApplication.nameRequest.legalName}}</span>
                   {% endif %}
                </p>
+               {% if listOfTranslations|length > 0 %}
+               <p>
+                  <div class="title">Translation(s) of Company Name:</div>
+                  {% for translation in listOfTranslations %}
+                  <div>{{translation}}</div>
+                  {% endfor %}
+               </p>
```
lekshmimallika-aot, line 99: Is an endif missing here?

legal-api/src/legal_api/reports/report.py
lekshmimallika-aot: Notice of Article is going to be a totally different approach. For incorporation flow, the IA and NOA make use of the same template with some if conditions

legal-api/report-templates/incorporationApplication.html
severinbeauvais: Fixed.

PR #905, opened 2020-07-09
Name Translation model changes

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4129

*Description of changes:*
Model changes for Name Translations


legal-api/src/legal_api/models/alias.py
```diff
@@ -0,0 +1,70 @@
...
+    """This class manages the aliases."""
+
+    class AliasType(Enum):
+        """Render an Enum of the types of aliases."""
+
+        TRANSLATION = 'TRANSLATION'
+        DBA = 'DBA'
+
+    __tablename__ = 'aliases'
+
```
thorwolpert, line 32: It needs to be versioned
 __versioned__ = {}

lekshmimallika-aot: Updated

PR #906, opened 2020-07-09
AR/COA/COD email templates

severinbeauvais:
*Issues #:* /bcgov/entity#4255 /bcgov/entity#4222 /bcgov/entity#4229

*Description of changes:*
- updated template names
- created AR email template
- created COA email templates (#4222)
- created COD email templates (#4229)
- renamed some templates


kialj876:
> does the AR need a completed template too?
> 
> Looks good - I have no way of testing it for visual representation, but I like the way you named things. It would be good to change 'registered' to completed for IA's as well, but since it's a different processor than the other ones it doesn't really matter

It doesn't say so in your ticket, but it was probably missed

severinbeauvais:
> > does the AR need a completed template too?
> > Looks good - I have no way of testing it for visual representation, but I like the way you named things. It would be good to change 'registered' to completed for IA's as well, but since it's a different processor than the other ones it doesn't really matter
> 
> It doesn't say so in your ticket, but it was probably missed

No, there is no NOA for an AR. I discussed this with Sienna and we updated the ticket and requirements doc.

PR #908, opened 2020-07-10
updated filing documents lists

severinbeauvais:
*Issue #:* /bcgov/entity#4107

*Description of changes:*
- added functions to process correction and alteration filings (for future)
- added COA NOA conditionally
- added COD NOA conditionally
- added CON NOA conditionally (assume same as COA/COD)
- simplified formal parameters / added helper functions / cleanup

### DIFF
--- a/legal-api/src/legal_api/services/document_meta.py
+++ b/legal-api/src/legal_api/services/document_meta.py
@@ -13,10 +13,9 @@
 # limitations under the License.
 
 """This provides the service for filings documents meta data."""
-
 from enum import Enum
 
-from legal_api.models import Filing
+from legal_api.models import Business, Filing
 from legal_api.utils.legislation_datetime import LegislationDatetime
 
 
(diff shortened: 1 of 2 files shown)


legal-api/tests/unit/services/test_document_meta.py
```diff
@@ -16,13 +16,61 @@
...
 """
-
+from legal_api.models import Business
 from legal_api.services import document_meta
-from legal_api.utils.legislation_datetime import LegislationDatetime
 
 
-def test_filing_status_documents_empty(app):
-    """Assert that no documents are returned for non PAID and COMPLETED filing."""
+def factory_business(identifier: str, legal_type: str = ''):
```
lekshmimallika-aot, line 23: These methods are already defined in the model tests. Please reuse them

legal-api/tests/unit/services/test_document_meta.py
```diff
@@ -16,13 +16,61 @@
...
+def test_business_factory(session):
+    """Assert that factory_business works correctly."""
+    business = factory_business('BC1234567', 'BC')
+    assert business.id is not None
+    assert business.identifier == 'BC1234567'
+    assert business.legal_type == 'BC'
+
+
+def test_find_by_identifier(session):
+    """Assert that find_by_identifier works correctly."""
```
lekshmimallika-aot, line 39: Not sure why this test is part of document meta service. Similar tests are already there for business model

legal-api/src/legal_api/services/document_meta.py
```diff
@@ -34,136 +33,236 @@ class ReportType(Enum):
...
+        return reports
+
+    def get_correction_reports(self):  # pylint: disable=no-self-use
+        """Return correction meta object(s)."""
+        # FUTURE: return applicable documents
+        return []
+
+    def get_alteration_reports(self):  # pylint: disable=no-self-use
+        """Return alteration meta object(s)."""
+        # FUTURE: return applicable documents
```
lekshmimallika-aot, line 195: I think this should return NOA

severinbeauvais: Perhaps, but it's future and not in scope for this ticket.

legal-api/src/legal_api/services/document_meta.py
```diff
@@ -34,136 +33,236 @@ class ReportType(Enum):
...
-            elif filing_type == 'voluntaryDissolution':
-                documents = self.get_voluntary_dissolution_report(filing_id, business_identifier, filing_date)
+            documents = self.get_incorporation_application_reports(filing)
+        elif filing_type == 'annualReport':
+            documents = self.get_ar_reports()
+        elif filing_type == 'changeOfAddress':
+            documents = self.get_coa_reports()
+        elif filing_type == 'changeOfDirectors':
+            documents = self.get_cod_reports()
+        elif filing_type == 'changeOfName':
```
lekshmimallika-aot, line 72: I had mentioned this in Ody's inital PR as well. These methods are basically doing the same thing with different values. I think it would be better to externalize the reports available for each filing type and status into a json so that a single method can replace all this. It would be great to have a tech debt ticket to take care of this.

severinbeauvais: Good idea. See /bcgov/entity#4335.

legal-api/tests/unit/services/test_document_meta.py
severinbeauvais: Fixed in next commit.

legal-api/tests/unit/services/test_document_meta.py
severinbeauvais: This was very helpful for me to build up the unit tests and verify that each step worked.
But you are right... fixed in next commit.

severinbeauvais:
> Please look into the code smells in the Sonar check

Fixed in new commit.

PR #921, opened 2020-07-17
BC Change Of Address PDF Output Template

cameron-freshworks:
*Issue #:* /bcgov/entity#4226

*Description of changes:*

* Contains just the template part of the Change of Address PDF output.

* Work left to be done within reports.py to populate and invoke the correct template.


lekshmimallika-aot:
The template needs is not rendering the data correctly. I will be creating a follow up PR  after fixing the issues.

PR #931, opened 2020-07-23
4166 data loader bc comp

kialj876:
*Issue #:* /bcgov/entity#4166

*Description of changes:*
- updated data loader for loading in bc companies
- refactored to be more usable, added linting etc.
- updated colin-api for some issues when getting bc companies
- updated insert/get on share structure class values (they were incorrect before - waiting for clarification from David and or Linda before merge)


lear-db/test_data/requirements/bcregistry-libraries.txt
```diff
@@ -0,0 +1,3 @@
+git+https://github.com/bcgov/business-schemas.git@2.5.11#egg=registry_schemas
```
thorwolpert, line 1: Should this move to the current version??

kialj876: Oh nice catch - I will update this to always get the current one

colin-api/src/colin_api/models/corp_party.py
```diff
@@ -22,7 +22,7 @@
 from flask import current_app
 
 from colin_api.exceptions import PartiesNotFoundException
-from colin_api.models import Address  # pylint: disable=cyclic-import
+from colin_api.models import Address, Business
 from colin_api.resources.db import DB
```
thorwolpert, line 26: move the DB before the models, as they import it.

kialj876: thanks, was wondering what to do here

kialj876: hmm it still doesn't like it so I'm going to put it back and add the pylint disable for now (it was there before - I don't remember why, but I took it off in my changes)

kialj876: alright changed it again haha - this was a tricky one because it wasn't coming up locally

PR #940, opened 2020-07-30
fixed COD effective date logic

severinbeauvais:
*Issue #:* /bcgov/entity#4311

*Description of changes:*
- removed unused code (effective date is always present, so no need to check filing date)
- compared dates instead of datetimes
- expanded unit tests

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/change_of_directors.py
+++ b/legal-api/src/legal_api/services/filings/validations/change_of_directors.py
@@ -21,6 +21,7 @@ from flask_babel import _ as babel  # noqa: N813, I004, I001; importing camelcas
 from legal_api.errors import Error
 from legal_api.models import Address, Business, Filing
 from legal_api.utils.datetime import datetime
+from legal_api.utils.legislation_datetime import LegislationDatetime
 
 from ...utils import get_str
 # noqa: I003; needed as the linter gets confused from the babel override above.
@@ -79,31 +80,38 @@ def validate_effective_date(business: Business, cod: Dict) -> List:
             (standalone or AR). If this is the case:
         - COD filing that was filed most recently is the most current director information.
     """
-    try:
-        # NB: COD effective date is at 0 hours UTC.
-        filing_effective_date = cod['filing']['header']['effectiveDate']
-    except KeyError:
-        return {'error': babel('No effective_date provided.')}
-
-    try:
-        effective_date = datetime.fromisoformat(filing_effective_date)
-    except ValueError:
-        return {'error': babel('Invalid ISO format for effective_date.')}
-
     msg = []
 
-    # The effective datetime cannot be in the future.
-    if effective_date > datetime.utcnow():
+    # get effective datetime string from filing
+    try:
+        effective_datetime_str = cod['filing']['header']['effectiveDate']
+    except KeyError:
+        return {'error': babel('No effective date provided.')}
+
+    # convert string to datetime
+    try:
+        effective_datetime_utc = datetime.fromisoformat(effective_datetime_str)
+    except ValueError:
+        return {'error': babel('Invalid ISO format for effective date.')}
+
+    # check if effective datetime is in the future
+    if effective_datetime_utc > datetime.utcnow():
         msg.append({'error': babel('Filing cannot have a future effective date.')})
 
-    # The effective date cannot be before their Incorporation Date.
-    if effective_date.date() < business.founding_date.date():
-        msg.append({'error': babel('Filing cannot be before businesses founding date.')})
+    # convert to legislation timezone and then get date only
+    effective_date_leg = LegislationDatetime.as_legislation_timezone(effective_datetime_utc).date()
 
-    # The effective date cannot be before the most recent COD or AR.
-    last_cod_filing = Filing.get_most_recent_legal_filing(business.id, Filing.FILINGS['changeOfDirectors']['name'])
+    # check if effective date is before their incorporation date
+    founding_date_leg = LegislationDatetime.as_legislation_timezone(business.founding_date).date()
+    if effective_date_leg < founding_date_leg:
+        msg.append({'error': babel('Effective date cannot be before businesses founding date.')})
+
+    # check if effective date is before their most recent COD or AR date
+    last_cod_filing = Filing.get_most_recent_legal_filing(business.id,
+                                                          Filing.FILINGS['changeOfDirectors']['name'])
     if last_cod_filing:
-        if effective_date.date() < last_cod_filing.effective_date.date():
-            msg.append({'error': babel('Filing cannot be before another Change of Director filing.')})
+        last_cod_date_leg = LegislationDatetime.as_legislation_timezone(last_cod_filing.effective_date).date()
+        if effective_date_leg < last_cod_date_leg:
+            msg.append({'error': babel('Effective date cannot be before another Change of Director filing.')})
 
     return msg
--- a/legal-api/src/legal_api/services/filings/validations/change_of_directors.py
+++ b/legal-api/src/legal_api/services/filings/validations/change_of_directors.py
@@ -104,6 +104,6 @@ def validate_effective_date(business: Business, cod: Dict) -> List:
     last_cod_filing = Filing.get_most_recent_legal_filing(business.id, Filing.FILINGS['changeOfDirectors']['name'])
     if last_cod_filing:
         if effective_date.date() < last_cod_filing.effective_date.date():
-            msg.append({'error': babel("Filing cannot be before another Change of Director filing.")})
+            msg.append({'error': babel('Filing cannot be before another Change of Director filing.')})
 
     return msg
--- a/legal-api/src/legal_api/services/filings/validations/change_of_directors.py
+++ b/legal-api/src/legal_api/services/filings/validations/change_of_directors.py
@@ -70,43 +70,40 @@ def validate_directors_addresses(cod: Dict) -> List:
 def validate_effective_date(business: Business, cod: Dict) -> List:
     """Return an error or warning message based on the effective date validation rules.
 
-    Rules:
+    Rules: (text from the BA rules document)
         - The effective date of change cannot be in the future.
-        - The effective date cannot be a date prior to their Incorporation Date
+        - The effective date cannot be a date prior to their Incorporation Date.
         - The effective date of change cannot be a date that is farther in the past
-            as a previous COD filing (Standalone or AR).
+            than a previous COD filing (standalone or AR).
         - The effective date can be the same effective date as another COD filing
-            (standalone OR AR). If this is the case:
-        - COD filing that was filed most recently as the most current director information.
+            (standalone or AR). If this is the case:
+        - COD filing that was filed most recently is the most current director information.
     """
     try:
+        # NB: COD effective date is at 0 hours UTC.
         filing_effective_date = cod['filing']['header']['effectiveDate']
     except KeyError:
-        try:
-            # we'll assume the filing is at 0 hours UTC
-            filing_effective_date = cod['filing']['header']['date'] + 'T00:00:00+00:00'
-        except KeyError:
-            return {'error': babel('No effective_date or filing date provided.')}
+        return {'error': babel('No effective_date provided.')}
 
     try:
         effective_date = datetime.fromisoformat(filing_effective_date)
     except ValueError:
-        return {'error': babel('Invalid ISO format for effective_date or filing date.')}
+        return {'error': babel('Invalid ISO format for effective_date.')}
 
     msg = []
 
-    # The effective date of change cannot be in the future
+    # The effective datetime cannot be in the future.
     if effective_date > datetime.utcnow():
         msg.append({'error': babel('Filing cannot have a future effective date.')})
 
-    # The effective date cannot be a date prior to their Incorporation Date
-    if effective_date < business.founding_date:
-        msg.append({'error': babel('Filing cannot be before a businesses founding date.')})
+    # The effective date cannot be before their Incorporation Date.
+    if effective_date.date() < business.founding_date.date():
+        msg.append({'error': babel('Filing cannot be before businesses founding date.')})
 
-    last_cod_filing = Filing.get_most_recent_legal_filing(business.id,
-                                                          Filing.FILINGS['changeOfDirectors']['name'])
+    # The effective date cannot be before the most recent COD or AR.
+    last_cod_filing = Filing.get_most_recent_legal_filing(business.id, Filing.FILINGS['changeOfDirectors']['name'])
     if last_cod_filing:
-        if effective_date < last_cod_filing.effective_date:
-            msg.append({'error': babel("Filing's effective date cannot be before another Change of Director filing.")})
+        if effective_date.date() < last_cod_filing.effective_date.date():
+            msg.append({'error': babel("Filing cannot be before another Change of Director filing.")})
 
     return msg
--- a/legal-api/src/legal_api/utils/legislation_datetime.py
+++ b/legal-api/src/legal_api/utils/legislation_datetime.py
@@ -23,12 +23,12 @@ class LegislationDatetime():
     """Date utility using legislation timezone for reporting and future effective dates."""
 
     @staticmethod
-    def now():
+    def now() -> datetime:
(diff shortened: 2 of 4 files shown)


legal-api/src/legal_api/services/filings/validations/change_of_directors.py
```diff
@@ -70,43 +70,40 @@ def validate_directors_addresses(cod: Dict) -> List:
...
-    # The effective date of change cannot be in the future
+    # The effective datetime cannot be in the future.
     if effective_date > datetime.utcnow():
         msg.append({'error': babel('Filing cannot have a future effective date.')})
 
-    # The effective date cannot be a date prior to their Incorporation Date
-    if effective_date < business.founding_date:
-        msg.append({'error': babel('Filing cannot be before a businesses founding date.')})
+    # The effective date cannot be before their Incorporation Date.
+    if effective_date.date() < business.founding_date.date():
```
thorwolpert, line 100: this would not allow this filing on the same date as incorporation, which it can be. It just has to be later then the date + time it was incoporated.

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
```diff
@@ -70,43 +70,40 @@ def validate_directors_addresses(cod: Dict) -> List:
...
 
-    last_cod_filing = Filing.get_most_recent_legal_filing(business.id,
-                                                          Filing.FILINGS['changeOfDirectors']['name'])
+    # The effective date cannot be before the most recent COD or AR.
+    last_cod_filing = Filing.get_most_recent_legal_filing(business.id, Filing.FILINGS['changeOfDirectors']['name'])
     if last_cod_filing:
-        if effective_date < last_cod_filing.effective_date:
-            msg.append({'error': babel("Filing's effective date cannot be before another Change of Director filing.")})
+        if effective_date.date() < last_cod_filing.effective_date.date():
+            msg.append({'error': babel('Filing cannot be before another Change of Director filing.')})
```
thorwolpert, line 107: Why make this less specific for the API user?

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
```diff
@@ -70,43 +70,40 @@ def validate_directors_addresses(cod: Dict) -> List:
...
-    if effective_date < business.founding_date:
-        msg.append({'error': babel('Filing cannot be before a businesses founding date.')})
+    # The effective date cannot be before their Incorporation Date.
+    if effective_date.date() < business.founding_date.date():
+        msg.append({'error': babel('Filing cannot be before businesses founding date.')})
 
-    last_cod_filing = Filing.get_most_recent_legal_filing(business.id,
-                                                          Filing.FILINGS['changeOfDirectors']['name'])
+    # The effective date cannot be before the most recent COD or AR.
+    last_cod_filing = Filing.get_most_recent_legal_filing(business.id, Filing.FILINGS['changeOfDirectors']['name'])
```
thorwolpert, line 104: in general we don't change formatting, for just the sake of formatting without any changes.

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
```diff
@@ -35,11 +26,14 @@
 
 def common_setup(identifier: str, now: datetime):
     """Set up the data for COD tests."""
+    founding_date = now - datedelta.YEAR
     filing_date = now - datedelta.DAY
 
+    # create business founded "a year ago"
     business = Business(identifier=identifier,
-                        founding_date=now - datedelta.YEAR)
+                        founding_date=founding_date)
```
thorwolpert, line 34: is this the only place founding_date gets used, why the formatting change?

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
```diff
@@ -35,11 +26,14 @@
 
 def common_setup(identifier: str, now: datetime):
     """Set up the data for COD tests."""
+    founding_date = now - datedelta.YEAR
     filing_date = now - datedelta.DAY
 
+    # create business founded "a year ago"
```
thorwolpert, line 32: comment was captured in previous code??

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
```diff
@@ -56,130 +50,153 @@ def test_effective_date_sanity_check(session):
...
-    Rules:
-        - The effective date cannot be a date prior to their Incorporation Date
-    """
+def test_validate_effective_date_not_before_founding(session):
+    """Assert that the effective date cannot be a date prior to their Incorporation Date."""
     # setup
     identifier = 'CP1234567'
-    now = datetime(2001, 8, 5, 0, 0, 0, 0, tzinfo=timezone.utc)
-
+    # assign 'now' with non-zero hour so Founding Date is mid-day
```
thorwolpert, line 96: this is not mid-day for the legal timezone.

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
severinbeauvais: This code rejects the filing if it is before the incorp date, so it allows this filing on the same day as the incorp date. Or am I confused?
Also, there is no effective time -- just a date with time = 00:00:00.000. So we can only check dates,

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
severinbeauvais: I was trying to be consistent with the other error messages above, but I can restore this if you like.

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
severinbeauvais: There was no comment in the previous code...?

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
severinbeauvais: I was being explicit (incl comment) about exactly what this variable means. (Eg, "now" isn't really now; it's a relative time within this function only. A bit confusing? Yes.)

I know it's valid syntax to put an expression as a function parameter, but I find that sometimes hides the meaning of the parameter. IMO this code is more clear.

I can revert if you feel strongly about this.

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
severinbeauvais: You're right. It's just not "0 h". I can reword.

severinbeauvais: I may have confused myself about what "time" things happen. All date comparisons should be in local time right?
I will review these and then re-submit for PR.

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
severinbeauvais: Reverted in next commit.

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
severinbeauvais: Message reverted in next commit.

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
thorwolpert: doesn't need a comment
founding_date=now - datedelta.YEAR
but now I have to read the comment or scroll up to look at the data to know that ... and if the variable assignment changes, then so does this (not close in proximity) comment. so less cognitively clean in my mind.
again, sort of a format style change without any real effects.

severinbeauvais: Comment reverted in next commit.

severinbeauvais:
Ready for re-review.

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
severinbeauvais: Updated in next commit.

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
thorwolpert: founding date is a datetime, for new IAs it has the time and other filings can be on the same day, but later than incorp time.

severinbeauvais: You're right, but:
- this is now old code
- there is no time for a COD filing -- just a date -- so we can't check whether the filing is before the incorp time; we can only check date

severinbeauvais:
Note: Thor says it's OK to ignore the SonarCloud code smell (which is due to duplicate date format strings -- but we don't want to update those).

legal-api/src/legal_api/utils/legislation_datetime.py
```diff
@@ -23,12 +23,12 @@ class LegislationDatetime():
     """Date utility using legislation timezone for reporting and future effective dates."""
 
     @staticmethod
-    def now():
+    def now() -> datetime:
```
thorwolpert, line 26: Hadn't noticed that before.
provides the correct **now()**  as **utcnow** so this shouldn't be used.
I can clean this later if needed.

legal-api/src/legal_api/utils/legislation_datetime.py
```diff
@@ -37,12 +37,17 @@ def tomorrow_midnight():
...
         return date_time.astimezone(pytz.timezone(current_app.config.get('LEGISLATIVE_TIMEZONE')))
 
     @staticmethod
-    def format_as_legislation_date(date_string: str):
+    def as_utc_timezone(date_time: datetime) -> datetime:
+        """Return a datetime adjusted to the GMT timezone (aka UTC)."""
+        return date_time.astimezone(pytz.timezone('GMT'))
+
+    @staticmethod
+    def format_as_legislation_date(date_string: str) -> str:
```
thorwolpert, line 50: doesn't this return a -> datetime ?

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
```diff
@@ -56,130 +50,170 @@ def test_effective_date_sanity_check(session):
...
-    """
+    """Assert that the effective date of change cannot be before a previous COD filing."""
     # setup
     identifier = 'CP1234567'
-    now = datetime(2001, 8, 5, 0, 0, 0, 0, tzinfo=timezone.utc)
-    filing_cod = copy.deepcopy(FILING_HEADER)
-    filing_cod['filing']['header']['name'] = 'changeOfDirectors'
-    filing_cod['filing']['changeOfDirectors'] = copy.deepcopy(CHANGE_OF_DIRECTORS)
-
+    founding_date = datetime(2000, 8, 5, 12, 0, 0, 0, tzinfo=timezone.utc)
```
thorwolpert, line 129: Why all the setup change? Styling preference?

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
```diff
@@ -70,43 +71,47 @@ def validate_directors_addresses(cod: Dict) -> List:
...
 
-    # The effective date cannot be a date prior to their Incorporation Date
-    if effective_date < business.founding_date:
-        msg.append({'error': babel('Filing cannot be before a businesses founding date.')})
+    # convert to legislation timezone and then get date only
+    effective_date_leg = LegislationDatetime.as_legislation_timezone(effective_datetime_utc).date()
+
+    # check if effective date is before their incorporation date
+    founding_date_leg = LegislationDatetime.as_legislation_timezone(business.founding_date).date()
+    if effective_date_leg < founding_date_leg:
```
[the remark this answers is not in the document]
severinbeauvais: I don't even know how we'd deal with this... the COD doesn't have a time, so how would you mark that it came after an IA? Or after another COD on the same day? Probably db order (index)?
What do you suggest to do here?

legal-api/src/legal_api/utils/legislation_datetime.py
severinbeauvais: Yes, shouldn't be used, but I just cleaned up the type. Please clean up later as needed. Thanks!

legal-api/src/legal_api/utils/legislation_datetime.py
severinbeauvais: No, the chained return statement below returns a formatted string (strftime).

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
severinbeauvais: Not at all. I am being more explicit (clearer) about initial conditions and what each test verifies. I also renamed some variables to reduce confusion (eg, "now").
(I figure code that's easy to understand/review has a better chance of being correct ;) )

thorwolpert: I guess that's a matter of opinion. more compact is clearer to lots of folks too.
As before, generally our accepted rule is everythings open in refactor, reformatting for preference is something we're avoiding.
It's also MUCH harder to see what you actually changed wrt to business or functionality, as opposed to just moved around.

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
thorwolpert: the submission is a datetime

legal-api/tests/unit/services/filings/validations/change_of_director/test_validation_effective_date.py
thorwolpert: they can all be local, utc, or mixed (harder for humans to reason). Date comparisons using UTC are preferable.

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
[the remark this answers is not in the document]
thorwolpert: the COD should have a submission datetime.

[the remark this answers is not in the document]
severinbeauvais: Yes, the COD could have an effective datetime < incorporation datetime, but isn't it up to other code to sort the transactions in order (eg, by filing id)?

I remember we once talked about injecting the "now" time into a COD's effective date but I'm not sure that will always work.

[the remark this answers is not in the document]
severinbeauvais: Example:
1. I incorporated a company 1 hour ago (effective 2020-08-19 17:30 UTC).
2. I just filed a COD (effective 2020-08-19 07:00 UTC).

Is the COD valid? Yes.
What code cares about the COD effective datetime?

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
severinbeauvais: OK, I am adding a check that the submission datetime (ie, "date" provided by the UI) is not less than the business founding date. This is thus a sanity check to validate that the UI is not providing corrupt data.

However, I cannot change the **effective datetime** of a COD -- it is 00:00 local on the date selected, which could be the same date they incorporated (but the COD **submission datetime** will be >= incorporation datetime as long as UI/BE times are close together and valid data is sent across).

legal-api/src/legal_api/services/filings/validations/change_of_directors.py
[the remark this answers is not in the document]
severinbeauvais: As in another conversation above, I am adding a check that COD submission datetime is >= business founding datetime.

As for the COD effective date-time, yes, it can appear to be earlier the IA. This is because the COD is effective on the selected date at 00:00 local. We just have to be careful to process the COD data (esp with an IA or other CODs on the same day) in submission datetime order (or filing id order).

PR #951, opened 2020-08-06
4256 colin bc maintenance filings

kialj876:
*Issue #:* /bcgov/entity#4256

*Description of changes:*
- see linked ticket for details

### DIFF
--- a/colin-api/src/colin_api/models/filing.py
+++ b/colin-api/src/colin_api/models/filing.py
@@ -23,7 +23,7 @@ from flask import current_app
 from colin_api.exceptions import FilingNotFoundException, InvalidFilingTypeException
 from colin_api.models import Address, Business, CorpName, Office, Party, ShareObject
 from colin_api.resources.db import DB
-from colin_api.utils import convert_to_json_date, convert_to_json_datetime, convert_to_pacific_time
+from colin_api.utils import convert_to_json_date, convert_to_json_datetime
 
 
 class Filing:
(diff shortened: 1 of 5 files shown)


colin-api/src/colin_api/models/filing.py
```diff
@@ -226,94 +235,104 @@ def _create_filing(cls, cursor, event_id: str,  # pylint: disable=too-many-argum
...
+                    event_id=filing.event_id,
+                    filing_type_code=filing.get_filing_type_code(),
+                    effective_dt=filing.effective_date,
                     period_end_date=ar_date if not agm_date else agm_date,
                     agm_date=agm_date
                 )
-            elif filing_type_code in ['OTADD', 'NOCAD', 'OTCDR', 'NOCDR', 'OTINC', 'BEINC']:
-                insert_stmnt = insert_stmnt + ', period_end_dt) '
-                values_stmnt = values_stmnt + ", TO_DATE(:period_end_date, 'YYYY-mm-dd'))"
+            elif filing_type_code in ['OTADD', 'OTCDR', 'OTINC']:
```
lekshmimallika-aot, line 251: Do you have to include the new fee code here BCADD, BCCDR etc.?

kialj876: No these are 'filing type codes' that are only used in colin

kialj876: We don't use these in our lear db at all -> in lear db we use 'changeOfAddress', etc. and no specific filing type code. The fee codes are different (we don't send those back to colin)

PR #959, opened 2020-08-11
Email trigger for maintenance filings

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4216

*Description of changes:*
Moved publish to email queue outside IA check
Changed the trigger message to include the actual filing status

### DIFF
--- a/queue_services/entity-pay/src/entity_pay/worker.py
+++ b/queue_services/entity-pay/src/entity_pay/worker.py
@@ -107,6 +107,15 @@ async def process_payment(payment_token, flask_app):
             db.session.add(filing_submission)
             db.session.commit()
 
+            try:
+                await publish_email_message(
+                    qsm, APP_CONFIG.EMAIL_PUBLISH_OPTIONS['subject'], filing_submission, Filing.Status.PAID.value)
+            except Exception as err:  # pylint: disable=broad-except, unused-variable # noqa F841;
+                # mark any failure for human review
+                capture_message(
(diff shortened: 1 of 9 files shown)


queue_services/entity-pay/src/entity_pay/worker.py
```diff
@@ -121,7 +121,7 @@ async def process_payment(payment_token, flask_app):
 
             try:
                 await publish_email_message(
-                    qsm, APP_CONFIG.EMAIL_PUBLISH_OPTIONS['subject'], filing_submission, 'filed')
+                    qsm, APP_CONFIG.EMAIL_PUBLISH_OPTIONS['subject'], filing_submission, filing_submission.status)
```
kialj876, line 124: We should move this to be before we publish to the filer queue (just in case the filer processes it super fast and then the status of the filing == completed before we publish the 'paid' email to the email queue). Not sure if that would ever happen, but I think its possible

lekshmimallika-aot: That is possible. Also if this message fails for some reason, when it gets reprocessed, it may send as complete. Do you think we should explicitly set the status to Filing.Status.PAID.value along with moving the code?

kialj876: Yes thats a great idea I think

kialj876:
@lekshmimallika-aot can you update the versions of these to 2.15.1 as well?

PR #963, opened 2020-08-13
small changes for adding email

kialj876:

*Issue #:* /bcgov/entity#4610

*Description of changes:*
- added contact point in for alterations


colin-api/src/colin_api/models/filing.py
```diff
@@ -109,8 +109,8 @@ def get_certified_by(self):
 
     def get_email(self):
         """Get email address."""
-        if self.body.get('incorporationApplication'):
-            return self.body['incorporationApplication']['contactPoint']['email']
+        if self.body.get(self.filing_type).get('contactPoint'):
```
thorwolpert, line 112: Doing this, making the default {}
```
self.body.get(self.filing_type,{}).get('contactPoint')
```
will avoid this
```
'NoneType' object has no attribute 'get'
```

PR #972, opened 2020-08-20
more COD date validation

severinbeauvais:
*Issue #:* /bcgov/entity#4311

*Description of changes:*
- added check that COD submission datetime not < business founding datetime
- added check that COD submission datetime not < other COD/AD submission datetime
- added/updated unit tests


severinbeauvais:
Thor pointed out that Filing Date (submission date) is actually calculated by the API upon filing -- so there is no need to validate whether the user provided a proper value. Therefore, this code change and PR are not necessary... so I'm closing it.

PR #976, opened 2020-08-24
Show Incorporation Number as `Pending` if file_type is incorporationApplication and status is PAID #4658

vysakh-menon-aot:
*Issue #:* /bcgov/entity#4658

*Description of changes:*
  - Show Incorporation Number as `Pending` if file_type is incorporationApplication and status is PAID

### DIFF
--- a/queue_services/entity-emailer/src/entity_emailer/email_templates/common/business-info.html
+++ b/queue_services/entity-emailer/src/entity_emailer/email_templates/common/business-info.html
@@ -19,14 +19,12 @@
 
   <tr>
     <td class="name">Incorporation Number:</td>
-    {% if header.status == 'PAID' %}
+    {% if header.status == 'PAID' and filing_type == 'incorporationApplication'
+    %}
     <td class="value">Pending</td>
-    {% elif header.status == 'COMPLETED' %}
+    {% else %}
     <!-- Eg, BC0878529 -->
     <td class="value">{{ business.identifier }}</td>
-    {% else %}
-    <!-- add other cases here -->
-    <td class="value">Unknown</td>
     {% endif %}
   </tr>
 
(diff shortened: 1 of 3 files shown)


queue_services/entity-emailer/src/entity_emailer/email_templates/common/business-info.html
```diff
@@ -19,14 +19,12 @@
 
   <tr>
     <td class="name">Incorporation Number:</td>
-    {% if header.status == 'PAID' %}
+    {% if header.status == 'PAID' and filing_type == 'incorporationApplication'
+    %}
```
severinbeauvais, line 23: I think there is confusion here. The bug is that the Incorporation Number displays as "Pending"; instead the actual incorporation number should be displayed.

vysakh-menon-aot: It shows "Pending" only when the filing_type is incorporationApplication and status is Paid (at this point we don't have  Incorporation Number). In all other cases (including COD, COA) it shows Incorporation Number. Isn't that the expectation?

severinbeauvais: The description of the bug refers to a Director Change (aka COD) filing confirmation email. In this case the business is already incorporated and has an incorporation number.
I am not sure that this code change is necessary.

thorwolpert: We should use the term **identifier**, as in our model, taken from json-ld/schema.org
Using too many different terms, will make it less clear.

vysakh-menon-aot: @severinbeauvais at this case the filing_type will be `changeOfDirectors`, so it shows incorporation number.

vysakh-menon-aot: @thorwolpert I am not aware of json-ld/schema.org, can you explain that?

severinbeauvais: Vysakh wrote:
> at this case the filing_type will be changeOfDirectors, so it shows incorporation number

Aaah, OK, now I get it: you're saying that this code inadvertently displayed "Pending" for all filings (including COD), but this code should apply only to IAs.

severinbeauvais: Please create a new conversation for the naming/schema comment, and then resolve THIS conversation.

PR #978, opened 2020-08-26
Share Structure update for reports and colin api

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4668

*Description of changes:*
 Updated colin api
 Updated colin api postman collection
 Updated report.py


legal-api/src/legal_api/reports/report.py
```diff
@@ -266,7 +266,10 @@ def _format_incorporation_data(self, filing):
         filing['listOfTranslations'] = filing['incorporationApplication'].get('nameTranslations', {})\
             .get('new', [])
         filing['offices'] = filing['incorporationApplication']['offices']
-        filing['shareClasses'] = filing['incorporationApplication']['shareClasses']
+        if filing['incorporationApplication'].get('shareClasses', None):
+            filing['shareClasses'] = filing['incorporationApplication']['shareClasses']
```
lekshmimallika-aot, line 270: This is to support existing incorporation applications which has the old structure in the filing json

lekshmimallika-aot:
> Looks good, but can you also change the postman tests in the share structure folder and the pytests for getting share structure!

The change is only in post and get IA. The Get Share Classes has no change in structure. Is that what you are referring to?

PR #980, opened 2020-08-27
4668 filer update ia alteration share structure

thorwolpert:
*Issue #:* /bcgov/entity#4668

*Description of changes:*
- created function to remove keys from dict type objects
- create common shareStructure processor
- change IA to use the new shareStructure processor
- change Alteration to use the new shareStructure processor


thorwolpert:
code smells:
in the testing helpers, these seem reasonable to me to ignore
other bare exception is also reasonable based on the catches.

PR #1006, opened 2020-09-23
5006 legal type change

kialj876:
*Issue #:* /bcgov/entity#5006

*Description of changes:*
- legal-api portion for this ticket (breaking it up so that it's smaller PR)

### DIFF
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -50,7 +50,8 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes
         """Render an Enum of the Business Legal Types."""
 
         COOP = 'CP'
-        BCOMP = 'BC'
+        COMP = 'BC'
+        BCOMP = 'BEN'
 
     __versioned__ = {}
     __tablename__ = 'businesses'
--- a/legal-api/src/legal_api/models/filing.py
+++ b/legal-api/src/legal_api/models/filing.py
@@ -25,6 +25,7 @@ from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import backref
 
 from legal_api.exceptions import BusinessException
+from legal_api.models import Business
 from legal_api.models.colin_event_id import ColinEventId
 from legal_api.schemas import rsbc_schemas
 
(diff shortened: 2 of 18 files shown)


legal-api/src/legal_api/models/filing.py
```diff
@@ -58,16 +58,54 @@ class Source(Enum):
...
+            'title': 'Notice of Alteration Filing',
+            'codes': {
+                'BC': 'ALTER'
+            }
+        },
+        'annualReport': {
+            'name': 'annualReport',
+            'title': 'Annual Report Filing',
+            'codes': {
+                'CP': 'OTANN',
```
lekshmimallika-aot, line 74: Nice!

legal-api/src/legal_api/models/business.py
```diff
@@ -50,7 +50,8 @@ class LegalTypes(Enum):
         """Render an Enum of the Business Legal Types."""
 
         COOP = 'CP'
-        BCOMP = 'BC'
+        COMP = 'BC'
+        BCOMP = 'BEN'
```
severinbeauvais, line 54: Do the UIs have to change anything?
See https://github.com/bcgov/business-filings-ui/blob/master/src/enums/legalTypes.ts

legal-api/src/legal_api/reports/report.py
```diff
@@ -364,5 +373,6 @@ class ReportMeta:  # pylint: disable=too-few-public-methods
 
     entity_description = {
         'CP': 'cooperative',
-        'BC': 'BC Benefit Company'
+        'BC': 'BC Company',
+        'BEN': 'BC Benefit Company',
```
severinbeauvais, line 377: See https://github.com/bcgov/business-filings-ui/blob/master/src/enums/legalNames.ts

legal-api/src/legal_api/models/business.py
lekshmimallika-aot: There is ticket to update UIs.

https://github.com/bcgov/entity/issues/5048

legal-api/src/legal_api/reports/report.py
kialj876: Hey sev, I don't think that isn't relevant here. This object maps the backend legal type to the description. Unless you're saying we should be exporting the description from the front end, but I would disagree with that because we will be implementing a legal type/class table that we will pull the legal type/description from in the future

kialj876: I guess I can update this to pull the types from our temp class in business

kialj876: Or are you saying that 'cooperative' is an existing bug and it should be 'BC Cooperative'?

lekshmimallika-aot (approved):
Looks good!

legal-api/src/legal_api/reports/report.py
severinbeauvais: I was thinking that we would want legalTypes and legalNames to be the same on both sides.

kialj876: Oh okay -- I updated it to pull the value from the LegalTypes object in the backend (it needs to follow that)

kialj876:
> OK just leave it.

I'm afraid if I change it might break tests or cause a failure somewhere unexpected :P

severinbeauvais (approved):

legal-api/src/legal_api/reports/report.py
thorwolpert: It should probably be called legalTypes everywhere, as that was the original name in the schema to reduce confusion.
I'd assume legalName was the businesses legal name.

legal-api/src/legal_api/reports/report.py
```diff
@@ -363,6 +372,7 @@ class ReportMeta:  # pylint: disable=too-few-public-methods
     }
 
     entity_description = {
-        'CP': 'cooperative',
-        'BC': 'BC Benefit Company'
+        Business.LegalTypes.COOP.value: 'cooperative',
```
thorwolpert, line 375: Everything else is init capped, other than cooperatives? Should it be consistent?

legal-api/src/legal_api/reports/report.py
severinbeauvais: In the UI, `legalNames` is an enum to map the legal type name string. FWIW

legal-api/src/legal_api/reports/report.py
thorwolpert: @kialj876 is going to open a tech debt tix for this. With that, it's resolved in the is PR.

thorwolpert (approved):

legal-api/src/legal_api/reports/report.py
kialj876: https://github.com/bcgov/entity/issues/5088 made for this

PR #1016, opened 2020-10-01
Fix - Default value of corp type if legal type is not present in filing

lekshmimallika-aot:
*Issue #:* /bcgov/entity###

*Description of changes:*
We should not be looking at the business identifier prefix to identify the corp type in case the legal type is not
present in the business json. We should take that value from the business model.

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -556,7 +556,7 @@ class ListFilingResource(Resource):
 
         else:
             mailing_address = business.mailing_address.one_or_none()
-            corp_type = filing.json['filing']['business'].get('legalType', business.identifier[:-7])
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -556,7 +556,7 @@ def _create_invoice(business: Business,  # pylint: disable=too-many-locals
 
         else:
             mailing_address = business.mailing_address.one_or_none()
-            corp_type = filing.json['filing']['business'].get('legalType', business.identifier[:-7])
+            corp_type = filing.json['filing']['business'].get('legalType', business.legal_type)
```
kialj876, line 559: nice catch -- can we actually make this the other way around though? I don't think we should trust the json from the front end for legal type unless we have to

kialj876 (commented):
Looks good! One comment, but I leave the decision to you haha

kialj876 (approved):

PR #1017, opened 2020-10-01
Fix for corrections adding fee for sub filing type

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4642

*Description of changes:*
Currently if it is a correction filing it adds the fee for the sub filing types under it. If it is a correction it has a defined fee for each entity type and it should not consider the sub filing types.

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -494,38 +494,45 @@ class ListFilingResource(Resource):
             business = Business.find_by_identifier(filing_json['filing']['business']['identifier'])
             legal_type = business.legal_type
 
-        for k in filing_json['filing'].keys():
+        if any('correction' in x for x in filing_json['filing'].keys()):
+            filing_type_code = Filing.FILINGS.get('correction', {}).get('codes', {}).get(legal_type)
+            filing_types.append({
+                'filingTypeCode': filing_type_code,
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -494,38 +494,45 @@ def _get_filing_types(filing_json: dict):
...
-                    'filingTypeCode': filing_type_code,
-                    'futureEffective': ListFilingResource._is_future_effective_filing(filing_json)
-                })
-            elif filing_type_code:
-                filing_types.append({
-                    'filingTypeCode': filing_type_code,
-                    'priority': priority,
-                    'waiveFees': filing_json['filing']['header'].get('waiveFees', False)
-                })
+        if any('correction' in x for x in filing_json['filing'].keys()):
```
kialj876, line 497: is this if statement needed? It looks like it would get picked up normally in the code as it was

kialj876 (changes requested):
One comment/question and also pls add a test case in the fee code pytests!

legal-api/src/legal_api/resources/business/business_filings.py
kialj876: oh nevermind! I see what the issue was after looking at the ticket -- disregard this comment

kialj876 (approved):

PR #1024, opened 2020-10-07
Draft GET core filing

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5197,  /bcgov/entity#5198
*Description of changes:*


legal-api/src/legal_api/core/filing.py
```diff
@@ -15,10 +15,11 @@
 from __future__ import annotations
 
 # from dataclasses import dataclass, field
+from datetime import date
 from enum import Enum
 from typing import Dict, Optional
 
-from legal_api.models import Filing as FilingStorage
+from legal_api.models import Business, Filing as FilingStorage, db
 from legal_api.utils.datetime import datetime
```
thorwolpert, line 23: Add the date import here

legal-api/src/legal_api/core/filing.py
```diff
@@ -61,17 +62,36 @@ def id(self) -> str:  # pylint: disable=invalid-name; defining the std ID
                 self._id = self._storage.id
         return self._id
 
+    @property
+    def filing_type(self) -> str:
+        """Property containing the filing type."""
+        if not self._filing_type:
+            if self._storage:
```
thorwolpert, line 69: The linter will ask you to make this an AND, instead of a nested if

legal-api/src/legal_api/core/filing.py
```diff
@@ -61,17 +62,36 @@ def id(self) -> str:  # pylint: disable=invalid-name; defining the std ID
...
+                self._raw = self._storage.json
+            else:
+                return {}
         return self._raw
 
     @property
     def json(self) -> Optional[Dict]:
         """Return a dict representing the filing json."""
-        return self._raw
+        if self._storage.status == Filing.Status.COMPLETED.value:
```
thorwolpert, line 86: need to check for storage first

legal-api/src/legal_api/core/filing.py
```diff
@@ -89,3 +109,38 @@ def save(self):
...
+        if identifier.startswith('T'):
+            q = db.session.query(FilingStorage). \
+                filter(FilingStorage.temp_reg == identifier)
+
+            if filing_id:
+                q = q.filter(FilingStorage.id == filing_id)
+
+            filing._storage = q.one_or_none()
+        else:
+            q = db.session.query(Business, FilingStorage). \
```
thorwolpert, line 127: This should call the storage model for the filing, not talk to the DB directly.

legal-api/src/legal_api/core/filing.py
```diff
@@ -61,17 +62,44 @@ def id(self) -> str:  # pylint: disable=invalid-name; defining the std ID
...
+        if self._storage:
+            if self._storage.status == Filing.Status.COMPLETED.value:
+                _json = VersionedBusinessDetailsService. \
+                    get_revision(self.filing_type, self.id, self._storage.business_id)
+            else:
+                _json = self.raw
+        return _json
+
+    @property
+    def storage(self) -> Optional[FilingStorage]:
```
lekshmimallika-aot, line 95: Not sure whether this storage which is the filing model itself should be exposed as a property like this or you should have getters for each of the property defined in the property list.

legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -88,28 +83,25 @@ def get(identifier, filing_id=None):  # pylint: disable=too-many-return-statemen
...
 
             if str(request.accept_mimetypes) == 'application/pdf':
                 report_type = request.args.get('type', None)
-                if rv[1].filing_type == 'incorporationApplication':
-                    ListFilingResource._populate_business_info_to_filing(rv[1], business)
-                return legal_api.reports.get_pdf(rv[1], report_type)
-            return jsonify(rv[1].json)
+                if rv.filing_type == 'incorporationApplication':
+                    ListFilingResource._populate_business_info_to_filing(rv.storage, business)
+                return legal_api.reports.get_pdf(rv.storage, report_type)
```
lekshmimallika-aot, line 94: Please validate whether the IA application is getting downloaded after this change.

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -34,9 +34,68 @@
...
+                VersionedBusinessDetailsService.get_ia_revision(filing, business)
+        elif filing_type == 'changeOfDirectors':
+            pass
+        elif filing_type == 'changeOfAddress':
+            pass
+        elif filing_type == 'annualReport':
+            pass
+        elif filing_type == 'correction':
+            pass
+        elif filing_type == 'alteration':
```
lekshmimallika-aot, line 61: Even though you dont have reports for other type of filings like specialResolution, voluntary dissolution, change of name etc. I think we have to handle that here

legal-api/src/legal_api/core/filing.py
```diff
@@ -61,17 +62,44 @@ def id(self) -> str:  # pylint: disable=invalid-name; defining the std ID
...
+        if not self._filing_type and self._storage:
+            self._filing_type = self._storage.filing_type
+        return self._filing_type
+
     @property
     def raw(self) -> Optional[Dict]:
         """Return the raw, submitted and unprocessed version on the filing."""
         if not self._raw:
-            return {}
+            if self._storage:
```
thorwolpert, line 76: should be collapsed to AND

legal-api/src/legal_api/core/filing.py
```diff
@@ -61,17 +62,44 @@ def id(self) -> str:  # pylint: disable=invalid-name; defining the std ID
...
+                self._raw = self._storage.json
+            else:
+                return {}
```
thorwolpert, line 79: can return None now as Dict is Optional (sorry original goof on my part)

legal-api/src/legal_api/core/filing.py
```diff
@@ -61,17 +62,44 @@ def id(self) -> str:  # pylint: disable=invalid-name; defining the std ID
...
     def json(self) -> Optional[Dict]:
         """Return a dict representing the filing json."""
-        return self._raw
+        _json = {}
+        if self._storage:
+            if self._storage.status == Filing.Status.COMPLETED.value:
+                _json = VersionedBusinessDetailsService. \
+                    get_revision(self.filing_type, self.id, self._storage.business_id)
+            else:
+                _json = self.raw
```
thorwolpert, line 91: i think it should just return raw here and not set _json

legal-api/src/legal_api/core/filing.py
```diff
@@ -61,17 +62,44 @@ def id(self) -> str:  # pylint: disable=invalid-name; defining the std ID
...
 
     @property
     def json(self) -> Optional[Dict]:
         """Return a dict representing the filing json."""
-        return self._raw
+        _json = {}
+        if self._storage:
+            if self._storage.status == Filing.Status.COMPLETED.value:
+                _json = VersionedBusinessDetailsService. \
+                    get_revision(self.filing_type, self.id, self._storage.business_id)
```
thorwolpert, line 89: Since you are passing the filing id, and then retrieving the filing, do we need to pass in the type?

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -34,9 +34,97 @@
 )
 
 
-class VersionedBusinessDetailsService:
+class VersionedBusinessDetailsService:  # pylint: disable=too-many-public-methods
     """Provides service for getting business details as of a filing."""
 
+    @staticmethod
+    def get_revision(filing_type, filing_id, business_id):
```
thorwolpert, line 41: I like this abstraction 👍

legal-api/src/legal_api/core/filing.py
vysakh-menon-aot: I had to expose model for the purpose of get_pdf

legal-api/src/legal_api/resources/business/business_filings.py
vysakh-menon-aot: Will do

legal-api/src/legal_api/core/filing.py
vysakh-menon-aot: That's true. Will change it.

legal-api/src/legal_api/resources/business/business_filings.py
lekshmimallika-aot: What I meant was with this report will still use the filing model  and it does not use the logic of raw or versioned json at this point.   There will be refactoring needed to make that work

lekshmimallika-aot: Another thing to verify is that UI continues to display incomplete and complete filings properly with this change atleasr for IA. You can run the UI locally pointing to these changes to test that

lekshmimallika-aot:
These get revision methods were originally written for NOA and it kind of reflects the company profile once a filing is completed.
Even though you can reuse most of the methods, there will be some exceptions

get_party_role_revision returns all party roles. For example for a BC company, if you are filing cod after an IA application I think a completed cod filing will return all the parties with all roles with this. If that is true you may have to include a filter for directors

Second one would be Name Translations. Name translations schema allow only new, modified and ceased entries. So if you have a filing with name translation it may not return that correctly with this method as this is supposed to show all the translations upto a point of time.

legal-api/src/legal_api/core/filing.py
thorwolpert: @lekshmimallika-aot I suggested we can move this in in stages as needed, and up to @vysakh-menon-aot to suggest in the PR where the lines should be.
Can this property be marked as Deprecated with a note, in case someone else does the report updates?

legal-api/src/legal_api/core/filing.py
```diff
@@ -45,7 +46,7 @@ def __init__(self):
         self._raw: Optional[Dict] = None
         self._completion_date: datetime
         self._filing_date: datetime
-        self._filing_type: str
+        self._filing_type: str = ''
```
thorwolpert, line 49: Looking feedback from @lekshmimallika-aot and @kialj876 as well.
Should we do
```
Optional[str]
```
or
```
str=''
```
I'm thinking we should note something about our coding standards here as we provide no guidance on typing yet.

thorwolpert (commented):
Please add test cases to cover / validate the changes.

legal-api/src/legal_api/core/filing.py
lekshmimallika-aot: Optional looks good to me

legal-api/src/legal_api/core/filing.py
lekshmimallika-aot: Sure. Sounds good

lekshmimallika-aot (approved):

lekshmimallika-aot (commented):
Could you please add a few tests for the enrichment filing (for BC and CP). I might have missed some scenarios when we discussed this. Scenarios for CP would be 

without any changes
with director change only
with address change only
with both director and address changes

thorwolpert (approved):

PR #1026, opened 2020-10-14
Registrar Signature block - reusable template

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4638

This enables us to fetch the correct registrar info based on the filing effective date.I think we need this in all outputs that contains the registrar signature (Not just the certificate). All other outputs have registrar info hardcoded atm.

I thought of putting this in registrar data in db as opposed to the meta json. The only advantage I could see is adding a new registrar/modifying the data will not need any code changes and deployments. But I dont know how frequently that will happen. So we may have to compare the expense of db call vs data update frequency

*Description of changes:*
Added a new utility class to get the registrar info based on the filing effective date
Removed the hardcoded base64 image data and added the code to generate it while fetching the registrar info
Modified the certificate to use the data returned by the meta info class


thorwolpert:
I think over the past 15yrs we've had 5 Registrars.
The addition of a new Registrar name and signature would be added by an SRE Team member, so I don't think having it in the fs is an issue.
My vote would be to leave it as you've designed it.

thorwolpert (approved):
nice!

PR #1027, opened 2020-10-14
5236 fix for modified name translation in correction/alteration

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5236

*Description of changes:*

### DIFF
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/aliases.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/aliases.py
@@ -31,7 +31,7 @@ def update_aliases(business: Business, aliases: Dict) -> Dict:
     if modified_aliases := aliases.get('modified'):
         for current_alias in business.aliases.all():
             for mod_alias in modified_aliases:
-                if current_alias.alias == str(mod_alias.get('oldValue')).upper():
+                if current_alias.alias.upper() == str(mod_alias.get('oldValue')).upper():
                     current_alias.alias = str(mod_alias.get('newValue')).upper()
 
     if new_aliases := aliases.get('new'):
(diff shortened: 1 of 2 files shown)


queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/aliases.py
```diff
@@ -31,7 +31,7 @@ def update_aliases(business: Business, aliases: Dict) -> Dict:
     if modified_aliases := aliases.get('modified'):
         for current_alias in business.aliases.all():
             for mod_alias in modified_aliases:
-                if current_alias.alias == mod_alias.get('oldValue'):
+                if current_alias.alias == str(mod_alias.get('oldValue')).upper():
```
kialj876, line 34: Can you add the .upper to both sides of this comparison? Not %100 sure the data will always be consistent when we load corps from COLIN so we might have some cases where current_alias.alias is mixed case

kialj876 (commented):
Looks good, just one comment. Also could you add in a test case for a mixed case current_alias.alias and mixed case mod_alias.get('oldValue')?

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/aliases.py
vysakh-menon-aot: Sure. I did not add based on the implementation below for "new" aliases

lekshmimallika-aot:
Another point discussed for this ticket during grooming was the correction output may not show the edited ones as it reuses the IA template and shows only the new entries as of now.But I see that this was not listed in the ticket. Please create a new ticket to fix the correction IA output / you can fix in this one. Up to you.

lekshmimallika-aot (approved):

kialj876 (approved):
Nice work!

PR #1029, opened 2020-10-19
5200 audit filing

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5200

*Description of changes:*
  - query param to get original filing json


legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -73,7 +74,7 @@ def get(identifier, filing_id=None):  # pylint: disable=too-many-return-statemen
             if str(request.accept_mimetypes) == 'application/pdf' and filing_id:
                 if rv.filing_type == 'incorporationApplication':
                     return legal_api.reports.get_pdf(rv.storage, None)
-            filing_json = rv.json
+            filing_json = rv.raw if original_filing else rv.json
             filing_json['filing']['documents'] = DocumentMetaService().get_documents(filing_json)
```
thorwolpert, line 78: Do we want to include the list of documents on the raw json, as we may not be able to render it directly anymore?

legal-api/tests/unit/api/test_business_filings/test_filings.py
```diff
@@ -111,6 +111,24 @@ def test_get_one_business_filing_by_id(session, client, jwt):
     assert rv.json['filing']['business'] == ar['filing']['business']
 
 
+def test_get_one_business_filing_by_id_raw_json(session, client, jwt):
+    """Assert that the business info cannot be received in a valid JSONSchema format."""
```
thorwolpert, line 115: This doesn't seem to reflect the test.

thorwolpert (commented):
A couple of comments to review

legal-api/tests/unit/api/test_business_filings/test_filings.py
vysakh-menon-aot: Yeah. It was just a try. Since there is no difference in json to identify whether its a raw or json I left it as is. Do you have any suggestion to test this scenario or should I remove it?

legal-api/src/legal_api/resources/business/business_filings.py
vysakh-menon-aot: So when user request for original filing we don't need to include documents. isn't it?

thorwolpert: Yes, I don't think that should be included. It should just be the raw json originally submitted by the customer.

We won't be "back" supporting outputs, as those will always be created using the versioned data in simple filing.

legal-api/tests/unit/api/test_business_filings/test_filings.py
thorwolpert: It'll fail linting without a comment. I'd suggest something like.
Assert that the raw json originally submitted is returned.

legal-api/src/legal_api/resources/business/business_filings.py
vysakh-menon-aot: ok

legal-api/tests/unit/api/test_business_filings/test_filings.py
vysakh-menon-aot: ok

legal-api/tests/unit/api/test_business_filings/test_filings.py
```diff
@@ -112,6 +112,24 @@ def test_get_one_business_filing_by_id(session, client, jwt):
...
+    filings = factory_filing(b, ANNUAL_REPORT)
+    ar = copy.deepcopy(ANNUAL_REPORT)
+    ar['filing']['header']['filingId'] = filings.id
+    ar['filing']['header']['colinIds'] = []
+
+    rv = client.get(f'/api/v1/businesses/{identifier}/filings/{filings.id}?original=true',
+                    headers=create_header(jwt, [STAFF_ROLE], identifier))
+
+    assert rv.status_code == HTTPStatus.OK
+    assert rv.json['filing']['annualReport'] == ar['filing']['annualReport']
```
lekshmimallika-aot, line 129: Not sure of the purpose of lines  121, 122 and 123. Here cant it be compared like ANNUAL_REPORT['filing']['annualReport']

lekshmimallika-aot (approved):

thorwolpert (approved):

PR #1032, opened 2020-10-21
Feature/simple filing fix

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5197

*Description of changes:*
 - Handled 2 diff structure of role one for IA and the other for COD, AR and NOA

### DIFF
--- a/legal-api/src/legal_api/services/business_details_version.py
+++ b/legal-api/src/legal_api/services/business_details_version.py
@@ -236,7 +236,8 @@ class VersionedBusinessDetailsService:  # pylint: disable=too-many-public-method
             .filter(party_role_version.transaction_id <= transaction_id) \
             .filter(party_role_version.operation_type != 2) \
             .filter(party_role_version.business_id == business_id) \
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -393,7 +405,11 @@ def party_revision_json(transaction_id, party_revision) -> dict:
             if party_revision.delivery_address:
                 member['mailingAddress'] = member['deliveryAddress']
 
-        member['id'] = party_revision.id
+        if is_ia_or_after:
+            member['officer']['id'] = party_revision.id
```
lekshmimallika-aot, line 409: Convert these ids to string type. ids are there in almost all objects except addresses I think. Whereever id is being returned currently add convert that to string

vysakh-menon-aot: ok

lekshmimallika-aot (approved):
Please add the string casting to other ids as well

thorwolpert (approved):

PR #1046, opened 2020-10-30
5185 Output - Make New Transition Application output

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5185

*Description of changes:*
 - Transition Application report


legal-api/report-templates/transitionApplication.html
```diff
@@ -0,0 +1,56 @@
...
+         </div>
+         <div class="section-data mt-4">
+            <span class="bold">Incorporation Number:</span>
+            {{business.identifier}}
+         </div>
+      </div>
+        
+      <div class="section-header mt-5 pt-4">NOTICE OF ARTICLES</div>
+      <div class="container pt-4">
+         [[notice-of-articles/benefitCompanyStmt.html]]
```
lekshmimallika-aot, line 48: This should not be there . Although this output is almost the same as NOA I dont think this section should be there. Check with Linda

legal-api/report-templates/template-parts/transition/businessDetails.html
```diff
@@ -0,0 +1,31 @@
+<div class="business-details-container">
```
lekshmimallika-aot, line 1: This section need not be a new file. It can be any of the existing business details I think the only difference is Transition date which you can definitely add based on the type of filing

thorwolpert (approved):

lekshmimallika-aot (approved):

lekshmimallika-aot (approved):

PR #1048, opened 2020-11-03
Fix to use the versioned business info only if transaction id is present

lekshmimallika-aot:
*Issue #:* /bcgov/entity#N/A

*Description of changes:*
The versioned business detail depends on the transaction id and is available only after a filing is complete. For a future effective
filing or immediately after the payment is done, transaction id will not be present in the filing and this causes issue while trying to download the report of a filing in paid status


vysakh-menon-aot (approved):

thorwolpert (approved):

kialj876 (approved):
Hey Lekshmi this looks good, but if you have time and its quick, can you add add in a test case? This was a tiny change so if not that's fine

lekshmimallika-aot:
> Hey Lekshmi this looks good, but if you have time and its quick, can you add add in a test case? This was a tiny change so if not that's fine

Sure. I will do it in another PR. This currently breaks the emails in test

PR #1052, opened 2020-11-03
Fix party version

vysakh-menon-aot:
*Issue #:* /bcgov/entity###

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/services/business_details_version.py
+++ b/legal-api/src/legal_api/services/business_details_version.py
@@ -91,6 +91,19 @@ class VersionedBusinessDetailsService:  # pylint: disable=too-many-public-method
             VersionedBusinessDetailsService.get_name_translations_revision(filing.transaction_id, business.id)
         ia_json['incorporationApplication']['incorporationAgreement'] = \
             VersionedBusinessDetailsService.get_incorporation_agreement_json(filing)
+
+        # setting completing party email from filing json
+        party_email = ''
+        for party in filing.json['filing']['incorporationApplication']['parties']:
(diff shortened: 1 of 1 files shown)


severinbeauvais (approved):
Should this have a check in case 'email' isn't present?

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -96,7 +96,7 @@ def get_ia_revision(filing, business) -> dict:
         party_email = ''
         for party in filing.json['filing']['incorporationApplication']['parties']:
             if next((x for x in party['roles'] if x['roleType'] == 'Completing Party'), None):
-                party_email = party['email']
+                party_email = party['officer']['email']
```
lekshmimallika-aot, line 99: To avoid issues like this it is a good practice to use something like this
party.get('officer',{}).get('email', None)

vysakh-menon-aot:
> Should this have a check in case 'email' isn't present?

Email is a mandatory field when its a completing party. Right?

lekshmimallika-aot (approved):

legal-api/src/legal_api/services/business_details_version.py
vysakh-menon-aot: ok

severinbeauvais:
> > Should this have a check in case 'email' isn't present?
> 
> Email is a mandatory field when its a completing party. Right?

Theoretically yes. Unless for some reason it couldn't be obtained.

+1 Lekshmi

severinbeauvais (approved):
LGTM

PR #1055, opened 2020-11-04
single query instead of multiple - party address

vysakh-menon-aot:
*Issue #:* /bcgov/entity#4936

*Description of changes:*
 - Fixed rewriting csv file by making into a single query


kialj876 (approved):
Ah I see so the old version was overwriting the delivery addresses with mailing. Nice work!

PR #1061, opened 2020-11-12
5199 corrected filing diff

thorwolpert:
*Issue #:* /bcgov/entity#5199

*Description of changes:*
- added generic diff for dictionaries
- added generic diff for lists made up of dictionaries with a unique ID
- added a diff block to correction filing types


legal-api/src/legal_api/core/filing.py
```diff
@@ -90,17 +115,35 @@ def payment_account(self, account_identifier):
...
+
     @property
     def json(self) -> Optional[Dict]:
         """Return a dict representing the filing json."""
         _json = {}
         if self._storage:
             if self._storage.status == Filing.Status.COMPLETED.value:
                 _json = VersionedBusinessDetailsService.get_revision(self.id, self._storage.business_id)
+
+                if self.filing_type == Filing.FILING_TYPES.CORRECTION.value:
```
lekshmimallika-aot, line 133: Should this be outside the completed check? We have the reports available even before the filing gets completed and that time also we may need this diff

legal-api/src/legal_api/core/filing.py
```diff
@@ -117,15 +160,19 @@ def storage(self, filing: FilingStorage):
...
     def save(self):
         """Save the filing."""
-        self.storage.filing_json = self._raw
-        self._storage.save()
+        raise NotImplementedError
+
+    def diff(self):
+        """Return the diff block for the filing this one corrects, if any."""
+        if self.json and self.status == Filing.Status.COMPLETED.value:
+            if correction_id := self.json.get('correction'):
```
lekshmimallika-aot, line 170: Should this be looking for correctedFilingId inside correction object

lekshmimallika-aot (approved):

legal-api/src/legal_api/core/filing.py
thorwolpert: These are all the states:
```
        COMPLETED = 'COMPLETED'
        DRAFT = 'DRAFT'
        EPOCH = 'EPOCH'
        ERROR = 'ERROR'
        PAID = 'PAID'
        PENDING = 'PENDING'
        PAPER_ONLY = 'PAPER_ONLY'
        PENDING_CORRECTION = 'PENDING_CORRECTION'
```
So probably should be available only in 
```
        COMPLETED = 'COMPLETED'
        PAID = 'PAID'
        PENDING = 'PENDING'
```

legal-api/src/legal_api/core/filing.py
thorwolpert: Yep, will get that in with the test to cover it.

legal-api/src/legal_api/core/filing.py
lekshmimallika-aot: Yes, I think so too

thorwolpert: Cool, on it. thx

thorwolpert:
Code Smells:
- Nested ifs, seems it has issues with the := walrus and using that in ifs
- Duplicate strings in tests, won't correct that

legal-api/src/legal_api/core/filing.py
```diff
@@ -90,16 +115,34 @@ def payment_account(self, account_identifier):
...
-            else:
-                return self.raw
-        return _json
+        if not self._storage or (self._storage and self._storage.status not in [Filing.Status.COMPLETED.value,
+                                                                                Filing.Status.PAID.value,
+                                                                                Filing.Status.PENDING.value,
+                                                                                ]):
+            return self.raw
+
+        filing_json = VersionedBusinessDetailsService.get_revision(self.id, self._storage.business_id)
```
lekshmimallika-aot, line 134: Versioned data is available only once the filing is complete because transaction id will be populated by the filer. For paid and pending shouldnt it add the diff to the raw json?

lekshmimallika-aot (approved):
Looks great!

thorwolpert:
@kialj876 is away, but approved this in the chat.

PR #1062, opened 2020-11-12
5238 refactored add_filing to make it more generic

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5238

*Description of changes:*

### DIFF
--- a/colin-api/src/colin_api/models/filing.py
+++ b/colin-api/src/colin_api/models/filing.py
@@ -997,6 +997,7 @@ class Filing:
             raise err
 
     @classmethod
+    # pylint: disable=too-many-arguments;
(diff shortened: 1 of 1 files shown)


kialj876 (changes requested):
Looks very good! Only a couple small issues in here: the AR changes the business name to '<corp num> B.C. LTD.' and the COA fails for bcomps with a ledger text error (we don't need to put an entry into the ledger text table for bcomps so can skip the entry if legal type is a benefit company)

kialj876 (approved):
Hey Vysakh, after looking at this again I think the ledger text insert needs to be altered a bit. There can only ever be 1 entry to the ledger text table per event (so 1 _insert_ledger_text() call per filing). I will approve this and then alter it on my branch. Nice work!

PR #1073, opened 2020-11-23
Legal API changes - Name Translations

lekshmimallika-aot:
*Issue #:* /bcgov/entity#5717

*Description of changes:*
Updated Versioned service
Updated incorporation application reports template. Corrections template will need additional work
Updated the keyname from alias to name in name translations to be in sync with schema
Updated tests

### DIFF
--- a/legal-api/src/legal_api/models/alias.py
+++ b/legal-api/src/legal_api/models/alias.py
@@ -48,7 +48,7 @@ class Alias(db.Model):  # pylint: disable=too-many-instance-attributes
         """Return a dict of this object, with keys in JSON format."""
         alias = {
             'id': self.id,
-            'alias': self.alias,
+            'name': self.alias,
             'type': self.type
         }
         return alias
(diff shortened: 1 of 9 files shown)


thorwolpert (approved):

legal-api/src/legal_api/models/alias.py
```diff
@@ -48,7 +48,7 @@ def json(self):
         """Return a dict of this object, with keys in JSON format."""
         alias = {
             'id': self.id,
-            'alias': self.alias,
+            'name': self.alias,
```
vysakh-menon-aot, line 51: edit-ui need a change for this. For alteration in edit-ui this response has been used.

lekshmimallika-aot: Ya Thats right. Edit-UI will need an update. alias is a property in the model. In the json it should be in sync with what we defined in schema

vysakh-menon-aot (approved):

PR #1077, opened 2020-11-25
Name Translation Changes colin API

lekshmimallika-aot:
*Issue #:* /bcgov/entity#5724

*Description of changes:*
Updated process name translations and get filing method
Updated postman collection


kialj876:
Hey Lekshmi, are you also going to do a PR that changes how the filer processes name translation if the filing is from colin?

colin-api/tests/postman/colin-api.postman_collection.json
```diff
@@ -9423,7 +9423,7 @@
...
...n\t\t                    \"priority\": 1,\n\t\t                    \"hasMaximumShares\": true,\n\t\t                    \"maxNumberOfShares\": 50,\n\t\t                    \"hasRightsOrRestrictions\": false\n\t\t                },\n\t\t                {\n\t\t                    \"id\": 2,\n\t\t                    \"name\": \"Share Series 2\",\n\t\t                    \"priority\": 2,\n\t\t                    \"hasMaximumShares\": true,\n\t\t                    \"maxNumberOfShares\": 100,\n\t\t                    \"hasRightsOrRestrictions\": false\n\t\t                }\n\t\t            ]\n\t\t        },\n\t\t        {\n\t\t            \"id\": 2,\n\t\t            \"name\": \"Share Class 2\",\n\t\t            \"priority\": 1,\n\t\t            \"hasMaximumShares\": false,\n\t\t            \"maxNumberOfShares\": null,\n\t\t            \"hasParValue\": false,\n\t\t            \"parValue\": null,\n\t\t            \"currency\": null,\n\t\t            \"hasRightsOrRestrictions\": true,\n\t\t            \"series\": []\n\t\t        }\n\t\t    ]\n    \t\t},\n\t\t\t\"contactPoint\": {\n\t\t\t\t\"email\": \"contact@point.incorp\",\n\t\t\t\t\"phone\": \"123-456-7890\"\n\t\t\t}\n\t\t}\n    }\n}",
```
kialj876, line 9426: This one just needs an update on the name translation section -- currently fails because it has the old format

lekshmimallika-aot:
> Hey Lekshmi, are you also going to do a PR that changes how the filer processes name translation if the filing is from colin?

Yes. Also Legal API postman collection needs to be updated. Will include those two in next PR

kialj876 (changes requested):
Looks very good, just one change in one pm test needed

colin-api/tests/postman/colin-api.postman_collection.json
lekshmimallika-aot: Sure. will do

thorwolpert (approved):

colin-api/src/colin_api/models/filing.py
```diff
@@ -1048,31 +1049,24 @@ def _process_correction(cls, cursor, business: Business, filing: Filing, corp_nu
                 name = change['newValue']['nameRequest'].get('legalName', None)
                 cls._create_corp_name(cursor, filing, corp_num, name)
 
-            elif f"/filing/{filing.body['correctedFilingType']}/nameTranslations" in change['path']:
-                old_translations = CorpName.get_by_event(
```
kialj876, line 1063: You will still need to do this check in case there was another filing after the corrected filing that changed the name (i.e. incorporation with name translations, then alteration that changes all name translations, then correction to the incorporation that changes the name translations would result in the company now having the name translations in both the alteration and the correction to incorporation). To avoid these complicated scenarios (that will be super rare) I was raising an error for manual intervention, but you could make it so that if the oldValue name has an end event id then you change that to the correction end event id (as you would currently be doing with this code) and then when you create the newValue name you also add the old end event id to it. That was a mouthful haha let me know if you have questions

kialj876 (changes requested):
Let me know if you want to chat after standup

colin-api/src/colin_api/models/filing.py
kialj876: I meant to select the code a few lines down where it raises the error

kialj876 (approved):
Nice work!

PR #1099, opened 2021-01-22
6198 include arMinDate and arMaxDate with the covid extension

vysakh-menon-aot:
*Issue #:* /bcgov/entity#6198

*Description of changes:*


legal-api/src/legal_api/resources/business/business_tasks.py
```diff
@@ -140,56 +135,44 @@ def construct_task_list(business):  # pylint: disable=too-many-locals; only 2 ex
...
-                    .get('annualGeneralMeetingDate', None)
-            if not last_ar_date:
-                last_ar_date = annual_report_filings[0].filing_date
-            last_ar_date = datetime.fromisoformat(last_ar_date)
-            if check_agm:
-                todo_start_date = (datetime(last_ar_date.year+1, 1, 1)).date()
-            else:
-                todo_start_date = (last_ar_date+datedelta.YEAR).date()
+            # Consider each filing as each year and add to find next ar year
+            next_ar_year += len(annual_report_filings)
```
kialj876, line 146: if a business was founded in 2018 and they had filed their 2019+2020 annual reports (so business.last_ar_year would equal 2020), then their next_ar_year = 2020 + 2 = 2022. Am I missing something or does this logic need to change?

legal-api/src/legal_api/services/filings/validations/annual_report.py
```diff
@@ -32,6 +31,28 @@ def requires_agm(business: Business) -> bool:
...
+def get_ar_dates(business: Business, start_date, next_ar_year):
+    """Get ar min and max date for the specific year."""
+    check_agm = requires_agm(business)
+    ar_min_date = datetime(next_ar_year, 1, 1).date()
+
+    # Make sure min date is greater than or equal to last_ar_date or founding_date
+    if ar_min_date < start_date:
+        ar_min_date = start_date
+
+    ar_max_date = datetime(next_ar_year, 12, 31).date()
```
kialj876, line 43: this ar_max_date will need to be decided differently for extensions and is what I was talking about in the RFC. I think we need to have a design discussion with Thor and Lekshmi on what we'll add for ARs to make it work for extensions

kialj876 (commented):
Hey Vysakh I think we should have a quick design session with thor and lekshmi before we move forward with this

legal-api/src/legal_api/resources/business/business_tasks.py
vysakh-menon-aot: `annual_report_filings` are the filings which are in not completed status, at that point `last_ar_year` will not have pending or draft filings (as per the functionality I dont see a possibility to have more than one in `annual_report_filings`). so in that case 2019 and 2020 will not be in `annual_report_filings`

kialj876: ah okay you are right. That method name 'get_filings_by_type' is a bit misleading (it was probably me way back when lol) as it doesn't return completed filings

kialj876 (approved):

lekshmimallika-aot (approved):

severinbeauvais (approved):

PR #1101, opened 2021-02-01
6256 Update staff comments

vysakh-menon-aot:
*Issue #:* /bcgov/entity#6256

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_comments.py
+++ b/legal-api/src/legal_api/resources/business/business_comments.py
@@ -35,13 +35,14 @@ from .api_namespace import API
 
 @cors_preflight('GET, POST')
 @API.route('/<string:identifier>/comments', methods=['GET', 'POST', 'OPTIONS'])
+@API.route('/<string:identifier>/comments/<int:comment_id>', methods=['GET', 'OPTIONS'])
 class BusinessCommentResource(Resource):
     """Business Comment service."""
 
     @staticmethod
     @cors.crossdomain(origin='*')
     @jwt.requires_auth
-    def get(identifier):
+    def get(identifier, comment_id=None):
         """Return a JSON object with meta information about the Service."""
         # basic checks
         business = Business.find_by_identifier(identifier)
@@ -50,6 +51,13 @@ class BusinessCommentResource(Resource):
             return jsonify(err_msg), err_code
 
         comments = db.session.query(Comment).filter(Comment.business_id == business.id, Comment.filing_id.is_(None))
+
+        if comment_id:
+            comment = comments.filter(Comment.id == comment_id).one_or_none()
+            if not comment:
+                return jsonify({'message': f'Comment {comment_id} not found'}), HTTPStatus.NOT_FOUND
+
+            return jsonify(comment.json)
 
         rv = []
         for comment in comments:
@@ -94,8 +102,7 @@ class BusinessCommentResource(Resource):
         except BusinessException as err:
             reply = json_input
             reply['errors'] = [{'error': err.error}, ]
-            return jsonify(reply), err.status_code or \
-                (HTTPStatus.CREATED if (request.method == 'POST') else HTTPStatus.ACCEPTED)
+            return jsonify(reply), err.status_code
 
         # all done
         return jsonify(comment.json), HTTPStatus.CREATED
--- a/legal-api/src/legal_api/resources/business/business_comments.py
+++ b/legal-api/src/legal_api/resources/business/business_comments.py
@@ -23,7 +23,7 @@ from flask import g, jsonify, request
 from flask_restplus import Resource, cors
 
 from legal_api.exceptions import BusinessException
-from legal_api.models import Business, Comment, Filing, User, db
+from legal_api.models import Business, Comment, User, db
 from legal_api.services import authorized
 from legal_api.services.comments import validate
 from legal_api.utils.auth import jwt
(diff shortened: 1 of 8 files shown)


legal-api/src/legal_api/resources/business/business_comments.py
```diff
@@ -0,0 +1,114 @@
...
+
+@cors_preflight('GET, POST')
+@API.route('/<string:identifier>/comments', methods=['GET', 'POST', 'OPTIONS'])
+class BusinessCommentResource(Resource):
+    """Business Comment service."""
+
+    @staticmethod
+    @cors.crossdomain(origin='*')
+    @jwt.requires_auth
+    def get(identifier):
```
lekshmimallika-aot, line 44: get by id is something that exists for all other objects. May have to check whether that should be supported

legal-api/src/legal_api/resources/business/business_comments.py
```diff
@@ -0,0 +1,114 @@
...
+            comment.comment = json_input['comment']['comment']
+            comment.staff_id = user.id
+            comment.business_id = business.id
+            comment.timestamp = datetime.datetime.utcnow()
+            comment.save()
+        except BusinessException as err:
+            reply = json_input
+            reply['errors'] = [{'error': err.error}, ]
+            return jsonify(reply), err.status_code or \
+                (HTTPStatus.CREATED if (request.method == 'POST') else HTTPStatus.ACCEPTED)
```
lekshmimallika-aot, line 98: If I understand correctly this gets called for a POST . Should this check be there

legal-api/src/legal_api/resources/business/business_comments.py
```diff
@@ -0,0 +1,114 @@
...
+        rv = []
+        for comment in comments:
+            rv.append(comment.json)
+
+        return jsonify(comments=rv)
+
+    @staticmethod
+    @cors.crossdomain(origin='*')
+    @jwt.requires_auth
+    def post(identifier):
```
lekshmimallika-aot, line 63: I have not gone through requirements. Is there a need to edit a comment? In that case PUT should be supported as well

legal-api/src/legal_api/resources/business/business_comments.py
vysakh-menon-aot: id of comments?

legal-api/src/legal_api/resources/business/business_comments.py
vysakh-menon-aot: no. they cannot edit

legal-api/src/legal_api/resources/business/business_comments.py
lekshmimallika-aot: ya

vysakh-menon-aot: as per the requirement this is not required. but if we follow that standard i can include it

legal-api/src/legal_api/resources/business/business_comments.py
vysakh-menon-aot: will fix it

legal-api/src/legal_api/resources/business/business_comments.py
lekshmimallika-aot: I think get all, get by id is more of a std for an endpoint

vysakh-menon-aot: will include that

lekshmimallika-aot (approved):

legal-api/src/legal_api/resources/business/business_comments.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # save comment
+        user = User.get_or_create_user_by_jwt(g.jwt_oidc_token_info)
+        try:
+            comment = Comment()
+            comment.comment = json_input['comment']['comment']
+            comment.staff_id = user.id
+            comment.business_id = business.id
+            comment.timestamp = datetime.datetime.utcnow()
+            comment.save()
```
severinbeauvais, line 101: Does this reuse the existing comments table?

Where is field `comment.submitterDisplayName`?

legal-api/src/legal_api/resources/business/business_comments.py
```diff
@@ -0,0 +1,121 @@
...
+    @jwt.requires_auth
+    def get(identifier, comment_id=None):
+        """Return a JSON object with meta information about the Service."""
+        # basic checks
+        business = Business.find_by_identifier(identifier)
+        err_msg, err_code = BusinessCommentResource._basic_checks(identifier, business, request)
+        if err_msg:
+            return jsonify(err_msg), err_code
+
+        comments = db.session.query(Comment).filter(Comment.business_id == business.id, Comment.filing_id.is_(None))
```
severinbeauvais, line 53: If we are reusing the existing comments table, does the result set need to be filtered by a comment type? Or is it enough that the comments we have have a business_id while the other have a filing_id?

legal-api/tests/unit/api/test_business_comments.py
```diff
@@ -0,0 +1,220 @@
...
+del SAMPLE_JSON_DATA['comment']['timestamp']
+del SAMPLE_JSON_DATA['comment']['submitterDisplayName']
+
+
+def test_get_all_business_comments_no_results(session, client, jwt):
+    """Assert that endpoint returns no-results correctly."""
+    identifier = 'CP7654321'
+    factory_business(identifier)
+
+    rv = client.get(f'/api/v1/businesses/{identifier}/comments',
```
severinbeauvais, line 44: 👍

legal-api/src/legal_api/resources/business/business_comments.py
vysakh-menon-aot: Yes its existing comments table. `submitterDisplayName` is `display_name` from `user` table, which will be in get response

legal-api/src/legal_api/resources/business/business_comments.py
vysakh-menon-aot: filing comments api is inserting only filing_id and the existing schema says one of the id required. so this should be enough to separate comments

severinbeauvais (approved):
Looks good.

severinbeauvais (approved):
Version updates look good.

PR #1105, opened 2021-02-10
fixed coop AR wrong filing year in output

severinbeauvais:
*Issue #:* /bcgov/entity#6443

*Description of changes:*

The AR output needs to use the "AR Filing Year" instead of the AR date, now that AGMs can be held in the year after.

### DIFF
--- a/legal-api/report-templates/annualReport.html
+++ b/legal-api/report-templates/annualReport.html
@@ -19,7 +19,7 @@
     <div class="column-report">
       <div class="column-report-headspace"></div>
       <div class="header-blue">
-        {{ annualReport.annualReportDate[:4] }} BC COOPERATIVE ANNUAL REPORT
+        {{ header.ARFilingYear }} BC COOPERATIVE ANNUAL REPORT
       </div>
       <div class="header-gold">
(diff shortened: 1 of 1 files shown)


severinbeauvais:
Before:

image

After:

image

legal-api/report-templates/annualReport.html
```diff
@@ -19,7 +19,7 @@
     <div class="column-report">
       <div class="column-report-headspace"></div>
       <div class="header-blue">
-        {{ annualReport.annualReportDate[:4] }} BC COOPERATIVE ANNUAL REPORT
+        {{ header.ARFilingYear }} BC COOPERATIVE ANNUAL REPORT
```
vysakh-menon-aot, line 22: does `ARFilingYear` is been set to header from UI

vysakh-menon-aot (approved):

cameron-freshworks (approved):

legal-api/report-templates/annualReport.html
severinbeauvais: I believe all header props are available. I tested (see below) to make sure :)

PR #1107, opened 2021-02-11
filer fix

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5109

*Description of changes:*
 - get name request from filing_json
 - avoid sentry from logging wrong info

### DIFF
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/name_request.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/name_request.py
@@ -28,24 +28,25 @@ def consume_nr(business: Business, filing: Filing, nr_num_path='/filing/incorpor
     """Update the nr to a consumed state."""
     try:
         nr_num = get_str(filing.filing_json, nr_num_path)
-        bootstrap = RegistrationBootstrap.find_by_identifier(filing.temp_reg)
-        namex_svc_url = current_app.config.get('NAMEX_API')
-        token = AccountService.get_bearer_token()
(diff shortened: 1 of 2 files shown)


queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/name_request.py
```diff
@@ -28,24 +28,25 @@ def consume_nr(business: Business, filing: Filing, nr_num_path='/filing/incorpor
     """Update the nr to a consumed state."""
     try:
         nr_num = get_str(filing.filing_json, nr_num_path)
-        bootstrap = RegistrationBootstrap.find_by_identifier(filing.temp_reg)
-        namex_svc_url = current_app.config.get('NAMEX_API')
-        token = AccountService.get_bearer_token()
+        if nr_num:
```
severinbeauvais, line 31: What does it mean if the NR Num is not found?

severinbeauvais: Under what conditions would this happen?

vysakh-menon-aot: for IA this function is called in both cases (numbered and named), so for numbered we have to skip this call

severinbeauvais: Thanks.
Please add code comment explaining this.

severinbeauvais (approved):

PR #1120, opened 2021-03-16
6416 - LEGAL API - Court Order and Plan of Arrangement

andrepestana-aot:
*Issue #:* /bcgov/entity#6416

*Description of changes:*

- Added columns to the model;
- Added migration;
- Added test;
- Added columns to lear_db.sql.

### DIFF
--- a/legal-api/tests/unit/models/test_filing.py
+++ b/legal-api/tests/unit/models/test_filing.py
@@ -25,8 +25,8 @@ import datedelta
 import pytest
 from flask import current_app
 from freezegun import freeze_time
-from legal_api.models import Business, Filing, User
-from legal_api.exceptions import BusinessException
+from legal_api.models import Business, Filing, User  # noqa: I001
+from legal_api.exceptions import BusinessException  # noqa: I001
 from registry_schemas.example_data import (
     ALTERATION_FILING_TEMPLATE,
     ANNUAL_REPORT,
--- a/legal-api/tests/unit/models/test_filing.py
+++ b/legal-api/tests/unit/models/test_filing.py
@@ -31,15 +31,21 @@ from registry_schemas.example_data import (
     CORRECTION_AR,
     FILING_HEADER,
     SPECIAL_RESOLUTION,
+    ALTERATION_FILING_TEMPLATE,
+    COURT_ORDER
 )
 from sqlalchemy_continuum import versioning_manager
 
 from legal_api.exceptions import BusinessException
-from legal_api.models import Filing, User
+from legal_api.models import Business, Filing, User
 from tests import EPOCH_DATETIME
 from tests.conftest import not_raises
-from tests.unit.models import factory_business, factory_completed_filing, factory_filing
-
+from tests.unit.models import (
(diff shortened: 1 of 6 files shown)


legal-api/tests/unit/models/test_filing.py
```diff
@@ -25,20 +25,27 @@
...
     ANNUAL_REPORT,
     CHANGE_OF_DIRECTORS,
     CORRECTION_AR,
+    COURT_ORDER,
     FILING_HEADER,
     SPECIAL_RESOLUTION,
 )
 from sqlalchemy_continuum import versioning_manager
 
-from legal_api.exceptions import BusinessException
```
thorwolpert, line 37: why were these moved?
the linter should complain about the move

legal-api/tests/unit/models/test_filing.py
```diff
@@ -672,3 +679,20 @@ def test_linked_not_correction(session):
     # test
     assert filing1.json['filing']['header']['isCorrected'] is False
     assert filing2.json['filing']['header']['affectedFilings'] is not None
+
+
+def test_alteration_filing_with_court_order(session):
```
thorwolpert, line 684: Can you add a test to ensure if any validators are necessary they work?

thorwolpert (commented):
is there no validation required other than schema correctness?

legal-api/tests/unit/models/test_filing.py
andrepestana-aot: I don't know why but my Flake8 local is slightly different from the GH's action. I moved because Flake8 complianed. How can I make it use the same rules?

legal-api/tests/unit/models/test_filing.py
andrepestana-aot: I don't know why but my Flake8 local is slightly different from the GH's action. I moved because Flake8 complianed. How can I make it use the same rules?

thorwolpert: are you on Windows+WSL or Mac/Linux? running ```make setup``` should get it all install correctly
if you are not, then make sure you do a ```pip install -e .``` in the services home directory, so it knows these modules are local and not external

andrepestana-aot: I'm using WSL 2. I used to run ```make setup``` but I also used to run ``` pip install .```. I see now that I cannot run pip after running make. I fixed the import. The strange is that the checks on github passed. Thanks!

andrepestana-aot:
> is there no validation required other than schema correctness?
I added another test to check saving invalid data. Is that what you mean?

andrepestana-aot:
I don't understand why codecov is failing. It was passing before the latest PR was merged.

legal-api/tests/unit/models/test_filing.py
andrepestana-aot: I added another test that checks what happened when trying to insert invalid data to the columns.

thorwolpert (approved):

PR #1130, opened 2021-03-30
Refactored Nr Query bug fix

cameron-freshworks:
*Issue #:* /bcgov/entity#6902

*Description of changes:*

* Just refactored bug fix to prevent Json conversion only where required.

### DIFF
--- a/legal-api/src/legal_api/services/namex.py
+++ b/legal-api/src/legal_api/services/namex.py
@@ -69,7 +69,7 @@ class NameXService():
             'Authorization': 'Bearer ' + token
         })
 
-        return nr_response.json()
+        return nr_response
 
     @staticmethod
     def update_nr(nr_json):
--- a/legal-api/src/legal_api/services/namex.py
+++ b/legal-api/src/legal_api/services/namex.py
@@ -69,6 +69,11 @@ class NameXService():
             'Authorization': 'Bearer ' + token
         })
 
+        # Return if not successful so downstream can handle the error
+        if nr_response.status_code != 200:
+            return nr_response
+
+        # Return the json if successful
         return nr_response.json()
 
     @staticmethod
--- a/legal-api/src/legal_api/services/namex.py
+++ b/legal-api/src/legal_api/services/namex.py
@@ -69,7 +69,7 @@ class NameXService():
             'Authorization': 'Bearer ' + token
         })
 
-        return nr_response
+        return nr_response.json()
 
     @staticmethod
     def update_nr(nr_json):
--- a/legal-api/src/legal_api/services/namex.py
+++ b/legal-api/src/legal_api/services/namex.py
@@ -69,12 +69,7 @@ class NameXService():
             'Authorization': 'Bearer ' + token
         })
 
-        # Return if not successful so downstream can handle the error
-        if nr_response.status_code != 200:
-            return nr_response
-
-        # Return the json if successful
-        return nr_response.json()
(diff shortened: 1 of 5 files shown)


thorwolpert (commented):
Please add/update the the tests for this.

cameron-freshworks:
> Please add/update the the tests for this.

Absolutely, will do. First thing in the am :)

cameron-freshworks:
> Please add/update the the tests for this.

Added a couple unit tests for the method changed. Not too sure what else to write tests for to appease the codecov 😬 
Open to suggestions though!

legal-api/src/legal_api/services/namex.py
```diff
@@ -69,6 +69,11 @@ def query_nr_number(identifier: str):
             'Authorization': 'Bearer ' + token
         })
 
+        # Return if not successful so downstream can handle the error
+        if nr_response.status_code != 200:
+            return nr_response
+
+        # Return the json if successful
         return nr_response.json()
```
argush3, line 77: I think converting this to json is the reason the name request GET endpoint is erroring out.  In looking at the legal-api logs, it looks like an error is being thrown from `namerequest.py` on the `if nr_response.status_code == 404:` line.  My guess is that since the `nr_response` object is being converted to json here, the get function in `namerequest.py` triggers the error we are seeing for retrieving NRs in DEV right now.

```2021-03-29 21:10:14,050 - legal_api - ERROR in namerequest:namerequest.py:46 - get: 'dict' object has no attribute 'status_code'
2021-03-29 21:10:14,058 - legal_api - ERROR in app:app.py:1891 - log_exception: Exception on /api/v1/nameRequests/NR 1637549 [GET]
```

cameron-freshworks: You bet! 
This pr is to address that error, as it was converting it later as well in the query_nr_number method.
So i updated that methods error handling and removed the second conversion, since we needed it to be json in a few places it made sense to apply it at the source.

cameron-freshworks: Line 44 in `namerequest.py` is the fix for this.

cameron-freshworks: Definitely open to any suggestions on solutions tho!

argush3, line 77: I think converting this to json is problematic because of the other places calling `query_nr_number`.  I noticed you had updated `namerequest.py` but I do see a bunch of other places where `query_nr_number` is called.  Potentially this may be an issue I'm thinking.

argush3 (commented):
Just took a quick look and added my thoughts.  Ping me if you want to talk directly.

legal-api/src/legal_api/services/namex.py
cameron-freshworks: Ah ok. I see two instances where it might call `.json()` before validations. 
So your suggesting i remove those two instances since we are already doing the conversion?

argush3: As per our conversation in RC, I think having `query_nr_number` return an object as opposed to json would be more ideal.  The consumer of `query_nr_number` can decide how in what format it wants to return the object if that is required.

cameron-freshworks: Sounds good, refactoring! 👍

argush3 (approved):
Looks good.

PR #1135, opened 2021-04-07
added technical json to filings

thorwolpert:
*Issue #:* /bcgov/entity#6618

*Description of changes:*
- added technical json feature to model & core
- fixed lint issue in reports

### DIFF
--- a/legal-api/src/legal_api/core/filing.py
+++ b/legal-api/src/legal_api/core/filing.py
@@ -15,8 +15,9 @@
 from __future__ import annotations
 
 # from dataclasses import dataclass, field
+import copy
 from enum import Enum
-from typing import Dict, Optional
+from typing import Dict, List, Optional
 
 from legal_api.core.utils import diff_dict, diff_list
 from legal_api.models import Business, Filing as FilingStorage  # noqa: I001
@@ -77,7 +78,7 @@ class Filing:
         self._payment_token: str
         self._payment_completion_date: datetime
         self._status: Optional[str] = None
-        self._paper_only: bool
+        self._paper_only: bool = False
         self._payment_account: Optional[str] = None
 
     @property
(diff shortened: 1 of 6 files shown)


thorwolpert:
skipping SonarCloud code smell as its duplicated text in a test

legal-api/src/legal_api/core/filing.py
```diff
@@ -246,3 +250,21 @@ def get_filings_by_status(business_id: int, status: [], after_date: date = None)
...
+
+        Returns: {
+            List: or None of the Legal Filing JSON segments.
+            }
+        """
+        if not (filing := self.json):
+            return None
+
+        legal_filings = []
+        for k in filing['filing'].keys():  # pylint: disable=unsubscriptable-object
```
jordiwes, line 265: is k an id, and if so, can we name it so?

legal-api/src/legal_api/models/filing.py
```diff
@@ -127,12 +127,41 @@ class Source(Enum):
     }
 
     __tablename__ = 'filings'
+    # this mapper is used so that new and old versions of the service can be run simultaneously,
+    # making rolling upgrades easier
+    __mapper_args__ = {
```
jordiwes, line 132: Where is it used?

legal-api/src/legal_api/core/filing.py
thorwolpert: the code is moved over from the model.Filings, and k is the key from keys().

legal-api/src/legal_api/models/filing.py
thorwolpert: it's used by SQLAlchemy as part of the framework.
i'll expand on the comments here.

jordiwes (approved):
Great, thanks LGTM

PR #1136, opened 2021-04-07
7038 - Saving an alteration as staff resulting in 500 error (court order related)

andrepestana-aot:
*Issue #:* /bcgov/entity#7038

*Description of changes:*

- Fixed validator bug and added tests for the validator

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/common_validations.py
+++ b/legal-api/src/legal_api/services/filings/validations/common_validations.py
@@ -107,17 +107,14 @@ def validate_court_order(court_order_path, court_order):
             msg.append({'error': 'Length of court order file number must be from 5 to 20 characters.',
                         'path': err_path})
 
-    if 'effectOfOrder' not in court_order and (len(court_order['effectOfOrder']) < 5 or
-                                               len(court_order['effectOfOrder']) > 500):
+    if 'effectOfOrder' in court_order and (len(court_order['effectOfOrder']) < 5 or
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/services/filings/validations/common_validations.py
```diff
@@ -107,17 +107,14 @@ def validate_court_order(court_order_path, court_order):
...
         err_path = court_order_path + '/effectOfOrder'
         msg.append({'error': 'Length of court order effect of order must be from 5 to 500 characters.',
                     'path': err_path})
 
     court_order_date_path = court_order_path + '/orderDate'
-    if 'orderDate'not in court_order:
-        err_path = court_order_date_path
-        msg.append({'error': 'Court order date is required.', 'path': err_path})
-    else:
+    if 'orderDate' in court_order:
```
cameron-freshworks, line 117: In the current state, this condition will never be met just an fyi. 
Current designs have no means of capturing this specific date. 
If it's a hard requirement that needs to be met, we would need to take this to design.

cameron-freshworks (approved):
Looks good! 
Just one comment.

legal-api/src/legal_api/services/filings/validations/common_validations.py
andrepestana-aot: Yes, but if there's a date there I think it should be validated.

thorwolpert (approved):

PR #1137, opened 2021-04-07
4716 colin only flag

jordiwes:
*Issue #:* /bcgov/entity#4716

*Description of changes:*
- add colin_only flag
- check and provide response in api


legal-api/tests/unit/services/test_document_meta.py
```diff
@@ -101,6 +103,30 @@ def test_available_on_paper_only(session, app):
...
+                },
+                'business': {
+                    'identifier': 'BC1234567'
+                }
+            }
+        }
+        assert len(document_meta.get_documents(filing)) == 0
+
+
+def test_available_in_colin_only(session, app):
```
andrepestana-aot, line 117: I didn't understand the purpose of this test as availableInColinOnly doesn't matter for the DocumentMetaService and nothing related to the flag is asserted.

legal-api/migrations/versions/9741616ab6d7_.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('filings', sa.Column('colin_only', sa.Boolean(), nullable=True))
+    
+
+def downgrade():
+    op.drop_column('filings', 'colin_only')
```
andrepestana-aot, line 24: Instead of removing and adding, wouldn't it be better to alter column? That way you're going to lose the data already inserted.
https://alembic.sqlalchemy.org/en/latest/ops.html

andrepestana-aot (changes requested):

legal-api/tests/unit/services/test_document_meta.py
jordiwes: it is a test that verifies that no documents are returned when the filing is in colin only

legal-api/migrations/versions/9741616ab6d7_.py
jordiwes: The upgrade will add the column (the column is a new column).
The downgrade will remove the column

andrepestana-aot: Ok. I see. Thanks.

andrepestana-aot (changes requested):
Please assign this to yourself and please add a test to check the flag is being saved to DB and retrieved as the tests we have in test_filing.py

thorwolpert (commented):
Is there a PR to business-schema to add **availableInColinOnly** to the filing header?

andrepestana-aot (approved):

legal-api/requirements.txt
```diff
@@ -8,47 +8,94 @@ Flask==1.1.2
...
+pycodestyle==2.7.0
 pycountry==20.7.3
+pycparser==2.20
+pydocstyle==3.0.0
+pyflakes==2.3.1
+pylint-plugin-utils==0.6
+pylint==2.7.4
+pyparsing==2.4.7
 pyrsistent==0.17.3
+pytest-asyncio==0.14.0
```
thorwolpert, line 77: looks like you included all of the modules in requirements/dev.txt into requirements.txt
Please remove the dev/test related ones.

thorwolpert (commented):
looks good, just the requirements.txt file needs to contain only the prod/bcreg packages.

thorwolpert (approved):

PR #1140, opened 2021-04-12
Fixed logic to display Name Translations in Alteration Notice

andrepestana-aot:
*Issue #:* /bcgov/entity#7006

*Description of changes:*

- Fixed logic to display Name Translations in Alteration Notice

### DIFF
--- a/legal-api/report-templates/template-parts/common/nameTranslation.html
+++ b/legal-api/report-templates/template-parts/common/nameTranslation.html
@@ -1,5 +1,5 @@
 <div class="no-page-break">
-    {% if listOfTranslations|length == 0 %}
+    {% if listOfTranslations|length or nameTranslations|length %}
         <div class="separator mt-4"></div>
         <div class="section-title mt-4">Company Name Translation(s)</div>
         <div class="section-data mt-3">
--- a/legal-api/report-templates/template-parts/common/nameTranslation.html
+++ b/legal-api/report-templates/template-parts/common/nameTranslation.html
@@ -1,10 +1,10 @@
 <div class="no-page-break">
-    {% if listOfTranslations|length or nameTranslations|length %}
+    {% if listOfTranslations|length or previousNameTranslations|length %}
         <div class="separator mt-4"></div>
         <div class="section-title mt-4">Company Name Translation(s)</div>
         <div class="section-data mt-3">
             <table class="section-translation-table" role="presentation">
-                {% if listOfTranslations|length == 0 and nameTranslations| length %}
+                {% if listOfTranslations|length == 0 and previousNameTranslations|length %}
                     <div class="mt-2">NONE</div>
                 {% else %}
                     {% for translation in listOfTranslations %}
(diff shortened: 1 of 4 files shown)


legal-api/report-templates/template-parts/common/nameTranslation.html
```diff
@@ -1,5 +1,5 @@
 <div class="no-page-break">
-    {% if listOfTranslations|length == 0 %}
+    {% if listOfTranslations|length or nameTranslations|length %}
```
severinbeauvais, line 2: Is there a unit test to verify this fix?

I'm actually a bit confused on why there are 2 lists here. How does it work with the logic on line 7? What are the various scenarios? Can any logic be simplified (or commented) to reduce confusion?

Also, are there any version numbers that need to be updated?

andrepestana-aot: It's just a fix for changes that Arturo did to the reports. 

There's no tests but I can add one. The variables are confusing. I'm going to change that as well. Thanks.

andrepestana-aot: I added tests. I think it's easier to understand now. What do you think?

severinbeauvais: Awesome! Yes, way better.

I was trying to understand the logic on line 7. I think it's still a bit weird but the unit tests do an excellent job of covering all the cases, so it's fine.

severinbeauvais (approved):
Looks good.

legal-api/tests/unit/reports/templates/test_name_translations.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_no_rendering(session):
+    """Test Company Name Translation(s) section should not rendered when no current translation nor previous ones."""
+    template = get_template()
+    rendered = template.render(listOfTranslations=[], previousNameTranslations=[])
+    assert title not in rendered
+
+
+def test_render_none(session):
```
thorwolpert, line 40: Could these be collapsed to a parameterized test?

thorwolpert (commented):
just a question on collapsing the tests to a table driven test.

legal-api/tests/unit/reports/templates/test_name_translations.py
andrepestana-aot: They could but I don't think we should do that for the sake of simplicity and readability. If you really want me to do that I can but I don't like the way become being parameterized.

jachurchill (approved):

severinbeauvais:
@thorwolpert Please merge if your conversation is resolved to your satisfaction.

legal-api/tests/unit/reports/templates/test_name_translations.py
thorwolpert: i think it would make it easier for the next person to add if needed, as they look pretty similar to me.
you can leave it and the next person can do that ...

thorwolpert (approved):

PR #1144, opened 2021-04-14
7101 - Legal API - prevent LTDs and ULCs from deleting draft alteration

andrepestana-aot:
*Issue #:* /bcgov/entity#7101

*Description of changes:*

- Added check for limited and unlimited companies; 
- Added tests

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -235,6 +235,13 @@ class ListFilingResource(Resource):
         if not filing:
             return jsonify({'message': _('Filing Not Found.')}), HTTPStatus.NOT_FOUND
 
+        business = Business.find_by_identifier(identifier)
+        if (filing.status == Filing.Status.DRAFT.value and
+                business.legal_type in [lt.value for lt in Business.LIMITED_COMPANIES + Business.UNLIMITED_COMPANIES]):
+            return jsonify({
+                            'message': _('You must complete this alteration filing to become a BC Benefit Company.')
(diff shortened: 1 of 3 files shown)


kialj876 (dismissed):

legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -235,6 +235,13 @@ def delete(identifier, filing_id=None):
         if not filing:
             return jsonify({'message': _('Filing Not Found.')}), HTTPStatus.NOT_FOUND
 
+        business = Business.find_by_identifier(identifier)
+        if (filing.status == Filing.Status.DRAFT.value and
```
thorwolpert, line 239: This blocks them from deleting any draft filing.
Once these corp_types get added, it'll block them and behave differently than coops & BEN.
It should also check the filing type is an alteration.
It also blocks them from deleting a draft AR, CN,COD, etc.

thorwolpert (changes requested):
the original discussion was to alter filings so that it could be marked as blocked.
this alters delete based on that these corp types will be blocked from deleting a draft which is not right

legal-api/src/legal_api/resources/business/business_filings.py
kialj876: Good point Thor. Also, why are we preventing Alteration draft deletion in the first place? I can see this leading to a very bad ops day for someone where they have to delete everything manually a year from now. Shouldn't this always be an option for staff at least?

kialj876: ah just read your comment above about it being set to blocked

PR #1150, opened 2021-04-19
Legal API - prevent LTDs and ULCs from deleting draft alteration

andrepestana-aot:
*Issue #:* /bcgov/entity#7101

*Description of changes:*

- Added deletion_locked to filings table; 
- Added logic to avoid deletion; 
- Added tests.

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -236,19 +236,7 @@ class ListFilingResource(Resource):
             return jsonify({'message': _('Filing Not Found.')}), HTTPStatus.NOT_FOUND
 
         if filing.deletion_locked:  # should not be deleted
-            business = Business.find_by_identifier(identifier)
-            if (filing.status == Filing.Status.DRAFT.value and
-                    filing.filing_type == 'alteration' and
-                    business.legal_type in [lt.value for lt in (Business.LIMITED_COMPANIES +
-                                                                Business.UNLIMITED_COMPANIES)]):
-                response = jsonify({
-                                'message': _('You must complete this alteration filing to become a BC Benefit Company.')
-                                }), HTTPStatus.UNAUTHORIZED
-            else:
-                response = jsonify({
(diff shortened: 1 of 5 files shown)


legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -235,6 +235,9 @@ def delete(identifier, filing_id=None):
         if not filing:
             return jsonify({'message': _('Filing Not Found.')}), HTTPStatus.NOT_FOUND
 
+        if filing.deletion_locked:  # should not be deleted
```
kialj876, line 238: Is deletion_locked as a new column necessary? If alteration/draft is not the only situation where we block deletion I would vote to add it to the filing as a property similar to 'locked' where we can define the business logic for this in one place. Otherwise if this is the only place we block deletion why not just have the 'if alteration/draft' check here since users will be able to delete this filing in the future anyways (once all bc companies are moved over to this system)?

andrepestana-aot: @kialj876 my first idea was blocking based on the filing properties (draft alteration for specific corp_types) but as @thor mentioned [here](https://github.com/bcgov/entity/issues/7101#issuecomment-819788501) and also discussed with @lmcclung  we also need to mark the filing as "blocked for deletion".

kialj876: Ah I see it would run into complications because the filing doesn't know what corp type it is. This looks good to me if we need to block other filing deletions for scenarios in the future, but if this is the only case and we're not planning for other ones then I still think this should just be `if alteration/draft/corp type is whatever, then block deletion`. That keeps it simple and less to remove after we have all bc companies filing through here (they will be able to delete this when that happens I think). Thoughts?

thorwolpert: @kialj876 that discussion is covered in the feature spec. There are other occasions where a filing stub will not be allowed to be deleted, so this becomes more general purpose.

thorwolpert (approved):

PR #1154, opened 2021-04-23
4711 colin api updates for data loader

argush3:
*Issue #:* /bcgov/entity#4711

*Description of changes:*
 

* Addition of endpoint to retrieve all events specific to a business.  This was specifically created for the data loader so is able to retrieve the event associated with an incorporation application or conversion ledger
* Updated event `/event/<string:corp_type>/<string:event_id>` endpoint to support other corp types.  The way it was implemented previously did not really allow query events for any companies where the corp num did not have an alphabetical prefix
* `FILING_TYPES` in `colin-api/src/colin_api/models/filing.py` was updated to include type codes for Colin specific filing type codes.  This is used by data loader to determine the generic application filing type.  e.g. `incorporationApplication` or `conversion`
* Addition of type codes and values for ULC and CCC
* Updated flask-restx to 0.3.0 to resolve dependency conflict

### DIFF
--- a/colin-api/src/colin_api/models/filing.py
+++ b/colin-api/src/colin_api/models/filing.py
@@ -53,27 +53,48 @@ class Filing:
         'annualReport': {
             'type_code_list': ['OTANN', 'ANNBC'],
             Business.TypeCodes.COOP.value: 'OTANN',
-            Business.TypeCodes.BCOMP.value: 'ANNBC'
+            Business.TypeCodes.BCOMP.value: 'ANNBC',
+            Business.TypeCodes.BC_COMP.value: 'ANNBC',
+            Business.TypeCodes.ULC_COMP.value: 'ANNBC',
+            Business.TypeCodes.CCC_COMP.value: 'ANNBC',
         },
         'changeOfDirectors': {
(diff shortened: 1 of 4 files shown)


colin-api/src/colin_api/models/filing.py
```diff
@@ -53,27 +53,48 @@ class LearSource(Enum):
...
+        },
+        'conversion': {
+            'type_code_list': ['CONVL'],
+            Business.TypeCodes.BC_COMP.value: 'CONVL',
+            Business.TypeCodes.ULC_COMP.value: 'CONVL',
+            Business.TypeCodes.CCC_COMP.value: 'CONVL'
         },
         'alteration': {
-            'type_code_list': ['NOALE', 'NOALR'],
+            'type_code_list': ['NOALA', 'NOALR'],
```
argush3, line 93: CDEV is using NOALA so hopefully changing it to this is ok?

thorwolpert: I'm going to assume the codes in CDEV are correct, so let's put it in and test across the CTST/SPRD releases.

thorwolpert (approved):

PR #1155, opened 2021-04-23
4711 data loader updates to support loading of LTDs, ULCs and CCCs

argush3:
*Issue #:* /bcgov/entity#4711

This is an update to data loader to support loading of LTDs, ULCs and CCCs.  The general change from the previous approach is to create business from an incorporation application filing or a conversion ledger filing as opposed to creating business stub from the current end state of the business in Colin and adding a lear epoch filing type as a marker.  With this approach, the incorporation application or conversion ledger filing will be retrieved from Colin and saved directly to the Lear database with some modifications where required.  This is followed by pushing the newly saved filing to the queue at which point the entity filer service with apply the incorporation application or conversion ledger filing.

*Description of changes:*

* restructure `load_corps` function in `lear-db/test_data/data_loader.py` to support the functionality laid out above.  Of particular note is that data loader has two new service calls:
  * a call to a newly created event endpoint in colin api to retrieve events for a given business
  * a call to publish the newly created incorporation application or conversion ledger filing to the entity filer queue
*  Update `requirements.txt` to resolve dependency conflicts


thorwolpert (commented):
Please review the changes.
The steps here will get converted to a notebook.

argush3:
> Please review the changes.
> The steps here will get converted to a notebook.

Reviewed and changes make sense to me.  The removal of the transaction was the only thing I had questions about.  Is that just not required in this instance?

FYI, I've pushed a new commit to fix some run time issues.

kialj876 (approved):
Very nice work

kialj876:
> > Please review the changes.
> > The steps here will get converted to a notebook.
> 
> Reviewed and changes make sense to me. The removal of the transaction was the only thing I had questions about. Is that just not required in this instance?
> 
> FYI, I've pushed a new commit to fix some run time issues.

I think it's because the created filing will be applied normally by the legal-api (so a transaction will be created for it when it is applied to the database). Is that correct Thor?

PR #1158, opened 2021-04-29
4711 legal api updates for data loader

argush3:
*Issue #:* /bcgov/entity#4711

*Description of changes:*

* Added `conversion` to `FILINGS` which is used by other lear components.  Specifically entity filer was updated to use `conversion`
* Updated filings save logic such that historical colin filings are not validated before saving
* Updated filings save logic such for colin filngs so that the filing is pushed to the entity filer queue in the following two scenarios:
  * business has epoch lear filing and the currently colin filing being saved is after the epoch lear filing date
  * business has no epoch lear filing 
* Updated flask-restx dependency conflict

### DIFF
--- a/legal-api/src/legal_api/resources/business/business_filings.py
+++ b/legal-api/src/legal_api/resources/business/business_filings.py
@@ -27,6 +27,7 @@ from flask_restx import Resource, cors
 from werkzeug.local import LocalProxy
 
 import legal_api.reports
+from legal_api.constants import BOB_DATE
 from legal_api.core import Filing as CoreFiling
 from legal_api.exceptions import BusinessException
 from legal_api.models import Address, Business, Filing, RegistrationBootstrap, User, db
@@ -172,9 +173,9 @@ class ListFilingResource(Resource):
         # get header params
         payment_account_id = request.headers.get('accountId', None)
 
-        # validate filing
-        if not draft and not ListFilingResource._is_before_epoch_filing(json_input,
-                                                                        Business.find_by_identifier(identifier)):
+        if not draft \
+                and not ListFilingResource._is_historical_colin_filing(json_input) \
+                and not ListFilingResource._is_before_epoch_filing(json_input, Business.find_by_identifier(identifier)):
             if identifier.startswith('T'):
                 business_validate = RegistrationBootstrap.find_by_identifier(identifier)
             else:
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -392,19 +392,32 @@ def _is_before_epoch_filing(filing_json: str, business: Business):
...
                 raise KeyError
-            if not ListFilingResource._is_before_epoch_filing(filing.filing_json, business):
-                payload = {'filing': {'id': filing.id}}
-                queue.publish_json(payload)
-            else:
+
+            epoch_filing = Filing.get_filings_by_status(business_id=business.id, status=[Filing.Status.EPOCH.value])
+
+            if epoch_filing and ListFilingResource._is_before_epoch_filing(filing.filing_json, business):
                 epoch_filing = Filing.get_filings_by_status(business_id=business.id, status=[Filing.Status.EPOCH.value])
```
argush3, line 413: remove this line

legal-api/src/legal_api/models/filing.py
```diff
@@ -112,6 +112,13 @@ class Source(Enum):
                 'BEN': 'BCINC'
             }
         },
+        'conversion': {
+            'name': 'conversion',
+            'title': 'Conversion Ledger',
+            'codes': {
+                'BEN': 'BCINC'
```
thorwolpert, line 119: Shouldn't this be for the existing types? There should be no conversion ledgers for BEN.

legal-api/src/legal_api/resources/business/business_filings.py
```diff
@@ -392,19 +392,32 @@ def _is_before_epoch_filing(filing_json: str, business: Business):
             filing_json['filing']['header']['date']).replace(tzinfo=datetime.timezone.utc)
         return filing_date < epoch_filing[0].filing_date
 
+    @staticmethod
+    def _is_historical_colin_filing(filing_json: str):
+        if filing_json['filing']['header']['source'] == 'COLIN' \
+                and filing_json['filing']['header']['date'] < '2019-03-08':
```
thorwolpert, line 398: Can this date be set somewhere?

legal-api/src/legal_api/resources/business/business_filings.py
thorwolpert, line 413: Instead of creating this twice, can you include it with a walrus :=  in the if block?

thorwolpert (changes requested):
some small questions.

legal-api/src/legal_api/models/filing.py
argush3: Looking into this a little closer and I realize this `FILINGS` dictionary is used as a mechanism to retrieve fee codes which is eventually passed to the payment api.  I'll remove this.  I must have been a little careless with copy/pasting from IA code above.

argush3: removed conversion ledger entry from `FILINGS` dictionary as per previous comment.

legal-api/src/legal_api/resources/business/business_filings.py
argush3: Created `constants.py` file and created a `BOB_DATE` constant which is now referenced in this method.

legal-api/src/legal_api/resources/business/business_filings.py
argush3: Good call.  Have updated as per your suggestion.

thorwolpert (approved):

thorwolpert (approved):
There are some broken tests, but I'll submit a PR to clean that.
The changes here are good to go.

PR #1160, opened 2021-04-29
4711 legal updater job updates for data loader

argush3:
*Issue #:* /bcgov/entity#4711

*Description of changes:*

* dependency conflict updates to package versions
* updated to support pulling ltd, ccc and ulc legal type events/filings from colin and pushing to the legal api where the business exists in lear database and the filing doesn't already exist.


argush3:
Flake8 fails on blind except.  This is from previous code.  Do we want to deal with this or ignore?

```
update_legal_filings.py:109:1: B902 blind except Exception: statement
update_legal_filings.py:249:1: B902 blind except Exception: statement
update_legal_filings.py:318:1: B902 blind except Exception: statement
```

kialj876 (approved):
Nice work. This looks good to me, but we'll need to double check everything still works once all the loader/updater changes are in dev

thorwolpert (approved):

PR #1164, opened 2021-05-10
4711 Add conversion filing to FILINGS dictionary in filing model

argush3:
*Issue #:* /bcgov/entity#4711

*Description of changes:*

* Add `conversion` filing to FILINGS dictionary in `legal-api/src/legal_api/models/filing.py`

### DIFF
--- a/legal-api/src/legal_api/models/filing.py
+++ b/legal-api/src/legal_api/models/filing.py
@@ -112,6 +112,10 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
                 'BEN': 'BCINC'
             }
         },
+        'conversion': {
+            'name': 'conversion',
+            'title': 'Conversion Ledger'
+        },
         'specialResolution': {'name': 'specialResolution', 'title': 'Special Resolution',
                               'codes': {
                                   'CP': 'RES'}},


legal-api/src/legal_api/models/filing.py
```diff
@@ -112,6 +112,10 @@ class Source(Enum):
                 'BEN': 'BCINC'
             }
         },
+        'conversion': {
+            'name': 'conversion',
+            'title': 'Conversion Ledger'
+        },
```
argush3, line 118: Removed this thinking that it was not required and used only for retrieving fee codes used for passing to the payment api during the invoice creation process.  After testing out the latest code in the main branch, conversion filings are not being processed by the `entity-filer` queue worker process due to this piece of missing code.  Without this code, the conversion filing processing logic is never triggered. 

Note that no fee codes(`codes` property) have been added as the colin conversion filing should not trigger the create invoice logic nor should it require any payment.

argush3:
@thorwolpert small PR that needs your review.  This will need to make it into main before we go to prod for conversion filings to be processed properly by the `entity-filer`.  Thanks!

kialj876 (approved):

thorwolpert (approved):

PR #1165, opened 2021-05-11
4711 bug fix filings endpoint for NOALA filing type

argush3:
*Issue #:* /bcgov/entity#4711

*Description of changes:

* Added support for `NOALA` filing type code for a GET on filings endpoint.  This was preventing the legal updater job from pulling over this type of filing.

### DIFF
--- a/colin-api/src/colin_api/models/filing.py
+++ b/colin-api/src/colin_api/models/filing.py
@@ -683,6 +683,8 @@ class Filing:
                     filing.body['business']['legalType'] = Business.TypeCodes.BC_COMP.value
                 elif filing_event_info['filing_type_code'] == 'NOALE':
                     filing.body['business']['legalType'] = Business.TypeCodes.BCOMP.value
+                elif filing_event_info['filing_type_code'] == 'NOALA':
+                    filing.body['business']['legalType'] = filing.business.corp_type
                 else:
                     raise InvalidFilingTypeException(filing_type=filing_event_info['filing_type_code'])
                 filing.body['business']['identifier'] = f'BC{filing.business.corp_num}'


colin-api/src/colin_api/models/filing.py
```diff
@@ -683,6 +683,8 @@ def get_filing(cls, filing: Filing, con=None, year: int = None) -> Dict:
                     filing.body['business']['legalType'] = Business.TypeCodes.BC_COMP.value
                 elif filing_event_info['filing_type_code'] == 'NOALE':
                     filing.body['business']['legalType'] = Business.TypeCodes.BCOMP.value
+                elif filing_event_info['filing_type_code'] == 'NOALA':
+                    filing.body['business']['legalType'] = filing.business.corp_type
```
kialj876, line 687: does this mean a filing with a NOALA isn't changing the legal type? This will set it to the current business legal type of the business. The output will look okay unless a company has multiple alterations in their history or something like that (and therefore the legal type would be incorrect and silently update lear incorrectly as well if the legal_updater was getting this filing). This is probably very rare but still possible

kialj876 (commented):
Might need to change this logic for finding what legal type to put in here. Not sure if we can check the corp_state table or if there are older rows in the corporations table we can check that correspond with the filing event_id

colin-api/src/colin_api/models/filing.py
argush3: After talking with @kialj876 about this and looking into the colin database structure, we are unable to determine what the legal type is at the time of a NOALA filing.  As per Kial's previous comment, this means there are scenarios where if a business has changed business types multiple times, we would not be able to pull over the right filing with the right legal type.

Currently, when the legal updater tries to pull over a NOALA filing it fails when trying to access the legal type.  Should we just keep letting it fail?  Or is there some way we can find the legal type of a business at the time of a particular NOALA filing?

thorwolpert: that sounds wrong. I did reach out to the DBA team and they confirm we always know what the corp_type is.

argush3:
@thorwolpert  @kialj876 Just emailed David Roberts about an hour ago to confirm how to get the corp type as i couldn’t see how you can get the corp type from the transaction and event…probably a gap in my understanding of the db. once i get confirmation of details i’ll create a new PR against main referencing this PR for historical purposes and implement a fix.

argush3:
Closing as this PR will be replaced with https://github.com/bcgov/lear/pull/1193

PR #1184, opened 2021-05-27
7514 - Fixed corrected label and datetime format

andrepestana-aot:
*Issue #:* /bcgov/entity#7514

*Description of changes:*

- Fixed diff_dict method; 
- Fixed date and time format.
- Added test

### DIFF
--- a/legal-api/src/legal_api/core/utils.py
+++ b/legal-api/src/legal_api/core/utils.py
@@ -49,7 +49,7 @@ def diff_dict(json1,
         if ignore_keys and key in ignore_keys:
             continue
 
-        if not json2.get(key):
+        if json2.get(key) is None and value is not None:
             diff.append(Node(old_value=None,
                              new_value=value,
                              path=path + [key]))
--- a/legal-api/tests/unit/core/test_filing_diff.py
+++ b/legal-api/tests/unit/core/test_filing_diff.py
@@ -16,6 +16,7 @@ import copy
 from typing import Final
 
 import datedelta
+import pytest
 
 from legal_api.core import Filing
 from legal_api.utils.datetime import datetime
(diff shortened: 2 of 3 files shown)


argush3 (approved):
LGTM 👍

legal-api/src/legal_api/core/utils.py
```diff
@@ -49,7 +49,7 @@ def diff_dict(json1,
         if ignore_keys and key in ignore_keys:
             continue
 
-        if not json2.get(key):
+        if json2.get(key) is None and value is not None:
```
thorwolpert, line 52: i don't think this is correct, as this is now added to the diff ... and it is NOT different

thorwolpert (changes requested):
it's now adding non-changed values to the diff, which is not correct.

legal-api/src/legal_api/core/utils.py
andrepestana-aot: The previous condition `not json2.get(key)` evaluated to True when `json2.get(key)` was `False` what caused some non-changed values to the diff with wrong oldValue:
image



Also, when the `json2.get(key)` was `None` the condition evaluated to True and add other non-changed values to the diff:
image

andrepestana-aot: Please check the tests.

legal-api/tests/unit/core/test_filing_diff.py
```diff
@@ -78,7 +82,35 @@ def test_filing_json_diff():
...
+        'path': RESOLUTION_PATH}]
+
+
+@pytest.mark.parametrize('test_name, value', [
+    ('None values', None),
+    ('False values', False),
+    ('True values', True)
+])
+def test_filing_json_diff_with_same_values(test_name, value):
+    """Assert same values are not added to diff."""
```
andrepestana-aot, line 94: I added some tests to verify the problem.

legal-api/tests/unit/core/test_filing_diff.py
```diff
@@ -78,7 +82,35 @@ def test_filing_json_diff():
     assert ld == [{
         'newValue': 'Be it resolved, and now it is.',
         'oldValue': 'Be it resolved, that it is resolved to be resolved.',
-        'path': '/filing/specialResolution/resolution'}]
+        'path': RESOLUTION_PATH}]
+
+
+@pytest.mark.parametrize('test_name, value', [
```
thorwolpert, line 88: could add these to the test_utils section, and increase those tests, rather than introduce general tests here?

andrepestana-aot: Ok. Done.

legal-api/src/legal_api/core/utils.py
thorwolpert: thanks for combining the test, makes it clearer

thorwolpert (approved):

PR #1188, opened 2021-05-28
7677 Fix issue with orderDetails being added to comments table

argush3:
*Issue #:* /bcgov/entity#7677

*Description of changes:* 

* Add `order_details` column to `filings` table
* Added `order_details` to `filings` model
* Update entity filer for registrar's notation, registrar's order and court order filing such that orderDetails value is now saved to new `filings.order_details` column when filing is processed. 
* Removed code saving orderDetails value to `comment` table for the 3 filing types
* Update entity filer processor unit tests

image

sample payload of what is returned by GET `/api/v1/businesses/BC0870342/filings`

```
{
  "filings": [
    {
      "filing": {
        "business": {
          "foundingDate": "2021-02-10T19:51:18.758421+00:00",
          "identifier": "BC0870342",
          "legalName": "1230084 B.C. LTD.",
          "legalType": "BEN"
        },
        "courtOrder": {
          "effectOfOrder": "planOfArrangement",
          "fileNumber": "23234324",
          "orderDetails": "A note about court order"
        },
        ...
        "header": {
          "affectedFilings": [
          ],
          "certifiedBy": "jane doe",
          "colinIds": [],
          "comments": [],
          "date": "2021-05-28T16:17:53.029400+00:00",
          "effectiveDate": "2021-05-28T16:17:53.029454+00:00",
          "filingId": 112177,
          ...
          "name": "courtOrder",
          "status": "COMPLETED",
          "submitter": "archiu@idir"
        }
      }
    }
  ]
}
```

### DIFF
--- a/legal-api/migrations/versions/1fcdb213657d_add_order_details_column.py
+++ b/legal-api/migrations/versions/1fcdb213657d_add_order_details_column.py
@@ -1,0 +1,24 @@
+"""add order_details column
+
+Revision ID: 1fcdb213657d
+Revises: 78ddb7f6f8b5
+Create Date: 2021-05-27 13:29:27.893982
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '1fcdb213657d'
+down_revision = '78ddb7f6f8b5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('filings', sa.Column('order_details', sa.String(length=2000), nullable=True))
+
+
+def downgrade():
+    op.drop_column('filings', 'order_details')
--- a/legal-api/src/legal_api/models/filing.py
+++ b/legal-api/src/legal_api/models/filing.py
@@ -159,6 +159,7 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
             'court_order_date',
             'court_order_effect_of_order',
             'court_order_file_number',
+            'order_details',
             'effective_date',
             'paper_only',
             'colin_only',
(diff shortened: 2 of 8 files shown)


severinbeauvais (commented):
Why not keep the order details comment in the comments table but add a mechanism to not return order details with the regular comments?

cameron-freshworks:
oh nice, so it doesn't appear any front end change would be required here? 
It's just a matter of removing the `orderDetails` from being returned in the `comment` array?

argush3:
> Why not keep the order details comment in the comments table but add a mechanism to not return order details with the regular comments?

Yes, I probably could have gone that way.  I haven't thought about it very deeply but it felt like it might introduce some complexity which we may eventually need.  But for this case, it seemed like it would be a simple approach that required minimal changes both from FE and BE perspective.  

If you think the comments approach is the right way, I can dig a bit deeper into it.

argush3:
> oh nice, so it doesn't appear any front end change would be required here?
> It's just a matter of removing the `orderDetails` from being returned in the `comment` array?

I think the only change for the FE is to grab the `orderDetails` from the `header` property in the filing.  The `comments` array  shouldn't contain anything other than comments staff have made against a filing.  i.e. it should function how it did before.

cameron-freshworks:
> > oh nice, so it doesn't appear any front end change would be required here?
> > It's just a matter of removing the `orderDetails` from being returned in the `comment` array?
> 
> I think the only change for the FE is to grab the `orderDetails` from the `header` property in the filing. The `comments` array shouldn't contain anything other than comments staff have made against a filing. i.e. it should function how it did before.

Currently it is taking the Order Details from the Staff Filing itself. That would remain unchanged no? 

Screen Shot 2021-05-28 at 1 08 28 PM

andrepestana-aot:
> > If you think the comments approach is the right way, I can dig a bit deeper into it.
> 
> It makes sense to me to keep all comments in one table and maybe not create a special case in the filings table.
> 
> Also, the comments table already contains 2 types of comments.
> 
> Let's see what others think.

I like this idea. Maybe adding a `type` to comments table would help to query the right kind of comment.

argush3:
Moved out of draft and ready for review. @cameron-freshworks @severinbeauvais @vysakh-menon-aot @andrepestana-aot @thorwolpert.  Main update since moving out of draft was the update of failing entity filer unit tests.

vysakh-menon-aot (approved):

thorwolpert (approved):

PR #1213, opened 2021-06-22
7731 corp_type table

vysakh-menon-aot:
*Issue #:* /bcgov/entity#7731

*Description of changes:*


legal-api/tests/unit/models/test_corp_type.py
```diff
@@ -0,0 +1,58 @@
...
+        'full_desc': 'BC Benefit Company',
+        'legislation': 'BC Business Corporations Act'
+    }
+    assert corp_type_json == corp_type.json
+
+
+def test_find_corp_type_by_id(session):
+    """Assert that the method returns the corp type matching the id."""
+    corp_type = CorpType.find_by_id(Business.LegalTypes.BCOMP.value)
+    corp_type_json = {
```
argush3, line 40: I think hardcoding the values in `corp_type_json` would make more sense for your testing assertions.

argush3 (approved):

legal-api/tests/unit/models/test_corp_type.py
```diff
@@ -0,0 +1,52 @@
...
+    corp_type = CorpType.find_by_id(Business.LegalTypes.BCOMP.value)
+
+    assert corp_type
+    assert corp_type_json == corp_type.json
+
+
+def test_find_all(session):
+    """Assert that the method returns all corp types."""
+    corp_types = CorpType.find_all()
+    assert len(corp_types) == 46
```
thorwolpert, line 50: this should match ```len(Business.LegalTypes)```

legal-api/tests/unit/models/test_corp_type.py
```diff
@@ -0,0 +1,52 @@
...
+    legal_types = [legal_type.value for legal_type in Business.LegalTypes]
```
thorwolpert, line 51: ```Business.LegalTypes.__members__``` returns a dict, that you can the match the key & value against the Enum, to ensure the table and Enum match.

legal-api/src/legal_api/models/corp_type.py
```diff
@@ -0,0 +1,57 @@
...
+    short_desc = db.Column('short_desc', db.String(25), nullable=False)
+    full_desc = db.Column('full_desc', db.String(100), nullable=False)
+    legislation = db.Column('legislation', db.String(100))
+
+    @property
+    def json(self):
+        """Return a dict of this object, with keys in JSON format."""
+        corp_type = {
+            'corp_type_cd': self.corp_type_cd,
+            'colin_ind': self.colin_ind,
```
thorwolpert, line 38: do we care that this in Colin?

legal-api/src/legal_api/models/corp_type.py
```diff
@@ -0,0 +1,57 @@
...
+    colin_ind = db.Column('colin_ind', db.String(1), nullable=False)
+    corp_class = db.Column('corp_class', db.String(10), nullable=False)
+    short_desc = db.Column('short_desc', db.String(25), nullable=False)
+    full_desc = db.Column('full_desc', db.String(100), nullable=False)
+    legislation = db.Column('legislation', db.String(100))
+
+    @property
+    def json(self):
+        """Return a dict of this object, with keys in JSON format."""
+        corp_type = {
```
thorwolpert, line 36: If you're tracking that the entity is managed in colin, do you want to track the lear manages it as well?

legal-api/src/legal_api/models/corp_type.py
```diff
@@ -0,0 +1,57 @@
+# Copyright © 2021 Province of British Columbia
```
thorwolpert, line 1: this is a very slow moving set of data.
Any changes to it probably require the test suite to run.
and sine there's no tracking to say if Lear manages this entity type yet, could it just be a set of Enum & Dicts?

thorwolpert (commented):
A couple of questions around our implementation style, and a suggested change on the Enum test check.

legal-api/src/legal_api/models/corp_type.py
vysakh-menon-aot: Not really. It's in the requirement to include all columns in corp_type table

legal-api/tests/unit/models/test_corp_type.py
vysakh-menon-aot: `Business.LegalTypes.__members__` can match the key, for values `Business.LegalTypes._value2member_map_` this seems to be right

legal-api/src/legal_api/models/corp_type.py
vysakh-menon-aot: That make sense. But we are planning to support all of it eventually in lear. right?

legal-api/src/legal_api/models/corp_type.py
vysakh-menon-aot: It could be but the requirement says we need a table with all entity type listed in the attached doc. Maybe we can include a new column to track `lear_ind`. But do we validate this now?

legal-api/tests/unit/models/test_corp_type.py
thorwolpert: I was suggesting
```assert len(corp_types) == len(Business.LegalTypes)``` instead of the fixed val 46

vysakh-menon-aot: I have already changed this

PR #1215, opened 2021-06-26
implemented NR Type Map web service

severinbeauvais:
*Issue #:* /bcgov/entity#7974

*Description of changes:*
- implemented NR Type Map web service
- added Request Actions enum
- updated Legal Types enum
- implemented unit test
- updated legal-api version number
- ~also disabled pylint error globally that was caused by a different PR~


severinbeauvais:
@thorwolpert Could you have a quick look and provide initial feedback please?

legal-api/setup.cfg
```diff
@@ -87,7 +87,8 @@ testpaths = tests
 addopts = --verbose
     --strict
     -p no:warnings
-    --cov=src --cov-report html:htmlcov --cov-report xml:coverage.xml
+    # --cov=src --cov-report html:htmlcov --cov-report xml:coverage.xml
+    --no-cov
```
severinbeauvais, line 91: I won't actually commit this.

legal-api/src/legal_api/models/business.py
```diff
@@ -48,23 +48,46 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes
...
         ULC_CO_1878 = 'UQC'
         ULC_CO_1890 = 'UQD'
         ULC_CO_1897 = 'UQE'
         BC_CCC = 'CC'  # BC Community Contribution Company
+        # *** the following are from namex project ***
+        # CORPORATION = 'CR'  # use COMP instead
+        UNLIMITED_LIABILITY_COMPANY = 'UL'
+        SOLE_PROPRIETORSHIP = 'FR'
+        GENERAL_PARTNERSHIP = 'GP'
+        DOING_BUSINESS_AS = 'DBA'
```
thorwolpert, line 73: DBA is not a business type.

This was a dump of all the business types from colin https://github.com/bcgov/business-schemas/blob/master/src/registry_schemas/schemas/business.json#L20
and bcgov/entity#7731 already defined all the corp types

legal-api/src/legal_api/resources/nr_type_map.py
```diff
@@ -0,0 +1,197 @@
...
+        nr_type_cd = (request.args.get('nrTypeCd', None).upper()) \
+            if request.args.get('nrTypeCd', None) else None
+        entity_type_cd = (request.args.get('entityTypeCd', None).upper()) \
+            if request.args.get('entityTypeCd', None) else None
+        request_action_cd = (request.args.get('requestActionCd', None).upper()) \
+            if request.args.get('requestActionCd', None) else None
+
+        # build list of tuples
+        # should match Request Type Mapping in namex project
+        request_type_mapping = [
```
thorwolpert, line 67: suggest this be a dict, gives a O(1) instead of an O(n)

legal-api/tests/unit/api/test_nr_type_map.py
```diff
@@ -0,0 +1,94 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests to assure the Name Request Type mapping service.
+
+Test-Suite to ensure that the /nrTypeMap endpoint is working as expected.
+"""
+
+# test with no parameters
+def test_no_parameters(client):
```
thorwolpert, line 21: consider collapsing these to a parameterized aka table test.
makes it easier for the next person to add cases.

thorwolpert (changes requested):
- some suggested changes
- collides with bcgov/entity#7731 so that bus types should get removed from this PR

legal-api/src/legal_api/models/business.py
severinbeauvais: Thanks. I rebased to the 7731 changes. (I still have a few types I don't know how to map across - see comment in ticket.)

legal-api/src/legal_api/resources/nr_type_map.py
severinbeauvais: I wanted to do this but:
- there is no unique key
- I wanted to allow querying by any column (eg, `GET /api/v1/nrTypeMap?nrTypeCd=CR`

Thoughts?

legal-api/src/legal_api/resources/nr_type_map.py
```diff
@@ -0,0 +1,197 @@
...
+            # NR Type Code, Legal Type the NR type code is valid for, Filing Type we can do
+            ('CR', Business.LegalTypes.COMP.value, 'NEW_AML'),
+            ('CR', Business.LegalTypes.COMP.value, 'NEW'),
```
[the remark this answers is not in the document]
severinbeauvais: I think so.

I think NEW is a Names concept (ie, new company) while CREATE DRAFT is a Lear concept.

Is this an issue... should we make this column more Lear centric?

legal-api/src/legal_api/models/business.py
severinbeauvais: Hey, should I sort these? By key or value?

severinbeauvais: I commented out the entity types that Lear doesn't yet support.

(I think this is better than deleting the code as we may want to enable these some time.)

legal-api/tests/unit/api/test_nr_type_map.py
severinbeauvais: Parameterized and complete.

legal-api/src/legal_api/models/business.py
```diff
@@ -94,6 +95,20 @@ class LegalTypes(Enum):
         MISC_FIRM = 'MF'
         FINANCIAL = 'FI'
         CONT_IN_SOCIETY = 'CS'
+        # UNLIMITED_LIABILITY_COMPANY = 'UL'
```
severinbeauvais, line 98: I expect we will enable (uncomment) these some time.

legal-api/src/legal_api/models/business.py
```diff
@@ -94,6 +95,20 @@ class LegalTypes(Enum):
...
+        # LIMITED_PARTNERSHIP = 'LP'
+        # LIMITED_LIABILITY_PARTNERSHIP = 'LL'
+        # COMMUNITY_CONTRIBUTION_COMPANY = 'CC'
+        # XPRO_CORPORATION = 'XCR'
+        # XPRO_UNLIMITED_LIABILITY_COMPANY = 'XUL'
+        # XPRO_LIMITED_LIABILITY_COMPANY = 'RLC'
+        # XPRO_LIMITED_PARTNERSHIP = 'XLP'
+        # XPRO_LIMITED_LIABILITY_PARTNERSHIP = 'XLL'
+        # XPRO_COOPERATIVE = 'XCP'
+        # FIRM = 'FIRM'
```
severinbeauvais, line 111: Note that these are not sorted, but their order is the same as in their original project (ie, to make it easy to compare code between projects).

severinbeauvais:
^^^ linting issue is from a different commit; someone else already said they would fix it

legal-api/setup.cfg
```diff
@@ -72,7 +72,7 @@ max_line_length=120
 notes=FIXME,XXX,TODO
 ignored-modules=flask_sqlalchemy,sqlalchemy,SQLAlchemy,alembic,scoped_session
 ignored-classes=scoped_session
-disable=C0301,W0511,W0703,R0801,R0902,R0401,R1718
+disable=C0301,W0511,W0703,R0801,R0902,R0401,R1718,E0605
```
thorwolpert, line 75: E0605 shouldn't be added

legal-api/src/legal_api/models/business.py
```diff
@@ -94,6 +95,20 @@ class LegalTypes(Enum):
         MISC_FIRM = 'MF'
         FINANCIAL = 'FI'
         CONT_IN_SOCIETY = 'CS'
+        # UNLIMITED_LIABILITY_COMPANY = 'UL'
+        # SOLE_PROPRIETORSHIP = 'FR'
+        # GENERAL_PARTNERSHIP = 'GP'
```
thorwolpert, line 100: a bunch of these are duplicates of ones already in the enum, according to the code and types.
these should be removed

legal-api/src/legal_api/resources/nr_type_map.py
```diff
@@ -0,0 +1,199 @@
...
+            # NR Type Code, Legal Type the NR type code is valid for, Filing Type we can do
+            ('CR', Business.LegalTypes.COMP.value, RequestActions.NEW_AML.value),
+            ('CR', Business.LegalTypes.COMP.value, RequestActions.NEW.value),
+            ('CR', Business.LegalTypes.COMP.value, RequestActions.AML.value),
+            ('CCR', Business.LegalTypes.COMP.value, RequestActions.CHG.value),
+            ('CT', Business.LegalTypes.COMP.value, RequestActions.MVE.value),
+            ('RCR', Business.LegalTypes.COMP.value, RequestActions.REST.value),
+            ('RCR', Business.LegalTypes.COMP.value, RequestActions.REH.value),
+            ('RCR', Business.LegalTypes.COMP.value, RequestActions.REN.value),
+            # ('XCR', Business.LegalTypes.XPRO_CORPORATION.value, RequestActions.NEW_AML.value),
```
thorwolpert, line 79: why are these commented out?

legal-api/src/legal_api/resources/nr_type_map.py
```diff
@@ -0,0 +1,199 @@
...
+            ('BERE', Business.LegalTypes.BCOMP.value, RequestActions.REH.value),
+            ('BERE', Business.LegalTypes.BCOMP.value, RequestActions.REN.value),
+            ('BECV', Business.LegalTypes.BCOMP.value, RequestActions.CNV.value),
+            ('BECR', Business.LegalTypes.COMP.value, RequestActions.CNV.value)
+        ]
+
+        rv = []
+
+        # get list of matching items
+        for item in sorted(request_type_mapping):
```
thorwolpert, line 184: not worth sorting the static info, but making it a runtime, all the time call?

thorwolpert (changes requested):
would like the duplicate enum (although commented out) removed.

legal-api/setup.cfg
andrepestana-aot: @severinbeauvais Those lint errors can be fixed adding a comma to make it a tuple: 
```
__all__ = ('API',)
```

legal-api/src/legal_api/resources/nr_type_map.py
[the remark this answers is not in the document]
andrepestana-aot: I don't think it is an issue if we always CREATE DRAFT for new filings in Lear.

legal-api/setup.cfg
severinbeauvais: Removed in next commit.

severinbeauvais:
> would like the duplicate enum (although commented out) removed.

Good catch. I merged the duplicate enums.

legal-api/src/legal_api/resources/nr_type_map.py
severinbeauvais: I didn't want to sort the static data (so that its order matched where we got it from, to make comparisons with the original code easier).

I removed `sorted()` from here.

legal-api/setup.cfg
andrepestana-aot: I think that now you can rebase it. I think I fixed all of them in a previous change.

legal-api/src/legal_api/resources/nr_type_map.py
severinbeauvais: Because they are valid mappings but Lear doesn't yet support the legal types they refer to. IMHO it's better to leave the mapping intact (ie, similar to its original project). These can simply be uncommented (and order preserved) when Lear supports the new legal types.

legal-api/src/legal_api/models/business.py
severinbeauvais: Good catch. I merged the duplicate enums.

legal-api/setup.cfg
severinbeauvais: Rebased in next commit. (Lint error was indeed fixed.)

legal-api/src/legal_api/resources/nr_type_map.py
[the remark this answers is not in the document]
severinbeauvais: OK.

The purpose of this code is to provide a mapping. Lear can do whatever it wants with the response, for example, Lear can say "NEW is a valid action so I will CREATE a DRAFT".

legal-api/src/legal_api/resources/nr_type_map.py
andrepestana-aot: I think this list will be used as a reference for what could be done but still every consumer will filter themselves to the legalTypes they want to deal with. 
E.g. [#6852](https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/6852) will filter by the 3 legalTypes (CR, UL, CC) and check the options to affiliate these 3 types only.
Maybe @thorwolpert has something different in mind... thoughts? :)

severinbeauvais: If all valid mappings should be returned then I will need to re-enable the Legal Types enums (even though Lear doesn't support them yet).

Let me know what you prefer. (I prefer returning all mappings.)

severinbeauvais: @thorwolpert What do you think about adding a tuple value indicating whether Lear supports the current type?

This would allow the table to return all mappings, and would allow Sumesh (Relationships) to know which types are / not supported.

For example, `GET /nrTypeMap?nrTypeCd=XCR` would return the 3 matching items and a flag on each of them indicating "xpro corporation is not yet supported".

PR #1216, opened 2021-06-26
Fixed annual report date logic; fixed tests

andrepestana-aot:
*Issue #:* /bcgov/entity#7776

*Description of changes:*

- Fixed AR dates logic
- Fixed tests


legal-api/src/legal_api/services/filings/validations/annual_report.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Validation for the Annual Report filing."""
+from datetime import timedelta
```
thorwolpert, line 15: should be using https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/utils/datetime.py
```from legal_api.utils.datetime import timedelta```
or fix the base one if necessary

File: legal-api/tests/unit/api/test_business_filings/test_filings.py [hunk 2]
```diff
@@ -18,6 +18,7 @@
 """
 import copy
 from datetime import datetime
+from dateutil.relativedelta import relativedelta
```
thorwolpert, line 21: i think everywhere else we use 
```import datedelta```

thorwolpert (changes requested):
A couple of changes to use our common libraries.

legal-api/src/legal_api/services/filings/validations/annual_report.py
andrepestana-aot: Ok. Used timedelta from legal_api.utils.datetime

legal-api/tests/unit/api/test_business_filings/test_filings.py
andrepestana-aot: ok. Changed to datedelta here.

vysakh-menon-aot (approved):

thorwolpert (approved):

PR #1222, opened 2021-07-05
8036 duplicate message tracking tests for entity emailer

argush3:
*Issue #:* /bcgov/entity#8036

*Description of changes:*

* Added tests to verify message tracking works for following test scenarios:
  * successfully track processed message
  * does not try to re-process processed message
  * successfully track failure when exception thrown
  * failed message will succeed on retry if original error was a send email exception and email exception no longer thrown
  * retry count is properly tracked

### DIFF
--- a/queue_services/entity-emailer/tests/conftest.py
+++ b/queue_services/entity-emailer/tests/conftest.py
@@ -22,6 +22,10 @@ from contextlib import contextmanager
 import pytest
 from flask import Flask
 from flask_migrate import Migrate, upgrade
+from entity_emailer.config import get_named_config  # noqa: I001
+from tracker.config import get_named_config as get_tracker_named_config  # noqa: I001
+from tracker.models import db as tracker_db  # noqa: I001
+from entity_emailer import worker  # noqa: I001
 from legal_api import db as _db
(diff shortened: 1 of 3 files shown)


vysakh-menon-aot (approved):

queue_services/entity-emailer/tests/conftest.py
```diff
@@ -102,11 +117,44 @@ def client_id():
 
 
 @pytest.fixture(scope='session')
-def db(app):  # pylint: disable=redefined-outer-name, invalid-name
+def db(app, tracker_app):  # pylint: disable=redefined-outer-name, invalid-name
```
thorwolpert, line 120: tracker should be its own fixture

thorwolpert (changes requested):
tracker should be in its own fixture

queue_services/entity-emailer/tests/conftest.py
argush3: moved tracker_db to own fixture

argush3:
@thorwolpert ready for another look.  thanks!

thorwolpert (approved):

PR #1232, opened 2021-07-14
7776 - Fix for AR max date

andrepestana-aot:
*Issue #:* /bcgov/entity#7776

*Description of changes:*

- Fixed ar_max_date (added 1 year and limited to today)
- Fixed nextAnnualReport time offset
- Added and improved tests

### DIFF
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -25,6 +25,7 @@ from sqlalchemy.orm import backref
 
 from legal_api.exceptions import BusinessException
 from legal_api.utils.datetime import datetime, timezone
+from legal_api.utils.legislation_datetime import LegislationDatetime
 
 from .db import db  # noqa: I001
 from .address import Address  # noqa: F401 pylint: disable=unused-import; needed by the SQLAlchemy relationship
@@ -250,7 +251,9 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes
             'identifier': self.identifier,
             'lastModified': self.last_modified.isoformat(),
             'lastAnnualReport': datetime.date(self.last_ar_date).isoformat() if self.last_ar_date else '',
-            'nextAnnualReport': datetime.from_date(ar_min_date).astimezone(timezone.utc).isoformat(),
+            'nextAnnualReport': LegislationDatetime.as_legislation_timezone_from_date(
+                self.next_anniversary
+            ).astimezone(timezone.utc).isoformat(),
             'lastAnnualGeneralMeetingDate': datetime.date(self.last_agm_date).isoformat() if self.last_agm_date else '',
             'lastLedgerTimestamp': self.last_ledger_timestamp.isoformat(),
             'legalName': self.legal_name,
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -179,14 +179,17 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes
             # is required. We need more discussion to understand different scenario's which can come across in future.
             if next_ar_year == 2020:
                 # For year 2020, set the max date as October 31th next year (COVID extension).
-                ar_max_date = datetime(next_ar_year, 10, 31).date()
+                ar_max_date = datetime(next_ar_year + 1, 10, 31).date()
             else:
                 # If this is a CO-OP, set the max date as April 30th next year.
-                ar_max_date = datetime(next_ar_year, 4, 30).date()
+                ar_max_date = datetime(next_ar_year + 1, 4, 30).date()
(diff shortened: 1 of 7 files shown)


legal-api/src/legal_api/utils/datetime.py
```diff
@@ -30,5 +34,5 @@ def utcnow(cls):
 
     @classmethod
     def from_date(cls, date_obj):
-        """Get a datetime object from a date object."""
-        return datetime(date_obj.year, date_obj.month, date_obj.day)
+        """Get a datetime object from a date object and localize it to Pacific time."""
+        return datetime(date_obj.year, date_obj.month, date_obj.day, tzinfo=PACIFIC_TZ)
```
severinbeauvais, line 38: As I don't know where else `from_date` is used, this seems dangerous.

Generally, the API should work with UTC dates.

The AGM date is an exception because it's working with "Pacific dates".

I'd be tempted to clone this method, name the copy `from_date_pacific`, and then use that only where you need it.

legal-api/src/legal_api/models/business.py
```diff
@@ -179,15 +179,18 @@ def get_ar_dates(self, next_ar_year):
...
-                ar_max_date = datetime(next_ar_year, 10, 31).date()
+                ar_max_date = datetime(next_ar_year + 1, 10, 31).date()
             else:
                 # If this is a CO-OP, set the max date as April 30th next year.
-                ar_max_date = datetime(next_ar_year, 4, 30).date()
+                ar_max_date = datetime(next_ar_year + 1, 4, 30).date()
         elif self.legal_type == self.LegalTypes.BCOMP.value:
             # For BCOMP min date is next anniversary date.
-            ar_min_date = datetime(next_ar_year, self.next_anniversary.month, self.next_anniversary.day).date()
+            ar_min_date = datetime(next_ar_year, self.founding_date.month, self.founding_date.day).date()
```
thorwolpert, line 188: next_anniversary is calculated as: 
```python
    def next_anniversary(self):
        """Retrieve the next anniversary date for which an AR filing is due."""
        last_anniversary = self.founding_date
        if self.last_ar_date:
            last_anniversary = self.last_ar_date

        return last_anniversary + datedelta.datedelta(years=1)
```
what's wrong with that?
```suggestion
            ar_min_date = datetime(next_ar_year, self.next_anniversary.month, self.next_anniversary.day).date()
```

thorwolpert (commented):
in addition to the DM request

legal-api/src/legal_api/models/business.py
andrepestana-aot: I thought that nextAnnualReport should be the date for the next_year to be filed and should change in every `task` element in the json as do arMinDate and arMaxDate. Otherwise it will be repeated in all `task` elements and should equal to the least arMinDate.

I'm gonna revert it to next_anniversary then.

legal-api/src/legal_api/utils/datetime.py
andrepestana-aot: It uses UTC dates. I'm using LegislationDatetime now. There are some tests to check it.

legal-api/src/legal_api/utils/legislation_datetime.py
```diff
@@ -41,6 +41,11 @@ def as_legislation_timezone(date_time: datetime) -> datetime:
         """Return a datetime adjusted to the legislation timezone."""
         return date_time.astimezone(pytz.timezone(current_app.config.get('LEGISLATIVE_TIMEZONE')))
 
+    @staticmethod
+    def as_legislation_timezone_from_date(_date: datetime.date) -> datetime.date:
+        """Return a datetime adjusted to the legislation timezone from a date object."""
+        return LegislationDatetime.as_legislation_timezone(datetime.datetime(_date.year, _date.month, _date.day))
```
andrepestana-aot, line 47: It seems that `as_legislation_timezone` doesn't make any difference in dev env. I'm going to change it to `gettz`

legal-api/src/legal_api/utils/datetime.py
```diff
@@ -14,8 +14,7 @@
 """Date time utilities."""
 # from datetime import datetime, timezone
 import time as _time
-from datetime import date, datetime as _datetime, timezone, \
-    timedelta  # pylint: disable=unused-import # noqa: F401, I001, I005
+from datetime import date, datetime as _datetime, timezone  # pylint: disable=unused-import # noqa: F401, I001, I005
```
severinbeauvais, line 17: Why not delete the unused import instead disabling the lint issue ?

andrepestana-aot: This import is needed. Some other classes import from here.

thorwolpert (approved):

PR #1245, opened 2021-08-20
8628 incorporate co-op schema change

vysakh-menon-aot:
*Issue #:* /bcgov/entity#8628

*Description of changes:*
  - Fixed entity-filer lint and test issues

### DIFF
--- a/queue_services/entity-filer/Makefile
+++ b/queue_services/entity-filer/Makefile
@@ -59,7 +59,9 @@ install-dev: ## Install local application
 #################################################################################
 # COMMANDS - CI                                                                 #
 #################################################################################
-ci: lint flake8 test ## CI flow
+ci: lint test ## CI flow
+
+lint: pylint flake8
 
 pylint: ## Linting with pylint
 	. venv/bin/activate && pylint --rcfile=setup.cfg src/$(PROJECT_NAME)
--- a/queue_services/entity-filer/Makefile
+++ b/queue_services/entity-filer/Makefile
@@ -61,7 +61,7 @@ install-dev: ## Install local application
 #################################################################################
 ci: lint flake8 test ## CI flow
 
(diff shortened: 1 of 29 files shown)


andrepestana-aot (approved):
LGTM

queue_services/entity-filer/Makefile
```diff
@@ -61,7 +61,7 @@ install-dev: ## Install local application
 #################################################################################
 ci: lint flake8 test ## CI flow
 
-lint: ## Linting with pylint
```
thorwolpert, line 64: we're moving to a target of just lint ... but people keep mucking with this.
please just make it lint, which should call both pylint & flake8 targets

queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
```diff
@@ -21,11 +21,9 @@
 from legal_api.models import Business, Filing
 from legal_api.utils.datetime import datetime
 
-from entity_filer.filing_processors.filing_components import create_office
+from entity_filer.filing_processors.filing_components import create_office, filings
```
thorwolpert, line 24: why is tis added to dissolution?

thorwolpert (commented):
in future, we should try and do a PR per sub-project as the CI is not designed to run as a mono-repo, so changes across the api, filer, etc. will not pass the CI when dependent

queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
vysakh-menon-aot: I have removed 2nd import line from `entity_filer.filing_processors.filing_components`. Lint was giving me that error

thorwolpert (approved):

PR #1252, opened 2021-09-01
8639 - Bumped versions for release 10.3

andrepestana-aot:
*Issue #:* /bcgov/entity#8639
*Description of changes:*

- Bumped versions for release 10.3


argush3 (changes requested):
I believe there are colin api changes that need to go out so the version number will need to be bumped for it as well.

argush3 (approved):

PR #1280, opened 2021-09-29
8927 - Fix for Org parties

andrepestana-aot:
*Issue #:* /bcgov/entity#8927

*Description of changes:*

- Fixed partyType;
- Added tests


legal-api/src/legal_api/models/party.py
```diff
@@ -32,7 +32,7 @@ class PartyTypes(Enum):
         """Render an Enum of the party types."""
 
         PERSON = 'person'
-        ORGANIZATION = 'organization'
+        ORGANIZATION = 'org'
```
thorwolpert, line 35: maybe we should change the schema to match ?
checking schema and they spell out the full item **organization**

legal-api/tests/unit/api/v1/test_business_director.py
```diff
@@ -34,7 +34,9 @@ def test_get_business_directors(session, client, jwt):
     officer = {
         'firstName': 'Michael',
         'lastName': 'Crane',
-        'middleInitial': 'Joe'
+        'middleInitial': 'Joe',
+        'partyType': 'person',
+        'orgName': ''
```
severinbeauvais, line 39: Does this mean that `orgName` is now a required property?

Does the schema need to be updated?

thorwolpert (commented):
question on updating schema

legal-api/src/legal_api/models/party.py
andrepestana-aot: If we change the schema, then we need to change the UIs too. If we do really need to change 'org' to 'organization' for some reason, I'm ok changing the schema now. Otherwise I would leave it as it is now.

legal-api/tests/unit/api/v1/test_business_director.py
andrepestana-aot: No, it is required only for the tests as part of the `officer` dict.

andrepestana-aot:
Converted to draft as it'll need to change the schema to use 'organization' instead of 'org'.

legal-api/src/legal_api/models/party.py
severinbeauvais: This conversation is outdated.

thorwolpert:
Probably need to rebase this so it isn't picking up all these commits that aren't part of this change

andrepestana-aot:
Sorry.. I'm closing this one. I created another one for this: https://github.com/bcgov/lear/pull/1283

PR #1287, opened 2021-10-05
adding proxy for receipt through the document end-point

thorwolpert:
*Issue #:* /bcgov/entity#7803

*Description of changes:*
- adding proxy for receipt through the document end-point


thorwolpert:
SonarCloud - is complaining that 2 of the tests are nearly identical.
1 hist the function and the other the endpoint, so there are differences.

vysakh-menon-aot (approved):

PR #1295, opened 2021-10-13
9221 - Removed draft filing for staff payment

andrepestana-aot:
*Issue #:* /bcgov/entity#9221
*Description of changes:*

- Removed draft filing for staff payment; 
- Added tests.


andrepestana-aot:
> I thought we decided to leave the current functionality: to retain the draft filing even when a staff payment fails, and to display a dialog allowing staff to stay on the filing page or return to the dashboard ???
> 
> -> need to make sure staff can delete the draft filing from the dashboard -> need to make sure staff can resume the filing and try paying again

I understood we wouldn't save draft for the staff and keep them at the filing for them to try again. They can save it as a draft in the filing form though.

PR #1318, opened 2021-11-01
Verify Payment for Documents

cameron-freshworks:
*Issue #:* /bcgov/entity#9638

*Description of changes:*
* Verify payment status code before applying a receipt to the documents list to the ledger item. 
* This is to prevent documents or receipts displaying for documents that have not been completed within our system.

### DIFF
--- a/legal-api/src/legal_api/core/filing.py
+++ b/legal-api/src/legal_api/core/filing.py
@@ -25,7 +25,6 @@ from typing import Dict, List, Optional
 from flask import current_app, url_for
 from flask_jwt_oidc import JwtManager
 from sqlalchemy import desc
-from pprint import pprint
 
 from legal_api.core.meta import FilingMeta
 from legal_api.core.utils import diff_dict, diff_list
--- a/legal-api/src/legal_api/core/filing.py
+++ b/legal-api/src/legal_api/core/filing.py
@@ -410,14 +410,7 @@ class Filing:
                                                    'legal_filing_name': None})
 
         documents = {'documents': {}}
-
-        has_payment_rules = [
-            filing.storage,
-            filing.storage.payment_status_code,
-            filing.storage.payment_status_code == 'COMPLETED'
-        ]
-
-        if all(has_payment_rules):
+        if filing.storage and filing.storage.payment_completion_date:
             documents['documents']['receipt'] = f'{base_url}{doc_url}/receipt'
 
         if filing.status in (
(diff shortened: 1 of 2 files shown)


thorwolpert:
Historical filings, and possibly the drafts will be need to have the documents link.
Is there a problem if the doc links returns a []?

cameron-freshworks:
> Historical filings, and possibly the drafts will be need to have the documents link. Is there a problem if the doc links returns a []?

oh ok for sure. 
No i don't think so, as long as we don't include the receipt conditionally i think. 
So could i do the same check but to apply the receipt conditionally in `get_document_list()`

legal-api/src/legal_api/core/filing.py
```diff
@@ -409,9 +409,16 @@ def get_document_list(business, filing, request) -> Optional[dict]:
...
 
-        documents = {'documents': {
-                     'receipt': f'{base_url}{doc_url}/receipt'
-                     }}
+        documents = {'documents': {}}
+
+        has_payment_rules = [
+            filing.storage,
+            filing.storage.payment_status_code,
+            filing.storage.payment_status_code == 'COMPLETED'
```
argush3, line 417: I think you want to add `filing.storage.payment_completion_date` to `has_payment_rules` too.  When the payment queue message is officially processed by entity-pay queue listener, `payment_completion_date` is populated.

cameron-freshworks: Is there scenarios where there could be a `COMPLETED` state without a `payment_completion_date` or vice versa?
Are they both required indicators that a payment has been made, or is one or another a better indicator?
Can definitely add it in, just trying to deduce the best option.

argush3: yeah, i think just using `payment_completion_date` would be ideal.  Let me dig through the data a bit.  @thorwolpert any thoughts?  should just checking with `payment_completion_date` be good enough?

thorwolpert: I'd be careful on the list set up, as it relies on those pieces being not None anyways, unlike an if evaluation, so they can be dropped in the list setup.


```status == 'COMPLETED'``` without payment for filings loaded from COLIN
and we'll need to figure out what to do for data loading, as we have the historical payment info there ... 
In our current state though, I think @argush3 is right that we only need to check the ```payment_completion_date``` for filings that will a receipt. 
For historical filings, where it's not paper_only and COMPLETED, the filings can be shown, but **not** the receipt

cameron-freshworks: Awesome! Ok thanks, will make these adjustments 👍

cameron-freshworks: Just in testing against DEV data, i've noticed a lot of filings that have a `payment_completion_date` value and the `payment_status_code` is `None`. 
Is this a scenario we might expect in a real environment. I ask, because if we use this DATE we are still presenting receipt links for the user but they all result in 404.

argush3 (approved):

PR #1319, opened 2021-11-02
9914 manage business state

thorwolpert:
*Issue #:* /bcgov/entity#9914

### Administrative locks
- they owe money
- bad actors

### Dissolution jobs
- Business move through notification states based on lack of AR filings, possible other things
- 

### Determine what filings can be undertaken
- Dissolved
  - Restoration
- state moving towards dissolution (missed an AR)
  - only an AR?

## State ```@property```
Do we need multiple data elements to provide the current _state_, do staff need to see multiple state(eg. dissolution_stage_2, and administrative lock).

## endpoint
Are valid_actions a second class item under business? Are they a parameter ask? Should they be a little hateos and be an included array of options valid for the business, in its current state?

##


thorwolpert:
Add discussion here, as we agree on things, we'll edit the main-block

lmcclung:
A few additions...

Dissolution jobs 
- "other things" includes in contravention of Registries legislation, on request of another governing body, or other administrative reasons
- Two types of dissolutions (involuntary due to missing AR's) and Administrative (everything else)
- In Limited Restoration - entity was restored for a set period of time.  When that end date happens it is picked up in a job and is dissolved.
- In Limited Restore Liquidation - entity that was in the process of being liquidated was dissolved for some reason in error and was subsequently restored for a set period of time.  When that end date happens it is picked up in a job and is dissolved.
- Failure to transition to a new version of legislation

New Section - Special in progress Entities with special filings were they are working towards dissolution and need to be flagged
- In Liquidation - from the time an entity appoints a liquidator there needs to be a flag to restrict allowed filings
- In Amalgamation - Linda to confirm we still use this (might have been just NWPTA)

New section - Amalgamations
- two or more entities (amalgamating parties) form to create a third entity in BC.  The amalgamating parties become historical.
- when a BC entity (amalgamating party) amalgamates with another foreign entity to create a third entity outside of BC.  The BC amalgamating party become historical (amalgamation out).

New section - Continuation / Amalgamation outs
- a BC entity wants to leave BC and move to another jurisdiction and retain all its corporate history (continuation out).  -

New section - Cancellation of a registration
- if the entity was registered in BC as an extraprovincial entity, when it ceases to do business in their home jurisdiction we need to make their registration historical in BC.

New section - Converted to BC Company (when converting an entity from one governing legislation to another - usually a society to a corporation)

New section - Withdrawn
- this is a legacy thing. 
- We don't apply filings until the effective date so not an issue?  Still need to figure out traceability for these.

New section - Do not strike
- we want this entity to remain active on our register and not get caught up in any dissolution job.

Allowable filings
Depends on the reason for why the entity was made historical

thorwolpert:
Closing - captured in RFC and other tickets

PR #1320, opened 2021-11-02
9734 filing dates

thorwolpert:
*Issue #:* /bcgov/entity#9734 - part 2

*Description of changes:*
- fix manual deletion lock on filings
- change effective_date to either be in the future, or when the filing is paid for.
- is_future_effective, is set by the Filer, calling set_processed() of Filings, which will set the effective_date


thorwolpert:
SonarCloud: I'm leaving the commented out code, as I don't think it does anything useful, hence commenting it out ... but want to check in Dev

vysakh-menon-aot (approved):

PR #1321, opened 2021-11-02
Send COOP Gazette messaging #8272

stevenc987:
*Issue #:* /bcgov/entity/#8272

*Description of changes:*
Create FTP flat file similar to the existing one generated by home team.


legal-api/src/legal_api/models/filing.py
```diff
@@ -227,6 +228,16 @@ class Source(Enum):
 
     parent_filing_id = db.Column(db.Integer, db.ForeignKey('filings.id'))
     parent_filing = db.relationship('Filing', remote_side=[id], backref=backref('children'))
+    _sent_to_gazette_date = db.Column('sent_to_gazette_date', db.DateTime(timezone=True))
```
thorwolpert, line 231: moves this to after line 195

legal-api/src/legal_api/models/filing.py
```diff
@@ -179,7 +179,8 @@ class Source(Enum):
             'submitter_roles',
             'tech_correction_json',
             'temp_reg',
-            'transaction_id'
+            'transaction_id',
+            '_sent_to_gazette_date'
```
thorwolpert, line 183: please move up so that is fits in alphabetically like the others.
probably after 164

thorwolpert (commented):
Please add some test-data and tests on the output using that data.
The update queries are not based on the processed set of records, so you could end up marking some as ```sent_to_gazette_date``` even though they haven't been sent.

stevenc987:
@thorwolpert 
Regarding to "The update queries are not based on the processed set of records, so you could end up marking some as sent_to_gazette_date even though they haven't been sent.", I followed up similar logic from the sql files provided. In the sql files provides, it select the data and create a file, then update the related database columns before the files were sent. Please let me know if we want to change the logic.

droberts27 (commented):
Is it possible to use dissolutionType to capture more detail for the dissolutions?

stevenc987:
@droberts27 @thorwolpert 
I doubt we can do that

legal-api/migrations/versions/409f6ba0602b_adding_sent_to_gazette_date_to_filings.py
```diff
@@ -0,0 +1,25 @@
+"""empty message
+
+Revision ID: 409f6ba0602b
+Revises: b6997f07700a
```
thorwolpert, line 4: This has split the sequence.
you'll need to change this to what ever is the HEAD of the alembic version ... currently 3d8eb786f4c2 a week ago.

thorwolpert:
@stevenc987 yes, I'm saying the logic needs to change as this is running across several transactions and you can't guarantee that the data set is the same each time you make a query.

stevenc987:
@thorwolpert 
I have sent you a few emails for different questions. Which question were you answering a few minutes ago? Sorry, I am confused.

thorwolpert:
I'm answering the question in this PR.
The logic as it exists can miss or mark business as processed that are not.

stevenc987:
@thorwolpert @droberts27 
Thanks. I will update logic so the update will happen only after the file is sent.

thorwolpert:
@stevenc987 How will you know if you get the same _set_ from your Select, that is the issue.
There are a number of approaches, a process table, temp table, cursor and update the rows as processed, setting a bounds based on date or IDs (if you have guarantees that they always increase).

If run a select based on state, and then run it again it very well could be different if there are new filings added. As the new system will be available 24x7, you can't guarantee a time when no new filing activity is occurring.

stevenc987:
@thorwolpert @droberts27 
It seems this program will run at 1:30 am on every Tuesday which is the same as old system. When you said 24/7, did you mean the new program will be running every minute or once in a day?

stevenc987:
@thorwolpert @droberts27 
Thor, you said this program will run 24/7. However, Linda said we should keep the same running time as David's program which is running on 1:30 am on every Tuesday (only run once a week). Could you please clarify this to me? Please see Linda's comment below. Thanks

image

thorwolpert:
@stevenc987 No,  I said the "the new system will be available 24x7"

Let's setup a time to talk. The SQL you are writing and transaction approach does NOT guarantee the same set of data is returned and you will mark filings as delivered to the gazette that are not included in your set, so we'll have missing filings.

stevenc987:
@thorwolpert 
I have updated the logic of my program. We will retrieve a set of data from filings and businesses table, then insert the filing_id into a new created table. Then all the data in the sent file and status update will be based on the new created table instead of the filings table. By doing this, it will make sure that the data are from the same set. I will check in codes and let you review it.
Thanks

stevenc987:
Used new one (#1330) so this one was closed.

PR #1332, opened 2021-11-16
9053 get modified user for business

vysakh-menon-aot:
*Issue #:* /bcgov/entity#9053

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #1338, opened 2021-11-19
Submitter Display Name

cameron-freshworks:
*Issue #:* /bcgov/entity#10145

*Description of changes:*
* Return the certifiedBy value for the submitter display name.
* Updated unit testing.

*Note:* 
- Why we chose to use the certifiedBy property here is that the `USER` object first name / last name properties seem to be blank when fetching that user data, and the `username` wasn't as 'friendly' from a display point of view. 
Definitely open to any options here though if there is another means of retrieving a submitter value.


lekshmimallika-aot (approved):
Looks good to me.

cameron-freshworks:
Closing PR as this was resolved in another (better) fashion in PR: https://github.com/bcgov/lear/pull/1332

PR #1339, opened 2021-11-21
10161 conditional outputs

vysakh-menon-aot:
*Issue #:* /bcgov/entity#10161

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/core/filing.py
+++ b/legal-api/src/legal_api/core/filing.py
@@ -364,7 +364,7 @@ class Filing:
             if (submitter := filing.filing_submitter) \
                 and submitter.username and jwt \
                     and not Filing.redact_submitter(filing.submitter_roles, jwt):
-                submitter_displayname = submitter.username
+                submitter_displayname = submitter.display_name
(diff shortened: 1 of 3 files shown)


lekshmimallika-aot (commented):
Could you please update the tests as well?

vysakh-menon-aot:
> Could you please update the tests as well?

Sure. I am waiting to get confirmation from @thorwolpert regarding this approach.

lekshmimallika-aot:
> > Could you please update the tests as well?
> 
> Sure. I am waiting to get confirmation from @thorwolpert regarding this approach.

One of the questions I had was whether the filing metadata can be used for this? I know for business type alteration metadata gets populated correctly. If name change also does that may be we can use that?

thorwolpert (commented):
The name change should be in the filing_meta, can we use that or update the meta for what is needed. the filing_json is an expensive call and the format is not guranteed.

thorwolpert (approved):

lekshmimallika-aot (approved):

PR #1341, opened 2021-11-22
Submitter Display Name

cameron-freshworks:
*Issue #:* /bcgov/entity#10145

*Description of changes:*
* Use the display_name property for Submitter name when NOT redacted.
* Updated unit testing and User Factory method to take firstname and lastname.

### DIFF
--- a/legal-api/src/legal_api/core/filing.py
+++ b/legal-api/src/legal_api/core/filing.py
@@ -364,7 +364,7 @@ class Filing:
             if (submitter := filing.filing_submitter) \
                 and submitter.username and jwt \
                     and not Filing.redact_submitter(filing.submitter_roles, jwt):
-                submitter_displayname = submitter.username
+                submitter_displayname = submitter.display_name
 
             ledger_filing = {
                 'availableOnPaperOnly': filing.paper_only,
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/core/filing.py
```diff
@@ -364,7 +364,7 @@ def ledger(business_id: int,
             if (submitter := filing.filing_submitter) \
                 and submitter.username and jwt \
                     and not Filing.redact_submitter(filing.submitter_roles, jwt):
-                submitter_displayname = submitter.username
+                submitter_displayname = submitter.display_name or submitter.username
```
vysakh-menon-aot, line 367: @thorwolpert should we change the logic in `display_name` to return `username` for bcsc aswell instead of this?

vysakh-menon-aot (approved):

legal-api/src/legal_api/core/filing.py
cameron-freshworks: good question!
Was wondering about that as well but was not sure of the implementations of `display_name` and if that would an issue wherever else it's used.

lekshmimallika-aot (approved):
Looks good

PR #1343, opened 2021-11-23
9016 - filer dissolution

andrepestana-aot:
*Issue #:* /bcgov/entity#9016

*Description of changes:*

- updated dissolution filer processor;
- added tests
- added dissolution_type to FilingMeta

### DIFF
--- a/queue_services/entity-filer/src/entity_filer/filing_meta.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_meta.py
@@ -61,7 +61,6 @@ class FilingMeta:
 
     application_date: Optional[datetime] = None
     legal_filings: list = field(default_factory=list)
-    dissolution_type: Optional[str] = None
 
     @property
     def asjson(self):
--- a/queue_services/entity-filer/src/entity_filer/filing_meta.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_meta.py
@@ -61,6 +61,7 @@ class FilingMeta:
 
     application_date: Optional[datetime] = None
     legal_filings: list = field(default_factory=list)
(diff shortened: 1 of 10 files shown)


andrepestana-aot:
@thorwolpert Can you answer Argus' question, please?
> I think we need to add a processor for special resolutions for COOP voluntary dissolutions. I'm thinking it might only require adding meta data to the filing.
> 
> @thorwolpert what are your thoughts? do we need to add anything to the `filing.meta_data` field for the Coop special resolution?

queue_services/entity-filer/src/entity_filer/filing_meta.py
```diff
@@ -61,6 +61,7 @@ class FilingMeta:
 
     application_date: Optional[datetime] = None
     legal_filings: list = field(default_factory=list)
+    dissolution_type: Optional[str] = None
```
thorwolpert, line 64: this will get added to filings we don't want to have it in.

either add it in the dissolution processor, or have it stripped out in the dict.

thorwolpert (commented):
filing_meta should get changed.

queue_services/entity-filer/src/entity_filer/filing_meta.py
andrepestana-aot: ok. Done!

thorwolpert (approved):

argush3:
I know this has been merged already but was talking to @lekshmimallika-aot and she asked whether we need to add an entry to the `resolutions` table when processing the voluntary dissolution filing for Coops?  I do notice that the table is being populated with entries currently.  In DEV anyways.  thoughts @thorwolpert?  cc: @andrepestana-aot

PR #1355, opened 2021-12-02
Resolution model changes

lekshmimallika-aot:
*Issue #:* /bcgov/entity#9016

*Description of changes:*
Added additional optional fields for resolution
Updated test

Note : I have not added a column for resolution confirmed . I think it is more like a UI field. I dont think a filing can be submitted with resolutionConfirmed as False. 


argush3 (commented):
This looks good to me.  @thorwolpert I would like input from you whether the use of the `party` table works for you.  We had introduced the `person` schema recently when working on special resolutions.  Did you want to introduce that in the database as well?

@lekshmimallika-aot yes, the confirm flag is just used for the UI so we can leave it out.

legal-api/migrations/versions/e6ff12b18bef_.py
```diff
@@ -0,0 +1,44 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('resolutions', sa.Column('resolution', sa.String(length=4096), nullable=True))
+    op.add_column('resolutions', sa.Column('signing_date', sa.Date(), nullable=True))
+    op.add_column('resolutions', sa.Column('signing_party_id', sa.Integer(), nullable=True))
+    op.add_column('resolutions', sa.Column('sub_type', sa.String(length=20), nullable=True))
+    op.add_column('resolutions_version', sa.Column('resolution', sa.String(length=4096), nullable=True))
```
argush3, line 25: schema and UI validation has max size at 1000 for validation but this length is probably a good call.  Gives us some wiggle room if we ever need to up it.  👍

argush3 (approved):

vysakh-menon-aot (approved):

PR #1362, opened 2021-12-07
9731 - Add devcontainer config and email payload validation.

pwei1018:
*Issue #:* [/bcgov/entity/9731](https://github.com/bcgov/entity/issues/9731)

*Description of changes:*
1. add devcontrainer setup;
2. add email payload validation in send_email function;

### DIFF
--- a/queue_services/entity-emailer/src/entity_emailer/worker.py
+++ b/queue_services/entity-emailer/src/entity_emailer/worker.py
@@ -80,13 +80,13 @@ def send_email(email: dict, token: str):
             or 'content' not in email \
             or 'body' not in email['content']:
         logger.debug('Send email: email object(s) is empty')
-        raise QueueException('Unsuccessful sending email.')
+        raise QueueException('Unsuccessful sending email - required email object(s) is empty.')
 
     if not email['recipients'] \
             or not email['content'] \
             or not email['content']['body']:
         logger.debug('Send email: email object(s) is missing')
-        raise QueueException('Unsuccessful sending email.')
+        raise QueueException('Unsuccessful sending email - required email object(s) is missing. ')
 
     resp = requests.post(
         f'{APP_CONFIG.NOTIFY_API_URL}',
--- a/queue_services/entity-emailer/src/entity_emailer/worker.py
+++ b/queue_services/entity-emailer/src/entity_emailer/worker.py
@@ -74,6 +74,20 @@ async def publish_event(payload: dict):
 
 def send_email(email: dict, token: str):
     """Send the email."""
+    # stop processing email when payload is incompleted.
+    if not email \
+            or 'recipients' not in email \
+            or 'content' not in email \
+            or 'body' not in email['content']:
+        logger.debug('Send email: email object(s) is empty')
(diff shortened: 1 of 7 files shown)


queue_services/entity-emailer/src/entity_emailer/worker.py
```diff
@@ -74,6 +74,20 @@ async def publish_event(payload: dict):
...
+            or 'content' not in email \
+            or 'body' not in email['content']:
+        logger.debug('Send email: email object(s) is empty')
+        raise QueueException('Unsuccessful sending email.')
+
+    if not email['recipients'] \
+            or not email['content'] \
+            or not email['content']['body']:
+        logger.debug('Send email: email object(s) is missing')
+        raise QueueException('Unsuccessful sending email.')
```
argush3, line 89: I think an ideal solution would be to throw another custom exception here and add another `except` block to `cb_subscription_handler`.  This`except` block would create or update a message in the tracker with a status of `TERMINATED` or something like that.  If we do that then OPs will be able to query tracker db for specific queue messages that were terminated intentionally by us.  With the current state of things, this means we would be able to pick out the NRs where we weren't able to retrieve a receipt from the pay api.

I am fine with this approach though of just throwing the `QueueException` to get the queue to stop re-processing the messages.  @thorwolpert does Patrick's current implementation work for you?

queue_services/entity-emailer/src/entity_emailer/worker.py
```diff
@@ -74,6 +74,20 @@ async def publish_event(payload: dict):
 
 def send_email(email: dict, token: str):
     """Send the email."""
+    # stop processing email when payload is incompleted.
+    if not email \
+            or 'recipients' not in email \
+            or 'content' not in email \
+            or 'body' not in email['content']:
+        logger.debug('Send email: email object(s) is empty')
+        raise QueueException('Unsuccessful sending email.')
```
argush3, line 83: Can we update the message we are using to instantiate `QueueException`?  To maybe something like "Unsuccessful sending email - required email object(s) is missing"?  This way the `message_processing.last_error` field will provide a bit more detail as to what the error was.  And same thing for QueueException call below on line 89.

queue_services/entity-emailer/tests/unit/test_worker.py
```diff
@@ -315,3 +315,52 @@ def test_nr_receipt_notification(app, session):
...
+        'requestBy': 'test@test.ca',
+        'content': {
+            'subject': 'test',
+            'attachments': []
+        }})
+])
+def test_send_email_with_incomplete_payload(app, session, email_msg):
+    """Assert that the email not have body can not be processed."""
+    # TEST
+    with pytest.raises(Exception) as excinfo:
```
argush3, line 363: I think this should be checking for `QueueException`.  i.e. `with pytest.raises(QueueException) as excinfo:`

argush3 (changes requested):
Looks good.  Just the update to the error message used when throwing the `QueueException` and test update.  Unless we add another custom exception and introduce an additional status for tracker message_processing table.

argush3 (approved):

PR #1367, opened 2021-12-10
Model changes - Business state

lekshmimallika-aot:
*Issue #:* /bcgov/entity#9348

*Description of changes:*
Added business state, admin freeze and state filing id to the business model
Updated test


vysakh-menon-aot (approved):

legal-api/migrations/versions/da43f7bef692_.py
```diff
@@ -0,0 +1,43 @@
...
+down_revision = 'd9e21a1b82ca'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    state = postgresql.ENUM('ACTIVE', 'HISTORICAL', 'LIQUIDATION', name='state')
+    state.create(op.get_bind(), checkfirst=True)
+    op.add_column('businesses', sa.Column('state', state, nullable=True))
```
argush3, line 23: what happens to the existing data when we run the migration for this column(`businesses.state`)?  is it null?

lekshmimallika-aot: We usually keep the DML out of migration. We will have to run a script to update the state of existing businesses as ACTIVE.

argush3 (approved):

PR #1377, opened 2021-12-16
Parties endpoint

lekshmimallika-aot:
*Issue #:* /bcgov/entity#10415

*Description of changes:*
Added endpoint for parties
Added tests

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_parties.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_parties.py
@@ -1,0 +1,59 @@
+# Copyright © 2021 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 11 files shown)


vysakh-menon-aot (approved):

legal-api/src/legal_api/resources/v2/business/business_parties.py
```diff
@@ -0,0 +1,59 @@
...
+# limitations under the License.
+"""Retrieve the parties for the entity."""
+from datetime import datetime
+from http import HTTPStatus
+
+from flask import jsonify, request
+from flask_cors import cross_origin
+
+from legal_api.models import Business, PartyRole
+
```
thorwolpert, line 22: Needs the JWT
```suggestion
from legal_api.utils.auth import jwt
```

legal-api/src/legal_api/resources/v2/business/business_parties.py
```diff
@@ -0,0 +1,59 @@
...
+from .bp import bp
+
+
+@bp.route('/<string:identifier>/parties', methods=['GET', 'OPTIONS'])
+@bp.route('/<string:identifier>/parties/<int:party_id>', methods=['GET', 'OPTIONS'])
+@cross_origin(origin='*')
+def get_parties(identifier, party_id=None):
```
thorwolpert, line 29: Needs at least a valid JWT
```suggestion
@jwt.requires_auth
def get_parties(identifier, party_id=None):
```

thorwolpert (commented):
LGTM, but it should be a protected end-point

legal-api/src/legal_api/resources/v2/business/business_parties.py
lekshmimallika-aot: Sure. I totally missed that

PR #1380, opened 2021-12-17
9329 Determine allowable filings actions

vysakh-menon-aot:
*Issue #:* /bcgov/entity#9329

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -26,6 +26,7 @@ from legal_api.core import Filing as CoreFiling
 from legal_api.models import Business, Filing, RegistrationBootstrap
 from legal_api.resources.v1.business.business_filings import ListFilingResource
 from legal_api.services import RegistrationBootstrapService
+from legal_api.services.authz import get_allowed
 from legal_api.utils.auth import jwt
 
 from .bp import bp
@@ -55,6 +56,11 @@ def get_businesses(identifier: str):
     if recent_filing_json:
         business_json['submitter'] = recent_filing_json['filing']['header']['submitter']
         business_json['lastModified'] = recent_filing_json['filing']['header']['date']
+
+    allowed_filings = str(request.args.get('allowed_filings', None)).lower() == 'true'
+    if allowed_filings:
+        business_json['allowedFilings'] = get_allowed(business.state, business.legal_type, jwt)
+
     return jsonify(business=business_json)
 
 
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -46,7 +46,7 @@ from legal_api.services import (
     namex,
     queue,
 )
-from legal_api.services.authz import get_allowed, is_allowed
+from legal_api.services.authz import is_allowed
 from legal_api.services.filings import validate
 from legal_api.services.utils import get_str
 from legal_api.utils import datetime
@@ -412,6 +412,8 @@ class ListFilingResource():
         sub_filing_type = None
         if filing_type == 'restoration':
             sub_filing_type = filing_json['filing'].get('restoration', {}).get('type')
+
+        # While filing IA business object will be None. Setting default values in that case.
         state = business.state if business else Business.State.ACTIVE
         legal_type = business.legal_type if business else filing_json['filing']['business'].get('legalType')
         admin_freeze = business.admin_freeze if business else False
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -62,7 +62,6 @@ class QueryModel(BaseModel):
 
     draft: Optional[bool]
     only_validate: Optional[bool]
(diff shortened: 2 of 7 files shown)


vysakh-menon-aot:
@thorwolpert Does the legal-api support any other legal types other than `['CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC']` for any filing types? With this pr we are blocking any filing with a different legal type

legal-api/src/legal_api/services/authz.py
```diff
@@ -88,3 +90,100 @@ def has_roles(jwt: JwtManager, roles: List[str]) -> bool:
...
+        'staff': ['CP', 'BEN'],
+    },
+    'courtOrder': {
+        'staff': ['CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC'],
+    },
+    'dissolution': {
+        'staff': ['CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC'],
+        'user': ['CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC'],
+    },
+    'incorporationApplication': {
```
lekshmimallika-aot, line 122: All entries for different role look the same to me. I feel this should be other way round. Based on the user role and entity state, you can return the filing types allowed. There could be one or more filing types that is condition based. Those alone can be added based on condition is what I feel

lekshmimallika-aot (commented):
Is admin freeze taken into account here? That is something that will block all actions for user and staff

legal-api/src/legal_api/services/authz.py
```diff
@@ -88,3 +90,100 @@ def has_roles(jwt: JwtManager, roles: List[str]) -> bool:
...
+    'registrarsNotation': {
+        'staff': ['CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC'],
+    },
+    'registrarsOrder': {
+        'staff': ['CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC'],
+    },
+    'restoration': {
+        'staff': ['CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC'],
+    },
+}
```
severinbeauvais, line 158: Please look at the kinds of actions that the UI currently "allows"... Will the API be able to provide True/False for a similar list?

https://github.com/bcgov/business-filings-ui/blob/main/src/mixins/allowable-actions-mixin.ts

legal-api/src/legal_api/services/authz.py
severinbeauvais: I agree with Lekshmi. This is essentially a 3-dimensional array of values but it feels like it might not easily extensible currently -- what if we add states (eg, NIGS) or other conditions?

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -55,6 +56,11 @@ def get_businesses(identifier: str):
     if recent_filing_json:
         business_json['submitter'] = recent_filing_json['filing']['header']['submitter']
         business_json['lastModified'] = recent_filing_json['filing']['header']['date']
+
+    allowed_filings = str(request.args.get('allowed_filings', None)).lower() == 'true'
```
severinbeauvais, line 60: I think spinal-case (aka kebab-case) is the recommendation for URL parameters, but check if we already use a specific convention.

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -404,17 +406,26 @@ def put_basic_checks(identifier, filing_id, client_request) -> Tuple[dict, int]:
...
     @staticmethod
-    def check_authorization(identifier, filing_json: str) -> Tuple[dict, int]:
+    def check_authorization(identifier, filing_json: str, business: Business) -> Tuple[dict, int]:
         """Assert that the user can access the business."""
-        action = ['edit']
         filing_type = filing_json['filing']['header'].get('name')
+        sub_filing_type = None
+        if filing_type == 'restoration':
+            sub_filing_type = filing_json['filing'].get('restoration', {}).get('type')
+        state = business.state if business else Business.State.ACTIVE
```
severinbeauvais, line 415: Add a comment explaining why you are assigning Active.

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -404,17 +406,26 @@ def put_basic_checks(identifier, filing_id, client_request) -> Tuple[dict, int]:
...
+        legal_type = business.legal_type if business else filing_json['filing']['business'].get('legalType')
+        admin_freeze = business.admin_freeze if business else False
```
severinbeauvais, line 417: Add a comment explaining why you are assigning False.

legal-api/src/legal_api/services/authz.py
```diff
@@ -88,3 +90,80 @@ def has_roles(jwt: JwtManager, roles: List[str]) -> bool:
     if jwt.validate_roles(roles):
         return True
     return False
+
+
+ALLOWABLE_FILINGS: Final = {
```
severinbeauvais, line 95: I wonder if this should be external to the code, say, in a db table, so that it could it be updated without a new deployment ?

thorwolpert (approved):

PR #1388, opened 2021-12-30
10557 name change from meta_data

vysakh-menon-aot:
*Issue #:* /bcgov/entity#10557

*Description of changes:*


lekshmimallika-aot (commented):
In _format_alteration_data, there is a code like
 **versioned_business = VersionedBusinessDetailsService.get_business_revision_before_filing(
            self._filing.id, self._business.id)
prev_legal_type = versioned_business['legalType']** 
We dont need any of those. Could you please correct that code as well and the corresponding part in the alteration notice template? I forgot to add this to the ticket

lekshmimallika-aot (approved):

PR #1389, opened 2022-01-05
Business Summary end point + basic framework + summary url in busines…

lekshmimallika-aot:
…s object

*Issue #:* /bcgov/entity#4463

*Description of changes:*
Include business summary url in the business json
New endpoint to download business summary
A new class for processing business summary as the current report class mainly deals with filings and already has lot of conditions to handle reports for different filing types.
Test updates

### DIFF
--- a/legal-api/report-templates/template-parts/common/businessDetails.html
+++ b/legal-api/report-templates/template-parts/common/businessDetails.html
@@ -1,7 +1,7 @@
 <div class="business-details-container">
    <table class="business-details-table" role="presentation">
       <tr>
-         {% if reportType == 'businessSummary' %}
+         {% if reportType == 'summary' %}
              <td>
                  <div class="bold">Incorporation Number:</div>
                  <div class="bold pt-2">Business Number:</div>
@@ -172,7 +172,7 @@
 </div>
 <!-- this block is hidden but is needed for the footer: -->
 <div class="filing-identifier">
-    {% if reportType == 'businessSummary' %}
+    {% if reportType == 'summary' %}
           <span>Incorporation #{{business.identifier}}</span>
     {% else %}
          Filing #{{header.filingId}}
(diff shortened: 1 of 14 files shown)


vysakh-menon-aot (approved):

legal-api/src/legal_api/models/business.py
```diff
@@ -335,9 +336,8 @@ def json(self):
         if self.tax_id:
             d['taxId'] = self.tax_id
         if self.state_filing_id:
-            base_url = current_app.config.get('LEGAL_API_BASE_URL')
             d['stateFiling'] = f'{base_url}/{self.identifier}/filings/{self.state_filing_id}'
-
+        d['summary'] = f'{base_url}/{self.identifier}/summary'
```
thorwolpert, line 340: should the summary be in the documents listing?
would it be an alternate view of the /businesses/identifier?output=summary as the default output would be json?

thorwolpert (commented):
question on URL pattern layout

legal-api/src/legal_api/models/business.py
lekshmimallika-aot: It need not be. The summary had  more information like all the alterations that happened previously. We dont get from any of the existing endpoints. All other info can be retrieved from existing endpoints.

I will move the code in the new endpoint to the existing businesses endpoint and delete the new endpoint + remove thhe url from the business json

legal-api/report-templates/template-parts/common/businessDetails.html
```diff
@@ -1,7 +1,30 @@
 <div class="business-details-container">
    <table class="business-details-table" role="presentation">
       <tr>
-         {% if header.reportType == 'noticeOfArticles' %}
+         {% if reportType == 'businessSummary' %}
```
argush3, line 4: should this be `header.reportType` as opposed to `reportType`?

argush3: nvm, i see where this comes from(`BusinessSummary._get_template_data` function) now.

argush3 (approved):

thorwolpert (approved):

PR #1400, opened 2022-01-19
10703 minimal Registration filing

vysakh-menon-aot:
*Issue #:* /bcgov/entity#10703

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/models/filing.py
+++ b/legal-api/src/legal_api/models/filing.py
@@ -134,7 +134,7 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
                 'CC': 'BCINC',
                 'CP': 'OTINC'
             },
-            'temperoryCorpTypeCode': 'TMP'
+            'temporaryCorpTypeCode': 'TMP'
         },
         'registration': {
             'name': 'registration',
@@ -143,7 +143,7 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
                 'SP': 'FRREG',
                 'GP': 'FRREG'
             },
-            'temperoryCorpTypeCode': 'RTMP'
+            'temporaryCorpTypeCode': 'RTMP'
         },
         'conversion': {
             'name': 'conversion',
--- a/legal-api/src/legal_api/models/filing.py
+++ b/legal-api/src/legal_api/models/filing.py
@@ -133,7 +133,17 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
                 'ULC': 'BCINC',
                 'CC': 'BCINC',
                 'CP': 'OTINC'
-            }
+            },
+            'temperoryCorpTypeCode': 'TMP'
+        },
+        'registration': {
+            'name': 'registration',
+            'title': 'Registration',
+            'codes': {
+                'SP': 'FRREG',
+                'GP': 'FRREG'
+            },
+            'temperoryCorpTypeCode': 'RTMP'
         },
         'conversion': {
             'name': 'conversion',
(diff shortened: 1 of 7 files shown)


legal-api/src/legal_api/models/filing.py
```diff
@@ -133,7 +133,17 @@ class Source(Enum):
                 'ULC': 'BCINC',
                 'CC': 'BCINC',
                 'CP': 'OTINC'
-            }
+            },
+            'temperoryCorpTypeCode': 'TMP'
```
thorwolpert, line 137: temporary is spelled wrong
What is this for? We currently have a bootstrap for businesses until they are created, is this for a new flow?

legal-api/src/legal_api/services/bootstrap.py
```diff
@@ -61,15 +61,17 @@ def delete_bootstrap(bootstrap: RegistrationBootstrap):
         return HTTPStatus.OK
 
     @staticmethod
-    def register_bootstrap(bootstrap: RegistrationBootstrap, business_name: str) \
-            -> Union[HTTPStatus, Dict]:
+    def register_bootstrap(bootstrap: RegistrationBootstrap,
+                           business_name: str,
+                           corp_type_code: str = 'TMP') -> Union[HTTPStatus, Dict]:
```
thorwolpert, line 66: Can we add an Enum (using the new BaseEnum) and add the corp_type for the BEN, CP, etc?
Or do we need it to fall under the act? Firm, Business, Other?

thorwolpert (commented):
Some questions around corp_type usage

legal-api/src/legal_api/services/bootstrap.py
vysakh-menon-aot: We don't have that flexibility since the new code has to be introduced in auth-api (corp_types table), can we create an Enum against filing type (incorporationApplication, registration) so that we can extend the current implementation.

legal-api/src/legal_api/models/filing.py
vysakh-menon-aot: Introduced here since this code is specific for incorporationApplication. We created a new ticket #10737 to support new temporary code for registration filing

thorwolpert: i think you can still use the legalType, I don't think we need another temp type.
we could create another range in the bootstrap if needed.

thorwolpert (commented):
still think it fits in the current schema, and should leverage existing legalType

legal-api/src/legal_api/models/filing.py
```diff
@@ -133,7 +133,17 @@ class Source(Enum):
...
+            'temporaryCorpTypeCode': 'TMP'
+        },
+        'registration': {
+            'name': 'registration',
+            'title': 'Registration',
+            'codes': {
+                'SP': 'FRREG',
+                'GP': 'FRREG'
+            },
+            'temporaryCorpTypeCode': 'RTMP'
```
thorwolpert, line 146: sorry, I'm missing why we need a temp type code as a business type.

thorwolpert (commented):
i don't get the need for a temp code, an incorporation or registration are for non-existing entities so what does adding temp to that help with?

legal-api/src/legal_api/models/filing.py
vysakh-menon-aot: auth-web use this code to differentiate a draft IA or Registration. Until now we had only draft IA listing in my business registry dashboard, so when auth-web receives the code TMP it identifies as draft IA. I have an open pr to update auth-web according to this https://github.com/bcgov/sbc-auth/pull/1905

thorwolpert (approved):

PR #1418, opened 2022-01-28
9250 registration validation

vysakh-menon-aot:
*Issue #:* /bcgov/entity#9250

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/registration.py
+++ b/legal-api/src/legal_api/services/filings/validations/registration.py
@@ -27,7 +27,7 @@ from ...utils import get_date, get_str
 from .common_validations import validate_court_order
 
 
-def validate(registration_json: Dict) -> Optional[list]:
+def validate(registration_json: Dict) -> Optional[Error]:
     """Validate the Registration filing."""
     if not registration_json:
         return Error(HTTPStatus.BAD_REQUEST, [{'error': babel('A valid filing is required.')}])
--- a/legal-api/src/legal_api/services/filings/validations/registration.py
+++ b/legal-api/src/legal_api/services/filings/validations/registration.py
@@ -1,0 +1,132 @@
+# Copyright © 2022 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Validation for the Registration filing."""
+from datetime import timedelta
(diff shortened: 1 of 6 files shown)


legal-api/src/legal_api/services/filings/validations/registration.py
```diff
@@ -0,0 +1,131 @@
...
+    """Validate the Registration filing."""
+    if not registration_json:
+        return Error(HTTPStatus.BAD_REQUEST, [{'error': babel('A valid filing is required.')}])
+    legal_type = get_str(registration_json, '/filing/business/legalType')
+    msg = []
+
+    msg.extend(validate_business_type(registration_json, legal_type))
+    msg.extend(validate_party(registration_json, legal_type))
+    msg.extend(validate_start_date(registration_json))
+    msg.extend(validate_delivery_address(registration_json))
```
lekshmimallika-aot, line 40: Shouldnt there be a name request validation?

lekshmimallika-aot (commented):
Do we have a set of mandatory attributes like incorporation number for an organization? If so, that needs to get validated I think

Are we planning to add any NAICS validation here

legal-api/src/legal_api/services/filings/validations/registration.py
vysakh-menon-aot: To verify if its can be consumed? I did not find that in IA validation. I was referring to that since both are creating a new business.

vysakh-menon-aot:
> Do we have a set of mandatory attributes like incorporation number for an organization? If so, that needs to get validated I think
> 
> Are we planning to add any NAICS validation here

It's not mandatory. Confirmed with Linda

legal-api/src/legal_api/services/filings/validations/registration.py
lekshmimallika-aot: Idk why it is missing from IA. But it has to be validated. Things that needs to get validated are 
1. whether it can be consumed
2. Legal type is the same

argush3: we should probably create a ticket for IA processor to make sure this is done too.

legal-api/src/legal_api/services/filings/validations/registration.py
```diff
@@ -0,0 +1,131 @@
...
+
+from ...utils import get_date, get_str
+from .common_validations import validate_court_order
+
+
+def validate(registration_json: Dict) -> Optional[Error]:
+    """Validate the Registration filing."""
+    if not registration_json:
+        return Error(HTTPStatus.BAD_REQUEST, [{'error': babel('A valid filing is required.')}])
+    legal_type = get_str(registration_json, '/filing/business/legalType')
```
argush3, line 34: I just noticed in the registration schema that we don't validate the business block like we do for IA.  I think we need to add this?  Even there, the business block only requires the `identifier`.  We may need to throw a validation error here if no legal type is provided I think.

https://github.com/bcgov/business-schemas/blob/394735852686aab57cf405def29433b5b5dbea01/src/registry_schemas/schemas/filing.json#L211-L228

lekshmimallika-aot (approved):

argush3 (approved):
looks good

PR #1421, opened 2022-02-02
10311 before expiry set datetime correctly

ozamani9:
*Issue #:* /bcgov/entity#10311

*Description of changes:*

Before Expiry email display UTC time, set it to actual time


argush3 (changes requested):
The change looks good but the before expiry tests are failing and could you add the same kind of expiry date test for the expired scenario.

https://github.com/bcgov/lear/blob/56cf6b91d2543008c28c3f868d142397d9feab33/queue_services/entity-emailer/tests/unit/email_processors/test_nr_notification.py#L80-L88

ozamani9:
will fix the pytest errors for sure.. but i dont use expiration date in the "EXPIRED" email at all.  its only used in BEFORE EXPIRY

argush3:
> will fix the pytest errors for sure.. but i dont use expiration date in the "EXPIRED" email at all. its only used in BEFORE EXPIRY

oh, ok.  Then yes, we just need to fix the failing tests for before expiry email then.

argush3 (approved):

PR #1432, opened 2022-02-18
9253 Create Registration statement output

vysakh-menon-aot:
*Issue #:* /bcgov/entity#9253

*Description of changes:*


argush3 (approved):

legal-api/migrations/versions/94586877d4bc.py
```diff
@@ -0,0 +1,29 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '94586877d4bc'
+down_revision = '9147d4bc9939'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
lekshmimallika-aot, line 20: Please check with Thor whether we add dml as part of the migration or it has to be a sql script on its own

thorwolpert: Seems small enough that we could leave it in.

We may want to cut a major version and redo our migration scripts to be just 1 new starting file. Probably something for us to huddle on.

legal-api/migrations/versions/94586877d4bc.py
```diff
@@ -0,0 +1,29 @@
...
+    op.execute("update corp_types set full_desc='BC Sole Proprietorship' where corp_type_cd = 'SP'")
+    op.execute("update corp_types set full_desc='BC General Partnership' where corp_type_cd = 'GP'")
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
```
thorwolpert, line 28: should we note that we accept this is not downgradeable?

thorwolpert (approved):

lekshmimallika-aot (approved):

PR #1450, opened 2022-03-11
11460 Update NAICS search logic to handle non-exact match scenarios

argush3:
*Issue #:* /bcgov/entity#11460

*Description of changes:*

* Update NAICS search logic to handle non-exact match scenarios which were previously returning no results
* Updated existing resource and model unit tests
* Added exact and non-exact match resource and model unit tests
* Updated element type to return string desc on GETs
* Removed `migrations/bulk_data` folder


argush3:
Not going to address the failing minio tests.

lekshmimallika-aot (approved):

vysakh-menon-aot (approved):

PR #1451, opened 2022-03-14
11385 naics service (api call) + validation

vysakh-menon-aot:
*Issue #:* /bcgov/entity#11385

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/registration.py
+++ b/legal-api/src/legal_api/services/filings/validations/registration.py
@@ -88,8 +88,11 @@ def validate_naics(filing: Dict, filing_type='registration') -> list:
     """Validate naics."""
     msg = []
     naics_code_path = f'/filing/{filing_type}/business/naics/naicsCode'
-    if (naics_code := get_str(filing, naics_code_path)) and not NaicsService.find_by_code(naics_code):
-        msg.append({'error': 'Invalid naics code.', 'path': naics_code_path})
+    naics_desc = get_str(filing, f'/filing/{filing_type}/business/naics/naicsDescription')
+    if naics_code := get_str(filing, naics_code_path):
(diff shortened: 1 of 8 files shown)


legal-api/src/legal_api/services/naics.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class NaicsService:
+    """Provides services to use the naics-api."""
+
+    @staticmethod
+    def find_by_code(naics_code: str):
+        """Return NAICS Structure matching code."""
+        naics_url = current_app.config.get('NAICS_API_URL')
+        token = AccountService.get_bearer_token()
```
vysakh-menon-aot, line 31: Using `system` token so that this can be called from both api and filer

lekshmimallika-aot (approved):

argush3:
the filer changes will be in a different PR right?

legal-api/src/legal_api/services/filings/validations/registration.py
```diff
@@ -83,6 +84,16 @@ def validate_business_type(filing: Dict, legal_type: str) -> list:
     return msg
 
 
+def validate_naics(filing: Dict, filing_type='registration') -> list:
+    """Validate naics."""
+    msg = []
+    naics_code_path = f'/filing/{filing_type}/business/naics/naicsCode'
+    if (naics_code := get_str(filing, naics_code_path)) and not NaicsService.find_by_code(naics_code):
```
argush3, line 91: I think we should validation that the description in the filing matches up to the value pulled back from the NAICS endpoint too.

vysakh-menon-aot: ok will do that

argush3 (approved):

PR #1455, opened 2022-03-15
11385 naics api call - filer

vysakh-menon-aot:
*Issue #:* /bcgov/entity#11385

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/services/naics.py
+++ b/legal-api/src/legal_api/services/naics.py
@@ -40,5 +40,6 @@ class NaicsService:
             else:
                 return None
         except Exception as err:
+            current_app.logger.error(naics_url)
             current_app.logger.error(err)
             raise err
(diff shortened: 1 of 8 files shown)


legal-api/src/legal_api/services/naics.py
```diff
@@ -27,14 +27,18 @@ class NaicsService:
...
-            'Content-Type': 'application/json',
-            'Authorization': 'Bearer ' + token
-        })
-
-        if response.ok:
-            return response.json()
-        else:
-            return None
+        try:
+            naics_url = current_app.config.get('NAICS_API_URL')
```
argush3, line 31: can we add a log statement to print out the `naics_url` here?  just want to be sure the service has the right url.

lekshmimallika-aot (approved):

argush3 (approved):

PR #1469, opened 2022-03-29
11625 Add COOP change of name special resolution notebook

argush3:
*Issue #:* /bcgov/entity#11625

*Description of changes:*

* Added notebook(`/tests/data/special_resolution/XXXX_create_coop_change_of_name_sr_XXXXX.ipynb`) for COOP change of name special resolution
* Updated common notebooks to support functionality in newly added notebook
* Updated create NR notebook to support COOP change of name NRs


vysakh-menon-aot (approved):

lekshmimallika-aot:
Looks good.
Do we support staff no fee option in this ?

lekshmimallika-aot (approved):

argush3:
> Looks good. Do we support staff no fee option in this ?

I don't think so for now.  The scenario we are supporting right now is specific to the ops ticket which only requires payment with a routing slip.

thorwolpert (approved):
looks ok to me!

PR #1473, opened 2022-03-30
Changes for SP/GP dissolutions

saravanpa-aot:
*Issue #:* https://github.com/bcgov/entity/issues/11635

*Description of changes:*

Changes for SP/GP dissolutions


vysakh-menon-aot (approved):

argush3 (changes requested):
looks good.  but can we update the `test_dissolution_type` test with parameters for SP and GP?  

https://github.com/bcgov/lear/blob/0f09b186c531adeeb22635e23ca18a0885ba972d/legal-api/tests/unit/services/filings/validations/test_dissolution.py#L32-L47

PR #1485, opened 2022-04-11
Changes in parties endpoint response

lekshmimallika-aot:
*Issue #:* /bcgov/entity#11799

*Description of changes:*

- Changed the role structure in parties endpoint response to match the schema
- Added email to the party json
- Updated tests


vysakh-menon-aot (approved):

vysakh-menon-aot:
Do you think its good to have the same in directors endpoint (v1 and v2)? It may not be required since it return only returning directors.

argush3 (approved):

lekshmimallika-aot:
> Do you think its good to have the same in directors endpoint (v1 and v2)? It may not be required since it return only returning directors.

Ya, As it returns only directors I don't think we need it in v1

PR #1507, opened 2022-05-17
11409 - Add a new attribute in backend to check if draft was save once [DRAFT]

ritvick:
*Issue #:* /bcgov/entity###
https://github.com/bcgov/entity/issues/11409

*Description of changes:*
Add a new column in filings to check if the draft was atleast saved once.


thorwolpert (commented):
I don't think this is needed.
It is already covered by the fact that there is a filing in the Draft state.
The v1 endpoints are legacy and should no longer be used.

ritvick:
@thorwolpert we have a new requirement to delete a draft if it was not intentionaly saved by the user once. 
Me Jyoti and Faiz had a discussion, and Faiz suggested to add a new column for this.

thorwolpert:
I don't think this is needed, and the linting/tests are failing. It has been sitting here for a while, so I'll close it for now.

PR #1552, opened 2022-06-13
Feature 12092 - Support for putback on filer changes

ritvick:
*Issue #:* /bcgov/entity###
https://github.com/bcgov/entity/issues/12094

*Description of changes:*
entity-filer changes to process putBackOn filing

### DIFF
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/put_back_on.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/put_back_on.py
@@ -15,6 +15,7 @@
 
 import dpath
 from entity_queue_common.service_utils import QueueException, logger
+from legal_api.models import Business, Document, Filing
 from contextlib import suppress
 from datetime import datetime
 from typing import Dict
@@ -24,7 +25,7 @@ from legal_api.models import Filing
 
 from entity_filer.filing_meta import FilingMeta
 
-def process(put_back_on_filing: Filing,  filing: Dict, filing_rec: Filing, filing_meta: FilingMeta):
+def process(business: Business,  filing: Dict, filing_rec: Filing, filing_meta: FilingMeta):
     """Render the put back on filing unto the model objects."""
     if not (put_back_on_filing := filing.get('putBackOn')):
         logger.error('Could not find putBackOn in: %s', filing)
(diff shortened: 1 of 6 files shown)


queue_services/entity-filer/src/entity_filer/filing_processors/put_back_on.py
```diff
@@ -0,0 +1,48 @@
...
+        logger.error('Could not find putBackOn in: %s', filing)
+        raise QueueException(f'legal_filing:putBackOn missing from {filing}')
+
+    logger.debug('processing putBackOn: %s', filing)
+
+    filing_meta.dissolution = {}
+    with suppress(IndexError, KeyError, TypeError):
+        put_back_on_details = dpath.util.get(filing, '/putBackOn/details')
+        filing_meta.put_back_on = {**filing_meta.put_back_on,
+                                   **{'details': put_back_on_details}}
```
vysakh-menon-aot, line 40: I don't think details is required in meta_data

ritvick: Removed

argush3: this might not belong in filing_meta but I think we need to save this somewhere.  Do we know if the UI uses this for display purposes?

queue_services/entity-filer/src/entity_filer/filing_processors/put_back_on.py
```diff
@@ -0,0 +1,41 @@
...
+
+    logger.debug('processing putBackOn: %s', filing)
+
+    # update court order, if any is present
+    with suppress(IndexError, KeyError, TypeError):
+        court_order_json = dpath.util.get(put_back_on_filing, '/courtOrder')
+        filings.update_filing_court_order(filing_rec, court_order_json)
+
+    business.state = Business.State.ACTIVE
+    business.state_filing_id = filing_rec.id
```
lekshmimallika-aot, line 41: So this was basically to identify the filing that moved the business to historical state or liquidation. I dont
think this should be set to the id of the put back filing. I think this should be set to null

queue_services/entity-filer/src/entity_filer/filing_processors/put_back_on.py
vysakh-menon-aot: yes. since this filing seems to be same as `registrarsNotation` and `registrarsOrder`. if its completely a different thing then this can either be a filing comments or a new field in filing table

lekshmimallika-aot (approved):

argush3 (approved):

PR #1561, opened 2022-06-16
Feature 12614 - Support for NOFEE code for admin dissolution

ritvick:
*Issue #:* /bcgov/entity###
https://github.com/bcgov/entity/issues/12614

*Description of changes:*
Support for No Fee code for admin dissolution

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -646,6 +646,12 @@ class ListFilingResource():
             for k in filing_json['filing'].keys():
                 filing_type_code = Filing.FILINGS.get(k, {}).get('codes', {}).get(legal_type)
                 priority = priority_flag
+                
+                #check if it is administrative dissolution, then pass NOFEE code
+                if k == 'dissolution':
+                    dissolutionType = filing_json['filing'][k].get('dissolutionType')
+                    if dissolutionType == 'administrative':
+                        filing_type_code = 'NOFEE'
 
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -647,6 +647,12 @@ def get_filing_types(business: Business, filing_json: dict):
                 filing_type_code = Filing.FILINGS.get(k, {}).get('codes', {}).get(legal_type)
                 priority = priority_flag
 
+                #check if it is administrative dissolution, then pass NOFEE code
+                if k == 'dissolution':
+                    dissolution_type = filing_json['filing'][k].get('dissolutionType')
+                    if dissolution_type == 'administrative':
+                        filing_type_code = 'NOFEE'
+
```
argush3, line 655: I don't really like this because it feels like it breaks our pattern of using the `FILINGS` dictionary in the filings model to get the fee code.  But if we update that to support a sub type(i.e. dissolution_type), that will likely require testing all the voluntary dissolution logic still works.  

I am ok with this for now but we should add a TODO here to update the dissolution related code to support sub types.  I'll create a ticket for this.  What are your thoughts @lekshmimallika-aot and @vysakh-menon ?

lekshmimallika-aot: Totally agree!

ritvick: Added the ToDo with ticket reference

vysakh-menon-aot (approved):

PR #1570, opened 2022-06-27
12199 Update Doc List: Suppress output for Coop Annual Report Correction for all Coop records

lewischenstudio:
*Issue #:* [/bcgov/entity#12199](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12199)

*Description of changes:*
- suppress `Notice of Articles` item for "Correction - Annual Report" in Recent Filing History, for all of the following conditions
  - the filing type is `CP`
  - the display name is `Correction - Annual Report`
  - the document link is for `noticeOfArticles`

### DIFF
--- a/legal-api/src/legal_api/core/meta/filing.py
+++ b/legal-api/src/legal_api/core/meta/filing.py
@@ -153,7 +153,7 @@ FILINGS: Final = {
             'CP': 'CRCTN'
         },
         'additional': [
-            {'types': 'CP,BEN', 'outputs': ['noticeOfArticles', ]},
+            {'types': 'BEN', 'outputs': ['noticeOfArticles', ]},
         ]
     },
     'courtOrder': {
--- a/legal-api/tests/unit/resources/v2/test_business_filings/test_filing_documents.py
+++ b/legal-api/tests/unit/resources/v2/test_business_filings/test_filing_documents.py
@@ -226,8 +226,7 @@ del ALTERATION_WITHOUT_NR['nameRequest']['legalName']
      ),
     ('ben_correction_completed', 'BC7654321', Business.LegalTypes.BCOMP.value,
      'correction', CORRECTION, 'incorporationApplication', INCORPORATION, Filing.Status.COMPLETED,
-     {'documents': {'noticeOfArticles': f'{base_url}/api/v2/businesses/BC7654321/filings/1/documents/noticeOfArticles',
-                    'legalFilings': [
+     {'documents': {'legalFilings': [
                         {'correction': f'{base_url}/api/v2/businesses/BC7654321/filings/1/documents/correction'},
                         {'incorporationApplication': f'{base_url}/api/v2/businesses/BC7654321/filings/1/documents/incorporationApplication'},
                     ]
@@ -238,7 +237,6 @@ del ALTERATION_WITHOUT_NR['nameRequest']['legalName']
     ('ben_correction_with_nr_completed', 'BC7654321', Business.LegalTypes.BCOMP.value,
      'correction', CORRECTION, 'incorporationApplication', CORRECTED_INCORPORATION, Filing.Status.COMPLETED,
      {'documents': {'certificate': f'{base_url}/api/v2/businesses/BC7654321/filings/1/documents/certificate',
-                    'noticeOfArticles': f'{base_url}/api/v2/businesses/BC7654321/filings/1/documents/noticeOfArticles',
                     'legalFilings': [
                         {'correction': f'{base_url}/api/v2/businesses/BC7654321/filings/1/documents/correction'},
                         {'incorporationApplication': f'{base_url}/api/v2/businesses/BC7654321/filings/1/documents/incorporationApplication'},
--- a/legal-api/tests/unit/resources/v2/test_business_filings/test_filing_documents.py
+++ b/legal-api/tests/unit/resources/v2/test_business_filings/test_filing_documents.py
@@ -278,6 +278,15 @@ del ALTERATION_WITHOUT_NR['nameRequest']['legalName']
       },
      HTTPStatus.OK, None
      ),
+    ('cp_correction_ar', 'CP7654321', Business.LegalTypes.COOP.value,
+     'correction', CORRECTION_AR, None, None , Filing.Status.COMPLETED,
+     {'documents': {'legalFilings': [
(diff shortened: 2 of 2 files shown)


legal-api/src/legal_api/core/meta/filing.py
```diff
@@ -332,8 +332,13 @@ def alter_outputs(filing_type: str, filing_meta_data: dict, outputs: set):
...
+            filing_final = FILINGS.get(filing_type, {})
             if not filing_meta_data.get('correction', {}).get('toLegalName') and 'certificate' in outputs:
                 # For IA correction, certificate will be populated in get_all_outputs since
                 # legalFilings list contains correction and incorporationApplication
                 # and it should be removed if correction does not contain name change.
                 outputs.remove('certificate')
+            if filing_final.get('displayName') == 'Correction - Annual Report' and \
+                filing_final.get('codes') == 'CP' and 'noticeOfArticles' in outputs:
+                # For Coop filing, Correction - Annual Report and NOA, remove key: noticeOfArticles
+                outputs.remove('noticeOfArticles')
```
vysakh-menon-aot, line 344: This won't work. `displayName` in `filing_final` is always `Correction` and `codes` is not a string (`filing_final.get('codes') == 'CP'`). Checking display name is a bad idea. Look into this example https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/reports/report.py#L259 and get original_filing.filing_type

lewischenstudio: Thanks. I've updated the code using the example. I also modified the method `alter_outputs` to better use the original filing instead of `filing_type` and `meta_data`.

legal-api/src/legal_api/core/meta/filing.py
```diff
@@ -337,3 +339,7 @@ def alter_outputs(filing_type: str, filing_meta_data: dict, outputs: set):
                 # legalFilings list contains correction and incorporationApplication
                 # and it should be removed if correction does not contain name change.
                 outputs.remove('certificate')
+            if filing.get('filing', {}).get('business', {}).get('legalType') == 'CP' and \
+                    'annualReport' in filing_meta_data.get('legalFilings', []) and 'noticeOfArticles' in outputs:
```
argush3, line 343: i don't think you will be able to check for "annualReport" in `filing_meta_data` here.  From looking at the `filing.meta_data` field for filings in Dev and Prod, `annualReport` is not included in the `legalFilings` property.  I do see that for other corrections like an incorporationApplication its included.  Maybe this is a bug?  do you have any insight into this @vysakh-menon-aot @lekshmimallika-aot ?

Meta data in prod for AR correction filings.
image

We typically don't want to reference the `filing.filing_json` when we don't have to but we can check if the `correctedFilingType` is "annualReport" in the correction filing.  If we don't do this, we may have to do a datafix and potentially update the correction logic for ARs if it is the case that "annualReport" should be included in the "legalFilings" array in the filing meta data.

argush3: I was actually talking about using the `correctedFilingType` field in the `filing_json` to determine if the correction filing is for an AR.  

image

The `filing_json` field can be found in the `filings` model.  

image

But lets wait to see if @vysakh-menon-aot or @lekshmimallika-aot has any other thoughts first.  And whether what I mentioned in my initial comment about "annualReport" not being in the `legalFilings` array in the `filing.meta_data` is a bug.

argush3: In that case I think your original code of looking for "annualReport" in the `legalFilings` array would work @lewischenstudio .  We just need to do a data fix in prod and create a ticket to fix the AR correction bug to populate "annualReport" in `legalFilings` in the `filings.meta_data` field.

legal-api/src/legal_api/core/meta/filing.py
lekshmimallika-aot: Cant we just remove CP from the line 
{'types': 'CP,BEN', 'outputs': ['noticeOfArticles', ]}, for correction?

lekshmimallika-aot: I dont think notice of articles is valid for CP ? Needs a confirmation from Linda

lewischenstudio: @lekshmimallika-aot I agree. I don't find any NOA for CP in BE or FE either, but this ticket wants to suppress all NOA for correction, CP and annualReport.

legal-api/src/legal_api/core/meta/filing.py
argush3: @lewischenstudio We should do the datafix in prod and go with your initial solution as per my previous comment.  The filing_json structure is not guaranteed to be always the same so we want to avoid using it directly whenever possible.  

But before all of this, can you confirm with Linda regarding the question above from @lekshmimallika-aot?  about whether we even need any NOAs at all for CPs?  if not, then it sounds like we don't even need this check?

lewischenstudio: @argush3 @lekshmimallika-aot I've got an update from Linda. It is right that we want to suppress NOA from all COOP records if they belong to type "correction" and "annual report" filing.

argush3: Ok, I'll write the script to update data in prod and all the other environments.  I'll let you know when the data has been updated @lewischenstudio in the ticket.  

I think I'm good with what you have currently but can we pass the business legal type as a parameter for `alter_outputs` function instead of retrieving the business again?

lewischenstudio: Thanks Argus. Yes, I can pass the business legal type as a parameter for `alter_outputs`, but no other place is accessing `business` than the inside the condition `filing_type == 'correction'. If I declare this outside `alter_ouputs` and end up not using it in the if statement, then it's an extra step that's not used. That's why I use this only for the condition `filing_type == 'correction'`. 

Let me know if you still want to pass this parameter to `alter_outputs`, and I'll update the method.

argush3: yes, please pass the `legal_type` as a parameter for `alter_outputs`.  I understand what you are saying but it's not that much work to remove legal_type if we don't end up using it.

argush3 (approved):

vysakh-menon-aot (approved):

lekshmimallika-aot (approved):

PR #1580, opened 2022-07-09
12786 admin api endpoints

vysakh-menon-aot:
*Issue #:* /bcgov/entity#12786

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/models/user.py
+++ b/legal-api/src/legal_api/models/user.py
@@ -37,6 +37,7 @@ class UserRoles(BaseEnum):
         return name
     #pragma warning enable S5720; # noqa: E265
 
+    # pylint: disable=invalid-name
     admin_edit = auto()
     system = auto()
     staff = auto()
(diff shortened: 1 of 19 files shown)


legal-api/src/legal_api/models/user.py
```diff
@@ -29,6 +29,7 @@
 class UserRoles(str, Enum):
```
thorwolpert, line 29: Could we take the opportunity to move this to the new BaseEnum?

legal-api/src/legal_api/resources/v2/admin.py
```diff
@@ -0,0 +1,71 @@
...
+from flask_cors import cross_origin
+from sentry_sdk import capture_message
+
+from legal_api.models import Business, UserRoles
+from legal_api.services import queue
+from legal_api.utils.auth import jwt
+from legal_api.utils.datetime import datetime
+
+
+bp = Blueprint('ADMINISTRATIVE', __name__, url_prefix='/api/v2/admin')
```
thorwolpert, line 28: as this is only for BN management, could it be stamped with just that instead of the whole admin namespace?

thorwolpert (commented):
couple of small questions

lekshmimallika-aot (approved):

PR #1601, opened 2022-07-21
Turn Sentry Off by using SENTRY_ENABLE FLAG instead of SENTRY_DSN#12422

stevenc987:
Signed-off-by: Steven Chen <steven.chen@gov.bc.ca>

*Issue #:* /bcgov/entity12422

*Description of changes:*
Turn Sentry Off by using SENTRY_ENABLE FLAG instead of SENTRY_DSN

### DIFF
--- a/colin-api/src/colin_api/__init__.py
+++ b/colin-api/src/colin_api/__init__.py
@@ -39,14 +39,15 @@ def create_app(run_mode=os.getenv('FLASK_ENV', 'production')):
     app = Flask(__name__)
     app.config.from_object(config.CONFIGURATION[run_mode])
     # Configure Sentry
-    if app.config.get('SENTRY_DSN', None):
-        sentry_sdk.init(  # pylint: disable=abstract-class-instantiated
-            dsn=app.config.get('SENTRY_DSN'),
-            integrations=[FlaskIntegration()]
-        )
-    app.register_blueprint(API_BLUEPRINT)
-    app.register_blueprint(OPS_BLUEPRINT)
-    # setup_jwt_manager(app, jwt)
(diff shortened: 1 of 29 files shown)


colin-api/src/colin_api/__init__.py
```diff
@@ -39,14 +39,15 @@ def create_app(run_mode=os.getenv('FLASK_ENV', 'production')):
...
-    app.register_blueprint(API_BLUEPRINT)
-    app.register_blueprint(OPS_BLUEPRINT)
-    # setup_jwt_manager(app, jwt)
+    if str(app.config.get('SENTRY_ENABLE')).lower() == 'true':
+        if app.config.get('SENTRY_DSN', None):
+            sentry_sdk.init(  # pylint: disable=abstract-class-instantiated
+                dsn=app.config.get('SENTRY_DSN'),
+                integrations=[FlaskIntegration()]
+            )
+        app.register_blueprint(API_BLUEPRINT)
```
thorwolpert, line 48: This will also turn off the API
That isn't the same as the code above.

thorwolpert (commented):
This should not be an envar, but an ldflag

colin-api/src/colin_api/__init__.py
stevenc987: @thorwolpert Thanks for your review. I have updated the file colin_api/__init__.py  and checked it in. Could you please take a look?
Thanks

stevenc987:
@thorwolpert 
I think the Entity Emailer CI failure was not caused by my change.

argush3:
@stevenc987 based off of @thorwolpert 's comment, we want to use an LD flag to turn sentry on and off as opposed your current implementation which is doing it through the environment variables.

PR #1606, opened 2022-07-28
12895 - Add in associationType for business call for Coop Association Type

seeker25:
*Issue #:* /bcgov/entity#12895

*Description of changes:*
Add in   'associationType': self.association_type to business call. 

Example URL:
GET <lear dev>/api/v2/businesses/CP1002552

### DIFF
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -341,6 +341,7 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes
             'nextAnnualReport': LegislationDatetime.as_legislation_timezone_from_date(
                 self.next_anniversary
             ).astimezone(timezone.utc).isoformat(),
+            'associationType': self.association_type
         }
 
         if self.last_coa_date:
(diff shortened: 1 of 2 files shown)


seeker25:
Not exactly sure why the unit tests aren't running as part of the check.

vysakh-menon-aot (approved):

legal-api/src/legal_api/models/business.py
```diff
@@ -341,6 +341,7 @@ def json(self):
             'nextAnnualReport': LegislationDatetime.as_legislation_timezone_from_date(
                 self.next_anniversary
             ).astimezone(timezone.utc).isoformat(),
+            'associationType': self.association_type
```
lekshmimallika-aot, line 344: I think this is applicable only for coops. So not sure whether this should be in the response only if self.legal_Type=='CP'?

vysakh-menon-aot: Its applicable for `SP` as well

lekshmimallika-aot (approved):

PR #1650, opened 2022-08-26
Model changes

lekshmimallika-aot:
*Issue #:* /bcgov/entity###

*Description of changes:*


argush3 (approved):

legal-api/migrations/versions/37acd517fe98_.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('businesses', sa.Column('jurisdiction', sa.String(length=10), nullable=True))
+    op.add_column('businesses', sa.Column('foreign_identifier', sa.String(length=15), nullable=True))
+    op.add_column('businesses', sa.Column('foreign_legal_name', sa.String(length=1000), nullable=True))
+    op.add_column('businesses', sa.Column('foreign_legal_type', sa.String(length=10), nullable=True))
+    op.add_column('businesses', sa.Column('foreign_incorporation_date', sa.DateTime(timezone=True), nullable=True))
+    op.add_column('businesses_version', sa.Column('jurisdiction', sa.String(length=10), nullable=True))
```
thorwolpert, line 26: I think it would be clearer to add **foreign_** as a prefix
```suggestion
    op.add_column('businesses_version', sa.Column('foreign_jurisdiction', sa.String(length=30), nullable=True))
```

thorwolpert (commented):
I think the _foreign_ prefix should be on the jurisdiction as well

thorwolpert (approved):

PR #1668, opened 2022-09-09
13330 Update the Entity Filer to apply BEN IA correction changes

lewischenstudio:
*Issue #:* /bcgov/entity#13330

*Description of changes:*
- allowed `correct_business_data` to process 'BEN' legal type
- update `business.share_classes` and `business.resolutions` if "shareStructure" is found in "correction"
- update `business.restriction_ind` if "provisionsRemoved" is found in "correction"

### DIFF
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py
@@ -26,8 +26,8 @@ from entity_filer.filing_processors.filing_components import (
     create_party,
     create_role,
     filings,
-    update_address,
     shares,
+    update_address,
 )
 
 
@@ -79,7 +79,7 @@ def correct_business_data(business: Business, correction_filing_rec: Filing,  # 
         court_order_json = dpath.util.get(correction_filing, '/correction/courtOrder')
         filings.update_filing_court_order(correction_filing_rec, court_order_json)
 
-        # update business start date, if any is present
+    # update business start date, if any is present
     with suppress(IndexError, KeyError, TypeError):
         business_start_date = dpath.util.get(correction_filing, '/correction/startDate')
         if business_start_date:
@@ -88,13 +88,14 @@ def correct_business_data(business: Business, correction_filing_rec: Filing,  # 
     # update share structure and resolutions, if any
     with suppress(IndexError, KeyError, TypeError):
         share_structure = dpath.util.get(correction_filing, '/correction/shareStructure')
-        shares.update_share_structure(business, share_structure)
-    
+        shares.update_share_structure_correction(business, share_structure)
+
     # update provisionsRemoved, if any
     with suppress(IndexError, KeyError, TypeError):
         provisions_removed = dpath.util.get(correction_filing, '/correction/provisionsRemoved')
         if provisions_removed is not None:
             business.restriction_ind = not provisions_removed
+
 
 def update_parties(business: Business, parties: dict, correction_filing_rec: Filing):
     """Create a new party or get them if they already exist."""
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py
@@ -27,6 +27,7 @@ from entity_filer.filing_processors.filing_components import (
     create_role,
     filings,
     update_address,
+    shares,
 )
 
 
@@ -84,6 +85,16 @@ def correct_business_data(business: Business, correction_filing_rec: Filing,  # 
         if business_start_date:
             business.start_date = datetime.datetime.fromisoformat(business_start_date) + timedelta(hours=8)
 
+    # update share structure and resolutions, if any
+    with suppress(IndexError, KeyError, TypeError):
+        share_structure = dpath.util.get(correction_filing, '/correction/shareStructure')
+        shares.update_share_structure(business, share_structure)
+    
+    # update provisionsRemoved, if any
+    with suppress(IndexError, KeyError, TypeError):
+        provisions_removed = dpath.util.get(correction_filing, '/correction/provisionsRemoved')
+        if provisions_removed is not None:
+            business.restriction_ind = not provisions_removed
 
 def update_parties(business: Business, parties: dict, correction_filing_rec: Filing):
     """Create a new party or get them if they already exist."""
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
@@ -71,6 +71,79 @@ def update_share_structure(business: Business, share_structure: Dict) -> Optiona
     return err
 
 
+def update_share_structure_correction(business: Business, share_structure: Dict) -> Optional[List]:
+    """Manage the share structure for a business.
+
+    Assumption: The structure has already been validated, upon submission.
+
+    Other errors are recorded and will be managed out of band.
+    """
+    if not business or not share_structure:
+        # if nothing is passed in, we don't care and it's not an error
+        return None
+
+    err = []
+
+    inclusion_entries = []
+    exclusion_entries = []
+    # Delete the ones that are present in db but not in the json and create the ones in json but not in db.
+    if resolution_dates := share_structure.get('resolutionDates'):
+        # Two lists of dates in datetime format
+        business_dates = [item.resolution_date for item in business.resolutions]
+        parsed_dates = [parse(resolution_dt).date() for resolution_dt in resolution_dates]
+
+        # Dates in both db and json
+        inclusion_entries = [business.resolutions[index] for index, date in enumerate(business_dates)
+                             if date in parsed_dates]
+        if len(inclusion_entries) > 0:
+            business.resolutions = inclusion_entries
+
+        # Dates in json and not in db
+        exclusion_entries = [date for date in parsed_dates if date not in business_dates]
+
+        resolution_dates = exclusion_entries
+
+        for resolution_dt in resolution_dates:
+            try:
+                d = Resolution(
+                    resolution_date=resolution_dt,
+                    resolution_type=Resolution.ResolutionType.SPECIAL.value
+                )
+                business.resolutions.append(d)
+            except (ValueError, OverflowError):
+                err.append(
+                    {'error_code': 'FILER_INVALID_RESOLUTION_DATE',
+                     'error_message': f"Filer: invalid resolution date:'{resolution_dt}'"}
+                )
+
+    if share_classes := share_structure.get('shareClasses'):
+        # business share classes array of jsons
+        business_share_ids = [share_class.id for share_class in business.share_classes]
+        share_structure_ids = [share_class.get('id') for share_class in share_classes]
+
+        # Dates in both db and json
+        inclusion_entries = [share_class for share_class in business.share_classes
+                             if share_class.id in share_structure_ids]
+        if len(inclusion_entries) > 0:
+            business.share_classes = inclusion_entries
+
+        # Dates in json and not in db
+        exclusion_entries = [share_class for share_class in share_classes
+                             if share_class.get('id') not in business_share_ids]
+
+        try:
+            for share_class_info in exclusion_entries:
+                share_class = create_share_class(share_class_info)
+                business.share_classes.append(share_class)
+        except KeyError:
+            err.append(
+                {'error_code': 'FILER_UNABLE_TO_SAVE_SHARES',
+                 'error_message': f"Filer: unable to save new shares for :'{business.identifier}'"}
+            )
+
+    return err
+
+
 def delete_existing_shares(business: Business):
     """Delete the existing share classes and series for a business."""
     if existing_shares := business.share_classes.all():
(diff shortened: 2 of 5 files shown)


queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py
```diff
@@ -84,6 +85,16 @@ def correct_business_data(business: Business, correction_filing_rec: Filing,  #
...
+    # update share structure and resolutions, if any
+    with suppress(IndexError, KeyError, TypeError):
+        share_structure = dpath.util.get(correction_filing, '/correction/shareStructure')
+        shares.update_share_structure(business, share_structure)
+    
+    # update provisionsRemoved, if any
+    with suppress(IndexError, KeyError, TypeError):
+        provisions_removed = dpath.util.get(correction_filing, '/correction/provisionsRemoved')
+        if provisions_removed is not None:
+            business.restriction_ind = not provisions_removed
```
lewischenstudio, line 97: I am not sure whether `restriction_ind` is the correct attribute to update for “provisionsRemoved”, which is present in BEN IA corrections.

argush3: I think this is the right field.

argush3 (commented):
can you write some tests to test the correction of these two(share structure and provisions removed) updates?  I was thinking maybe since corrections are fairly similar across businesses now maybe we can just rename `test_correction_firms.py` to `test_correction_busineses` under the `test_worker` folder and just use that to test BENs too.  

What do you think @lekshmimallika-aot and @vysakh-menon-aot ?

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py
```diff
@@ -84,6 +85,16 @@ def correct_business_data(business: Business, correction_filing_rec: Filing,  #
         if business_start_date:
             business.start_date = datetime.datetime.fromisoformat(business_start_date) + timedelta(hours=8)
 
+    # update share structure and resolutions, if any
+    with suppress(IndexError, KeyError, TypeError):
+        share_structure = dpath.util.get(correction_filing, '/correction/shareStructure')
+        shares.update_share_structure(business, share_structure)
```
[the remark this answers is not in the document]
lewischenstudio: Sorry Lekshmi. I only found `update_share_structure` in the entity-filer project. Is there any other method I wasn't aware of?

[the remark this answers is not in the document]
lekshmimallika-aot: That is a new method that needs to be written by the person working on this ticket :)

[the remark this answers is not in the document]
lewischenstudio: I was implementing this new method that compares the differences, and I was a bit confused. Did you mean that we want to append only the differences to `business.resolutions` and not everything from `resolutionDates`? If `resolutionDates` is the same as `business.resolutions`, then the output will be an empty array. I'm not sure whether this will have any impact on other places.

lewischenstudio:
@argush3 @lekshmimallika-aot Can you take a look at why the testing fails? I've gone through both legal api and schemas projects, but I don't quite understand how this one fails. I think it's related to the schema, probably due to `legalType: 'BC' `, but I can't really tell exactly how.

argush3:
> @argush3 @lekshmimallika-aot Can you take a look at why the testing fails? I've gone through both legal api and schemas projects, but I don't quite understand how this one fails. I think it's related to the schema, probably due to `legalType: 'BC' `, but I can't really tell exactly how.

I noticed one issue with the BEN correction test and that is legal api is not referring to a schema version that you need.  Update it to the latest one in the following file.  It can be updated to 2.15.34.
https://github.com/bcgov/lear/blob/21f1a52a1fe03984a8e929523921cd9641813a38/legal-api/requirements.txt#L61

Also a heads up that one of the tests(tests/integration/test_worker_queue.py::test_cb_subscription_handler) is known to fail already.  We have a ticket to fix it.

lewischenstudio:
> > @argush3 @lekshmimallika-aot Can you take a look at why the testing fails? I've gone through both legal api and schemas projects, but I don't quite understand how this one fails. I think it's related to the schema, probably due to `legalType: 'BC' `, but I can't really tell exactly how.
> 
> I noticed one issue with the BEN correction test and that is legal api is not referring to a schema version that you need. Update it to the latest one in the following file. It can be updated to 2.15.34.
> 
> https://github.com/bcgov/lear/blob/21f1a52a1fe03984a8e929523921cd9641813a38/legal-api/requirements.txt#L61
> 
> Also a heads up that one of the tests(tests/integration/test_worker_queue.py::test_cb_subscription_handler) is known to fail already. We have a ticket to fix it.

Okay. I think I should update legal api for BEN IA correction before working on entity-filer. I found some issues related to showing reports for BEN IA correction in legal api. I'll work on it and ask for help if I get stuck. Thanks.

argush3:
oops, I meant to say update the filer requirements file.  https://github.com/bcgov/lear/blob/83ef3f05ef29e07833e84b99a5f9b4af474fad19/queue_services/entity-filer/requirements.txt#L26

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py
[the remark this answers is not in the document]
lekshmimallika-aot: For corrections we have to do a compare operation for the objects. If an object in the json has an id that means the object already exists and you need to update it. If there is no id you have to create it. If an object in db with a specific id is not present in the input json that has to be deleted from db. This can be used as a reference
https://github.com/bcgov/lear/blob/main/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py#L88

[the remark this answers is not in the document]
lewischenstudio: Sorry Lekshmi. I am confused how this is possible to update/create/delete the resolution dates because `shareStructure.resolutionDates` is an array of strings, which don't have an associated `id`. For `shareStructure.shareClasses`, the current implementation is working properly, because the frontend doesn't update this object. For `shareStructure.resolutionDates`, the current implementation treats it as an array of strings and append to `business.resolutions` using the model class `Resolution`. 

image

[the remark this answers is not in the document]
lekshmimallika-aot: For resolution dates you can delete the ones that are present in db but not in the json and create the ones in json but not in db.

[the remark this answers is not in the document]
lewischenstudio: Okay. Thanks. I will implement the method in this way. In order to do this, I use `business.resolutions` to check with the json. I will assume `business.resolutions` is an array of the class `Resolution`, from which I can call `resolution_date` from each instance.

lewischenstudio:
@lekshmimallika-aot @argush3 Please have a look at the unit test error. I've been trying to fix this using minor changes, but it seems we to use fixture for methods inside a class.

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/correction.py
```diff
@@ -78,12 +79,23 @@ def correct_business_data(business: Business, correction_filing_rec: Filing,  #
...
+    # update business start date, if any is present
     with suppress(IndexError, KeyError, TypeError):
         business_start_date = dpath.util.get(correction_filing, '/correction/startDate')
         if business_start_date:
             business.start_date = datetime.datetime.fromisoformat(business_start_date) + timedelta(hours=8)
 
+    # update share structure and resolutions, if any
+    with suppress(IndexError, KeyError, TypeError):
+        share_structure = dpath.util.get(correction_filing, '/correction/shareStructure')
+        shares.update_share_structure_ben_correction(business, share_structure)
```
argush3, line 91: can we rename this to `update_share_structure_correction`?  Other legal types will have share structures as well.

lewischenstudio: Yeah. Sure.

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
```diff
@@ -71,6 +71,73 @@ def update_share_structure(business: Business, share_structure: Dict) -> Optiona
...
+    if resolution_dates := share_structure.get('resolutionDates'):
+        # Two lists of dates in datetime format
+        business_dates = [item.resolution_date for item in business.resolutions]
+        parsed_dates = [parse(resolution_dt).date() for resolution_dt in resolution_dates]
+
+        # Dates in both db and json
+        inclusion_dates = [business.resolutions[index] for index, date in enumerate(business_dates)
+                           if date in parsed_dates]
+        if len(inclusion_dates) > 0:
+            business.resolutions = inclusion_dates
```
argush3, line 97: i'm not sure what happens when you do this with respect to the versioning history.  In particular I would like to know if the unmodified `resolutions` entries result in a new `resolutions_version`.  If you can confirm the unmodified entries don't result in new version records,  you can do it this way.

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
```diff
@@ -71,6 +71,73 @@ def update_share_structure(business: Business, share_structure: Dict) -> Optiona
...
+
+        # Dates in json and not in db
+        exclusion_dates = [date for date in parsed_dates if date not in business_dates]
+
+        resolution_dates = exclusion_dates if len(exclusion_dates) > 0 else parsed_dates
```
argush3, line 102: won't `exclusion_dates` be good enough to use for appending new `resolutions` entries to the `business` object? I'm not seeing the need for L102...maybe I'm missing something.

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
```diff
@@ -71,6 +71,73 @@ def update_share_structure(business: Business, share_structure: Dict) -> Optiona
...
+
+        try:
+            for share_class_info in share_classes:
+                share_class = create_share_class(share_class_info)
+                business.share_classes.append(share_class)
+        except KeyError:
+            err.append(
+                {'error_code': 'FILER_UNABLE_TO_SAVE_SHARES',
+                 'error_message': f"Filer: unable to save new shares for :'{business.identifier}'"}
+            )
```
argush3, line 136: As per Lekshimi's previous comment, this needs to actually only remove the share classes that have been removed and add the new ones.  This code is removing all share classes associated with the business and adding all old and new share classes from scratch.  It's problematic as we will not be able to re-construct the versioning history accurately.

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
lewischenstudio: You are right. `resolution_dates = exclusion_dates` is enough for the looping below. I'll change that.

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
lewischenstudio: Sure. I will implement this for `shareClass` in just in case this method is used later on.

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
argush3: As per our conversation, we will test this out in DEV.  If versioning of records in the db is not correct, another PR will  be created to fix this.

lewischenstudio: Sure. Thanks.

argush3 (approved):

vysakh-menon-aot (approved):

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py
```diff
@@ -71,6 +71,79 @@ def update_share_structure(business: Business, share_structure: Dict) -> Optiona
...
+    if share_classes := share_structure.get('shareClasses'):
+        # business share classes array of jsons
+        business_share_ids = [share_class.id for share_class in business.share_classes]
+        share_structure_ids = [share_class.get('id') for share_class in share_classes]
+
+        # Dates in both db and json
+        inclusion_entries = [share_class for share_class in business.share_classes
+                             if share_class.id in share_structure_ids]
+        if len(inclusion_entries) > 0:
+            business.share_classes = inclusion_entries
```
lekshmimallika-aot, line 128: Will this work if there is a new share series under an existing share class json?

lekshmimallika-aot:
@lewischenstudio We need different tests for share classes , Addition, Deletion, Modification, Removing just a series under share class, Adding a series under the share class etc. I see that this PR is merged now. You can add that in another PR

lewischenstudio:
@lekshmimallika-aot I tried to add tests in this PR for the ben correction, but it has been failing because of the schema exception.

argush3:
I just tested the fixes in this PR.  There are definitely some issues that we will need to resolve.  For example updates to anything in an existing share does not get picked up.  Don't think deletion works either but I need to double check.

argush3:
here's some info for what i'm seeing.  

**Corrections made to BEN:**
image

**LEAR db values after correction filing processed**
image

PR #1674, opened 2022-09-13
13330 Update the Legal API to apply BEN IA correction changes 

lewischenstudio:
*Issue #:* /bcgov/entity#13330

*Description of changes:*
- updated schemas version in requirements.txt
- updated lega-api version

### DIFF
--- a/legal-api/src/legal_api/version.py
+++ b/legal-api/src/legal_api/version.py
@@ -22,4 +22,4 @@ Post-release segment: .postN
 Development release segment: .devN
 """
 
-__version__ = '2.58.1'  # pylint: disable=invalid-name
+__version__ = '2.58.0'  # pylint: disable=invalid-name
--- a/legal-api/src/legal_api/version.py
+++ b/legal-api/src/legal_api/version.py
@@ -22,4 +22,4 @@ Post-release segment: .postN
 Development release segment: .devN
 """
 
-__version__ = '2.58.0'  # pylint: disable=invalid-name
+__version__ = '2.58.1'  # pylint: disable=invalid-name
--- a/legal-api/requirements.txt
+++ b/legal-api/requirements.txt
@@ -34,7 +34,7 @@ flask-restx==0.3.0
 gunicorn==20.1.0
 idna==2.10
 itsdangerous==1.1.0
-jsonschema==3.2.0
+jsonschema==4.16.0
 launchdarkly-server-sdk==7.1.0
 protobuf==3.15.8
 psycopg2-binary==2.8.6
--- a/legal-api/requirements.txt
+++ b/legal-api/requirements.txt
@@ -58,5 +58,5 @@ urllib3==1.26.4
 minio==7.0.2
 PyPDF2==1.26.0
 reportlab==3.6.1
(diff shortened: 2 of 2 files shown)


argush3:
The report updates were actually intended to be done in #13504 which has a pretty large estimate attached to it.  If you intend to do this work, you should do it in #13504.  We should have taken some of todos in this(filer) ticket that already had separate tickets created.

In general, I think we want to move BEN corrections to use the same approach as what we are doing for SP/GPs.  There is some existing in the reporting logic that needs to be updated to follow model.  I've included some things that I have found from an initial look.  @lekshmimallika-aot @vysakh-menon-aot please chime in if I'm saying anything incorrect or if you have any specific thoughts.

I don't think the report key needs to be changed to contain the value of `correctedFilingType` anymore for a BEN anymore.
https://github.com/bcgov/lear/blob/9894a0929d2c191713fd361417ec843b7b1de30e/legal-api/src/legal_api/reports/report.py#L59-L64


`self._format_with_diff_data(filing)` is no longer required and can follow something more similar to what corrections for registrations is doing.  Perhaps update `_format_alteration_date` function...
https://github.com/bcgov/lear/blob/9894a0929d2c191713fd361417ec843b7b1de30e/legal-api/src/legal_api/reports/report.py#L233-L238

determine if changes need to happen here
https://github.com/bcgov/lear/blob/9894a0929d2c191713fd361417ec843b7b1de30e/legal-api/src/legal_api/reports/report.py#L883-L892

Is there a need to have a new output like how SP/GP has the correct registration statement?
https://github.com/bcgov/lear/blob/5712ddcaf899fb76554a9d4d193050b5de471cc2/legal-api/src/legal_api/core/meta/filing.py#L163-L166

lewischenstudio:
Thanks Argus. I have dropped the reports from this pull request since it will be handled in a larger ticket #13504. For this PR, I only updated the schemas version in the legal api.

argush3:
> The test failure is due to the recent change from `draft7_format_checker` to `Draft7Validator.FORMAT_CHECKER`. Maybe we should use `Draft7Validator.format_checker` since `AttributeError: type object 'Draft7Validator' has no attribute 'FORMAT_CHECKER'` and `format_checker` is a valid attribute?

The business schema was updated to use `jsonschema` version 4.16.0  - https://github.com/bcgov/business-schemas/pull/110.  

Try updating corresponding `jsonschema` entry in requirements.txt to use 4.16.0

https://github.com/bcgov/lear/blob/21f1a52a1fe03984a8e929523921cd9641813a38/legal-api/requirements.txt#L37

legal-api/src/legal_api/version.py
```diff
@@ -22,4 +22,4 @@
 Development release segment: .devN
 """
 
-__version__ = '2.58.0'  # pylint: disable=invalid-name
+__version__ = '2.58.1'  # pylint: disable=invalid-name
```
argush3, line 25: we usually only bump version numbers before a release.  please revert.

lewischenstudio: Okay. It's reverted.

argush3 (approved):

PR #1684, opened 2022-09-22
12421 legal-api - missing business info warning check returning warnings for deleted/ceased partners

lewischenstudio:
*Issue #:* /bcgov/entity#12421

*Description of changes:*
- added `else` condition for not checking deleted partners


argush3:
I don't believe the change you've made to the complete party is required.  What needs to be done is to update the `check_parties` function to filter `firm_party_roles` such that any party role pulled back doesn't have a cessation date.

https://github.com/bcgov/lear/blob/46fdad9e5f1c7d1007c55b72d6bc44a8d1979ee4/legal-api/src/legal_api/services/warnings/business/business_checks/firms.py#L84

PR #1711, opened 2022-10-07
13834 Add dissolution date to filing meta data

argush3:
*Issue #:* /bcgov/entity#13834

*Description of changes:*

* Update filer to add dissolution date to filing meta data
* Update filer test
* Update custom SP/GP pipeline filer to add dissolution date to filing meta data
* Update business endpoint to return dissolution date in same format as all other places.  i.e. only the date and not timestamp
* Update business json test

### DIFF
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """File processing rules and actions for Dissolution and Liquidation filings."""
 from contextlib import suppress
+from datetime import datetime
 from typing import Dict
 
 import dpath
@@ -39,10 +40,6 @@ def process(business: Business, filing: Dict, filing_rec: Filing, filing_meta: F
 
     filing_meta.dissolution = {}
     dissolution_type = dpath.util.get(filing, '/dissolution/dissolutionType')
-
-    with suppress(IndexError, KeyError, TypeError):
-        filing_meta.dissolution = {**filing_meta.dissolution,
-                                   **{'dissolutionType': dissolution_type}}
 
     # hasLiabilities can be derived from dissolutionStatementType
     # FUTURE: remove hasLiabilities from schema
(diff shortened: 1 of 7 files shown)


vysakh-menon-aot (approved):

severinbeauvais (approved):
Could there be a unit test to verify each place the date is formatted?

Should the code version be updated?

severinbeauvais:
Is that a known unit test error?

queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
```diff
@@ -78,6 +75,11 @@ def process(business: Business, filing: Dict, filing_rec: Filing, filing_meta: F
     if business.legal_type == Business.LegalTypes.COOP:
         _update_cooperative(dissolution_filing, business, filing_rec, dissolution_type)
 
+    with suppress(IndexError, KeyError, TypeError):
+        filing_meta.dissolution = {**filing_meta.dissolution,
+                                   **{'dissolutionType': dissolution_type},
+                                   **{'dissolutionDate': datetime.date(filing_rec.effective_date).isoformat()}}
```
lekshmimallika-aot, line 81: dissolution date is different from effective date for voluntary dissolution I think

argush3:
> Could there be a unit test to verify each place the date is formatted?
> 
> Should the code version be updated?

yes, i can add an assertion to verify the date format.  Will do that.

Code version is good.  We will update right before moving to test.

argush3:
> Is that a known unit test error?

yes, this is a known unit test error.  we'll need to get around to fixing it.  there is actually a ticket i think.

queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
lekshmimallika-aot: We can fix this in a follow up PR because there might be a gap in setting the dissolution date in the business too.

lekshmimallika-aot (approved):

PR #1720, opened 2022-10-13
13872 update business start date - change of registration

vysakh-menon-aot:
*Issue #:* /bcgov/entity#13872

*Description of changes:*


argush3 (approved):

lekshmimallika-aot (commented):
Change of registration was not supposed to correct the start date as per the requirement. Only correction could change the start date. The changes look good. But we need to confirm that this is a valid business requirement

vysakh-menon-aot:
Obsolete - Have confirmed with Linda that clients should not be allowed to change the Business start date.

PR #1774, opened 2022-10-30
Feature/13459 - New way ( non legacy way of validating con)

ritvick:
*Issue #:* /bcgov/entity/issues/13459

*Description of changes:*
Change of name in SR should be validated properly as per new schema -> NR request validation


legal-api/src/legal_api/services/filings/validations/change_of_name.py
```diff
@@ -13,26 +13,55 @@
...
     if not legal_name:
-        msg.append({'error': _('Legal Name must be provided.'),
+        msg.append({'error': babel('Legal Name must be provided.'),
                     'path': legal_name_path})
     if msg:
         return Error(HTTPStatus.BAD_REQUEST, msg)
     return None
+
+
+def validate_v2(business: Business, filing: Dict) -> Error:
```
argush3, line 42: can we update the existing change of name validation tests to test this new validation function?

argush3 (approved):

PR #1789, opened 2022-11-03
14097 update address fields to empty when none - reports

vysakh-menon-aot:
*Issue #:* /bcgov/entity#14097

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/reports/business_document.py
+++ b/legal-api/src/legal_api/reports/business_document.py
@@ -411,6 +411,10 @@ class BusinessDocument:
 
     @staticmethod
     def _format_address(address):
+        address['streetAddressAdditional'] = address.get('streetAddressAdditional') or ''
+        address['addressRegion'] = address.get('addressRegion') or ''
+        address['deliveryInstructions'] = address.get('deliveryInstructions') or ''
+
         country = address['addressCountry']
         country = pycountry.countries.search_fuzzy(country)[0].name
         address['addressCountry'] = country
--- a/legal-api/src/legal_api/services/business_details_version.py
+++ b/legal-api/src/legal_api/services/business_details_version.py
@@ -526,10 +526,10 @@ class VersionedBusinessDetailsService:  # pylint: disable=too-many-public-method
             country_description = pycountry.countries.search_fuzzy(address_revision.country)[0].name
         return {
             'streetAddress': address_revision.street,
-            'streetAddressAdditional': address_revision.street_additional,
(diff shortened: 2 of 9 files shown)


legal-api/src/legal_api/reports/business_document.py
```diff
@@ -411,6 +411,10 @@ def _set_liquidation_details(self, business: dict):
 
     @staticmethod
     def _format_address(address):
+        address['streetAddressAdditional'] = address.get('streetAddressAdditional') or ''
+        address['addressRegion'] = address.get('addressRegion') or ''
```
severinbeauvais, line 415: `addressRegion` is a [required property](https://github.com/bcgov/business-schemas/blob/35a99eabfa8b70527124f90c4b6807d159aea3b4/src/registry_schemas/schemas/address.json#L47) so you shouldn't need to do this. If you do then there's a data error(s).

vysakh-menon-aot: as per the schema it can be null https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/address.json#L23

severinbeauvais: Oh, so it has to be present but can be null? That sounds like "not required" to me, haha.

Any way we can update the schema to be consistent?

vysakh-menon-aot: One thing we could do is remove `addressRegion` from required and allow null in `streetAddressAdditional`

severinbeauvais: That makes sense to me.

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -526,10 +526,10 @@ def address_revision_json(address_revision):
             country_description = pycountry.countries.search_fuzzy(address_revision.country)[0].name
         return {
             'streetAddress': address_revision.street,
-            'streetAddressAdditional': address_revision.street_additional,
+            'streetAddressAdditional': address_revision.street_additional or '',
             'addressType': address_revision.address_type,
             'addressCity': address_revision.city,
-            'addressRegion': address_revision.region,
+            'addressRegion': address_revision.region or '',
```
severinbeauvais, line 532: Hooray for consistency :)

severinbeauvais (approved):
LGTM

argush3 (approved):

PR #1816, opened 2022-11-22
Add idp userid to users table.

pwei1018:
*Issue #:* /bcgov/entity###
https://github.com/bcgov/entity/issues/14367

*Description of changes:*
1. add idp_userid column to users table;


bolyachevets (approved):

legal-api/migrations/versions/1ba027b074a5_add_idp_userid_to_users.py
```diff
@@ -0,0 +1,32 @@
...
+# revision identifiers, used by Alembic.
+revision = '1ba027b074a5'
+down_revision = 'e10f724fad6d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('users', sa.Column('idp_userid', sa.String()))
```
argush3, line 21: can we update the column def to the following?  length 256 is what auth user table has. 

`op.add_column('users', sa.Column('idp_userid', sa.String(length=256), nullable=True)`

legal-api/migrations/versions/1ba027b074a5_add_idp_userid_to_users.py
```diff
@@ -0,0 +1,32 @@
...
+    #op.drop_constraint('users_sub_key', 'users', type_='unique')
```
seeker25, line 22: Any reason for the commented out code?

legal-api/migrations/versions/1ba027b074a5_add_idp_userid_to_users.py
```diff
@@ -0,0 +1,32 @@
...
+    #op.create_unique_constraint('users_idp_userid_key', 'users', ['idp_userid'])
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    #op.create_unique_constraint('users_sub_key', 'users', ['sub'])
+    #op.drop_constraint('users_idp_userid_key', 'users', ['idp_userid'])
```
seeker25, line 30: Any reason for the commented out code?

legal-api/migrations/versions/1ba027b074a5_add_idp_userid_to_users.py
pwei1018: I can update the length. And this column will be unique key later.

legal-api/migrations/versions/1ba027b074a5_add_idp_userid_to_users.py
pwei1018: This is just the first step. The constraint may stop the data loading step on the next.

legal-api/migrations/versions/1ba027b074a5_add_idp_userid_to_users.py
argush3: It's possible for it to be null though right?

pwei1018: no, it shouldn't be null. It has to have the value from Keycloak.

argush3: There is user data from COLIN where we don't have that info.  So it needs to be nullable I think.

pwei1018: What's information in the 'sub' column now of Colin user? We can copy the value from 'sub' to this column.

argush3: "sub" info was not available in COLIN db from what I could see.  This is the table where the user info came from.

image

pwei1018: Do you have example in prod or test database? I need to look at the data first.

argush3: Here are some examples in LEAR Prod db.

I think in a lot of instances, there was a userid match and that user was just used so there was a sub.  There are 172 user entries in Prod without a sub and I think a lot of those were brought over from COLIN.

```
+----+--------+---------+--------+-----+----+----+---------------------------------+
|id  |username|firstname|lastname|email|sub |iss |creation_date                    |
+----+--------+---------+--------+-----+----+----+---------------------------------+
|2358|SC70011 |         |        |     |null|null|1995-10-27 07:00:00.000000 +00:00|
|2095|PV81928 |         |        |     |null|null|2000-01-04 08:00:00.000000 +00:00|
|2096|PY75075 |         |        |     |null|null|2011-04-08 07:00:00.000000 +00:00|
|2097|PT81928 |         |        |     |null|null|2011-01-24 08:00:00.000000 +00:00|
+----+--------+---------+--------+-----+----+----+---------------------------------+
```

pwei1018: @bolyachevets thought?

pwei1018: @argush3 we should copy the username to the new column. I checked the existing database only few username got duplicated.

argush3: ok, sounds good

argush3 (approved):

PR #1817, opened 2022-11-22
14339 AR reminder for ULC, CCC, BC

vysakh-menon-aot:
*Issue #:* /bcgov/entity#14339

*Description of changes:*
  - included pagination 
  - feature flag enable-bc-ccc-ulc

### DIFF
--- a/jobs/email-reminder/email_reminder.py
+++ b/jobs/email-reminder/email_reminder.py
@@ -121,13 +121,16 @@ def get_businesses(legal_types: list):
         " END  + interval '1 year'= CURRENT_DATE")
     return db.session.query(Business).filter(
         Business.legal_type.in_(legal_types), where_clause
-    ).all()
+    ).order_by(Business.id).paginate(per_page=20)
 
 
 async def find_and_send_ar_reminder(app: Flask, qsm: QueueService):  # pylint: disable=redefined-outer-name
     """Find business to send annual report reminder."""
     try:
-        legal_types = [Business.LegalTypes.BCOMP.value]  # entity types to send ar reminder
+        legal_types = [Business.LegalTypes.BCOMP.value,
+                       Business.LegalTypes.COMP.value,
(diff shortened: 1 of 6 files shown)


lekshmimallika-aot (approved):

jobs/email-reminder/email_reminder.py
```diff
@@ -121,13 +121,16 @@ def get_businesses(legal_types: list):
...
 
 
 async def find_and_send_ar_reminder(app: Flask, qsm: QueueService):  # pylint: disable=redefined-outer-name
     """Find business to send annual report reminder."""
     try:
-        legal_types = [Business.LegalTypes.BCOMP.value]  # entity types to send ar reminder
+        legal_types = [Business.LegalTypes.BCOMP.value,
+                       Business.LegalTypes.COMP.value,
+                       Business.LegalTypes.BC_CCC.value,
+                       Business.LegalTypes.BC_ULC_COMPANY.value]  # entity types to send ar reminder
```
argush3, line 133: We may want to grab these values from a feature flag.  I was thinking that if any ULC/CC/BC data is loaded into Prod ahead of time AR reminders will start firing off which we don't want.  

What do you think?

vysakh-menon-aot: I think its nice to have at the moment since David confirmed the plan is to release everything together by March. Do you foresee a chance to load data to prod before?

argush3: I think I will need to load into Prod before the go live date which is what I did for SP/GPs.  Search needed the data in Prod and for the API sandbox.  

I guess we can also just do a quick test in Dev that this is working and just comment these out for now.

vysakh-menon-aot: In that case I will include it

argush3 (approved):

PR #1834, opened 2022-12-06
14528 Fix json serialization and partial update issues

argush3:
*Issue #:* /bcgov/entity#14528

*Description of changes:*


vysakh-menon-aot:
We do the similar update in entity-filer, example: https://github.com/bcgov/lear/blob/main/queue_services/entity-filer/src/entity_filer/filing_processors/incorporation_filing.py#L177 Do you think we have to update here as well?

vysakh-menon-aot:
> We do the similar update in entity-filer, example: https://github.com/bcgov/lear/blob/main/queue_services/entity-filer/src/entity_filer/filing_processors/incorporation_filing.py#L177 Do you think we have to update here as well?

I was wrong. We replace json here

PR #1850, opened 2022-12-08
use idp guid to uniquely identify users

bolyachevets:
*Issue #:* /bcgov/entity###
bcgov/entity#14596

*Description of changes:*

after keycloak migration to gold, sub from jwt token (i.e. keycloak guid) will be changed for users, as it is tied to a specific IDP instance, and idp guid should be used instead to uniquely identify users

### DIFF
--- a/legal-api/src/legal_api/models/user.py
+++ b/legal-api/src/legal_api/models/user.py
@@ -94,10 +94,6 @@ class User(db.Model):
     @classmethod
     def find_by_jwt_token(cls, token: dict):
         """Return a User if they exist and match the provided JWT."""
-        return cls.query.filter_by(sub=token['sub']).one_or_none()
-
-    @classmethod
-    def find_by_jwt_idp_userid(cls, token):
         return cls.query.filter_by(sub=token['idp_userid']).one_or_none()
 
     @classmethod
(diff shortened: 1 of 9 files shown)


pwei1018 (approved):

legal-api/src/legal_api/models/user.py
```diff
@@ -107,7 +109,9 @@ def create_from_jwt_token(cls, token: dict):
                 firstname=token.get(current_app.config.get('JWT_OIDC_FIRSTNAME'), None),
                 lastname=token.get(current_app.config.get('JWT_OIDC_LASTNAME'), None),
                 iss=token['iss'],
-                sub=token['sub']
+                sub=token['sub'],
+                idp_userid=token['idp_userid'],
+                login_source=token['loginSource']
```
argush3, line 114: I'm not sure why a lot of the tests are failing but maybe in the `helper_create_jwt` function of `legal-api/tests/unit/services/utils.py` you do not include the `loginSource` for the token.  Based on the column definition for `loginSource` in the `users` table, this seems like a valid scenario.  

If this is the source of the test errors, you will need to be able to handle an incoming token that has no `login_source` property.

bolyachevets: thank you - missed that. fixed

argush3: so the incoming token will always have the `login_source` property?

bolyachevets: currently loginSource is a 'hardcoded attribute' for idir/bcsc/bceid IDPs

argush3: ok, sounds like it will always be there so L114(`login_source=token['loginSource']`) won't have a scenario where it will error out due to the property not been available in the token.

argush3 (approved):

PR #1851, opened 2022-12-12
[Draft] Feature/13480 Backend for Freezing/Unfreezing an Entity

ritvick:
*Issue #:* /bcgov/entity/issues/13480

*Description of changes:*
Support for feeze and unfreeze


kialj876 (changes requested):
Hey Ritvick, this needs to ensure they are staff otherwise anyone with a valid token can freeze any business and it would be good to have tests in here that secure that.

Also I think this should be a patch endpoint since you are just updating one field of the existing record

argush3:
accidentally closed....  re-opened now.

so is there a need to drive freezing of an entity through a filing?

ritvick:
> accidentally closed.... re-opened now.
> 
> so is there a need to drive freezing of an entity through a filing?

https://app.zenhub.com/workspaces/relationships-team-space-61435c47e483c4000f08e9f6/issues/bcgov/entity/13348

As per comments in this ticket, this will not be treated as filing.

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -125,3 +125,22 @@ def post_businesses():
...
 
     return ListFilingResource.put(bootstrap.identifier, None)
+
+@bp.route('/<string:identifier>', methods=['PUT'])
+@cross_origin(origin='*')
+@jwt.requires_auth
+def update_businesses(identifier: str):
+    """Update a business."""
+
+    business = Business.find_by_identifier(identifier)
```
argush3, line 135: should return 404 if business not found

argush3:
please add/update tests

kialj876:
> > Hey Ritvick, this needs to ensure they are staff otherwise anyone with a valid token can freeze any business and it would be good to have tests in here that secure that.
> > Also I think this should be a patch endpoint since you are just updating one field of the existing record
> 
> Converted to patch

Great, can you check they are staff and return unauthorized if not too please?

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -129,3 +130,30 @@ def post_businesses():
...
+    if not business:
+        return jsonify({'message': f'{identifier} not found'}), HTTPStatus.NOT_FOUND
+
+    json_input = request.get_json()
+
+    admin_freeze = json_input['adminFreeze'] or False
+
+    business.admin_freeze = admin_freeze
+
+    business.save()
```
thorwolpert, line 156: Generally we don't change a businesses data through the API, it's done indirectly through a filing.
For instance when we apply a BN, a _Registrars Notation_ is added with a reason, so that all data changes are reflected in the ledger.
This breaks that whole pattern, and as such doesn't provide the customer with a reason it is frozen, or what they need to do next.

thorwolpert (changes requested):
This is not in keeping with the API pattern in place to date.
There needs to be some changes in the design on this.

legal-api/src/legal_api/resources/v2/business/business.py
ritvick: Hi @thorwolpert we discussed about this in grooming couple of times. The freeze and unfreeze option from what I understand is an operation which should not appear in ledger, also it is communicated it is not a filing too. 

Documented in comments here: https://app.zenhub.com/workspaces/relationships-team-space-61435c47e483c4000f08e9f6/issues/bcgov/entity/13348

Happy to chat about this in standup.

ritvick:
Closing this PR, and opened a new one https://github.com/bcgov/lear/pull/1865/files
For the change in design to support freeze and unfreeze

PR #1859, opened 2022-12-14
14257-Upgrade CI/CD.

mengdong19:
*Issue #:* /bcgov/entity/14257

*Description of changes:*
Upgrade codecov version and checkout action version,

### DIFF
--- a/.github/workflows/future-effective-filings-ci.yml
+++ b/.github/workflows/future-effective-filings-ci.yml
@@ -18,7 +18,7 @@ jobs:
     if: github.repository == 'bcgov/lear'
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - run: "true"
 
   linting:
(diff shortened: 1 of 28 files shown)


eve-git (approved):

.github/workflows/future-effective-filings-ci.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         run: |
           make test
       #- name: Upload coverage to Codecov
-      #  uses: codecov/codecov-action@v1
```
bolyachevets, line 67: without knowing much about codecov-action API, maybe the commented out sections can be removed, as they were relevant to v1 (I assume they were commented out during migration to v2, which appears to have a different pattern)

bolyachevets (approved):

.github/workflows/future-effective-filings-ci.yml
mengdong19: We'll still need it when test cased got fixed.

PR #1892, opened 2023-01-18
14978 Add fee code and allowable filings for restoration filings

argush3:
*Issue #:* /bcgov/entity#14978

*Description of changes:*

* Add fee code and allowable filings support for restoration filings(limited restoration extension & limited restoration to full)
* Add/update tests


legal-api/src/legal_api/services/authz.py
```diff
@@ -119,7 +119,9 @@ def has_roles(jwt: JwtManager, roles: List[str]) -> bool:
             'registrarsOrder': ['SP', 'GP', 'CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC'],
             'restoration': {
                 'fullRestoration': ['CP', 'BC', 'BEN', 'CC', 'ULC', 'LLC'],
-                'limitedRestoration': ['BC', 'BEN', 'CC', 'ULC', 'LLC']
+                'limitedRestoration': ['BC', 'BEN', 'CC', 'ULC', 'LLC'],
+                'limitedRestorationExtension': ['BC', 'BEN', 'CC', 'ULC', 'LLC'],
+                'limitedRestorationToFull': ['BC', 'BEN', 'CC', 'ULC', 'LLC']
```
severinbeauvais, line 124: Are you sure about enabling "LLC" right now?

severinbeauvais (approved):
LGTM

vysakh-menon-aot (approved):

argush3:
Merging.  Will create another PR if Mihai comes back that LLC should not be allowed for staff.

legal-api/src/legal_api/services/authz.py
severinbeauvais: FWIW, the UI doesn't currently handle LLC restorations.

argush3: I'm assuming it was for some ops work when they allowed it for "limitedRestoration" and "fullRestoration"

severinbeauvais: Yeah, so the API accepts those but if, somehow, Create UI is called with a draft LLC restoration then this will happen:
https://github.com/bcgov/business-create-ui/blob/0f7ce6d9fa4ce0bb1497a7d61e3341ea637cce3b/src/App.vue#L733

LLCs are similar not handled (well) in Filings UI atm. Anyway, I'm not worried about it after all.

PR #1898, opened 2023-01-20
15003 set common LOG_LEVEL from environment var

jonathan-longe:
*Issue #:* /bcgov/entity15003

Set default common log level to `ERROR` for the following queue_services:
- entity-bn
- entity-emailer
- entity-filer
- entity-pay

Anyone with access to change the environment variables, can use the `LOG_LEVEL` environment variable to change the log level in containers as needed.  Admittedly, we don't have the flexibility to set the log level for `asyncio` separately from `aiohttp.access`.  Let me know if you see that as a problem.

### DIFF
--- a/queue_services/common/src/entity_queue_common/service_utils/service_logger.py
+++ b/queue_services/common/src/entity_queue_common/service_utils/service_logger.py
@@ -29,6 +29,10 @@ def before_breadcrumb(crumb, hint):  # pylint: disable=unused-argument; callback
 
 # Configure Sentry
 SENTRY_DSN = os.getenv('SENTRY_DSN', None)
+
+# Configure default log level to ERROR
+# Override with envirionment variable if needed
+LOG_LEVEL = os.getenv('LOG_LEVEL', logging.ERROR).upper()
 if SENTRY_DSN:
 
     SENTRY_LOGGING = LoggingIntegration(
@@ -41,14 +45,11 @@ if SENTRY_DSN:
     )
 
 logging.basicConfig(
-    level=logging.INFO,
+    level=LOG_LEVEL,
     format='%(asctime)s,%(msecs)d-%(name)s-%(levelname)s > %(module)s:%(filename)s:%(lineno)d-%(funcName)s:%(message)s',
     datefmt='%H:%M:%S',
 )
-logging.getLogger('asyncio').setLevel(logging.DEBUG)
-
-if os.getenv('DISABLE_HTTP_ACCESS_LOGS', 'False').lower() == 'true':
-    # This disables the healthz / readyz / meta logging messsages.
-    logging.getLogger('aiohttp.access').setLevel(logging.ERROR)
+logging.getLogger('asyncio').setLevel(LOG_LEVEL)
+logging.getLogger('aiohttp.access').setLevel(LOG_LEVEL)
 
 logger = logging.getLogger('asyncio')


queue_services/common/src/entity_queue_common/service_utils/service_logger.py
```diff
@@ -41,14 +45,11 @@ def before_breadcrumb(crumb, hint):  # pylint: disable=unused-argument; callback
...
     format='%(asctime)s,%(msecs)d-%(name)s-%(levelname)s > %(module)s:%(filename)s:%(lineno)d-%(funcName)s:%(message)s',
     datefmt='%H:%M:%S',
 )
-logging.getLogger('asyncio').setLevel(logging.DEBUG)
-
-if os.getenv('DISABLE_HTTP_ACCESS_LOGS', 'False').lower() == 'true':
-    # This disables the healthz / readyz / meta logging messsages.
-    logging.getLogger('aiohttp.access').setLevel(logging.ERROR)
+logging.getLogger('asyncio').setLevel(LOG_LEVEL)
+logging.getLogger('aiohttp.access').setLevel(LOG_LEVEL)
```
jonathan-longe, line 53: The variable `DISABLE_HTTP_ACCESS_LOGS` doesn't appear to be used.  By setting the log level as an environment variable we can't set different log levels, but we have more control to reduce the logging volume.

jonathan-longe, line 53: The previous comment didn't quote all the lines I intended.  Trying again ...

vysakh-menon-aot (approved):

queue_services/common/src/entity_queue_common/service_utils/service_logger.py
argush3: you might want if to check if `DISABLE_HTTP_ACCESS_LOGS` really isn't being used anywhere.  I think you added this env var @seeker25?

queue_services/common/src/entity_queue_common/service_utils/service_logger.py
```diff
@@ -29,6 +29,10 @@ def before_breadcrumb(crumb, hint):  # pylint: disable=unused-argument; callback
 
 # Configure Sentry
 SENTRY_DSN = os.getenv('SENTRY_DSN', None)
+
+# Configure default log level to ERROR
+# Override with envirionment variable if needed
+LOG_LEVEL = os.getenv('LOG_LEVEL', logging.ERROR).upper()
```
argush3, line 35: just a note that I think there are queue services that are not in the LEAR repo that are referencing this service logger module.  This will mean log levels will be set to `ERROR` if they don't have the `LOG_LEVEL` config right?

jonathan-longe: That's correct.  To change the log level in any container / pod, add "LOG_LEVEL" to the appropriate container and environment in 1password.  Once saved, redeploy the container and the environment variable will automatically be added.  Please let me know if you have ANY concerns.

argush3: That's fine as long as we let the other teams know about it.  I think at least a bunch of the auth listeners in other repos are using this common service logger.  The next time they push deploy through the different environments they will have unexpected log levels if they are unaware.  Maybe give a heads up in developer channel after this change makes it into `main`.

For example, this listener in the auth repo is using this logger.  https://github.com/bcgov/sbc-auth/blob/7bf026b76e5094b1b9f5f6aef68b073d918fcd2c/queue_services/business-events-listener/requirements.txt#L34

seeker25:
I remember attempting to get this working, if I recall correctly I think Thor wanted to put this under a launch darkly flag not an environment variable?

jonathan-longe:
> I remember attempting to get this working, if I recall correctly I think Thor wanted to put this under a launch darkly flag not an environment variable?

Thanks @seeker25.  I'll ask @thorwolpert for his review of this change too.

thorwolpert (approved):
I'm good with this change.
The launchdarkly switch will be there as part of the new queue common libraries, so no need to worry about that here (yet)

argush3 (approved):

PR #1906, opened 2023-01-27
15003 - default to 'ERROR' string not integer

jonathan-longe:
*Issue #:* /bcgov/entity15003

When LOG_LEVEL not provided, default to `ERROR` string not the integer `logging.ERROR`.


argush3 (approved):

vysakh-menon-aot (approved):

severinbeauvais:
What about the failed checks?

vysakh-menon-aot:
Those failures are due to the previous implementation. With this fix upcoming PR's will pass checks.

PR #1907, opened 2023-01-30
15114 Update create invoice logic to pass filing id

argush3:
*Issue #:* /bcgov/entity#15114

*Description of changes:*

* Update create invoice logic to pass filing id.  The pay api will use this to populate the `filingIdentifier` property in the payment queue message that it sends to the entity pay listener.  This will allow entity pay to not require the sleep/retry code it currently has in place as it tries to determine which filing to process by looking for the `filing.payment_id` instead of the `filing.id`.  


argush3:
2 breaking tests are unrelated to this PR

vysakh-menon-aot (approved):

seeker25 (approved):

PR #1910, opened 2023-01-30
14641 Update model for Restoration

vysakh-menon-aot:
*Issue #:* /bcgov/entity#14641, /bcgov/entity#13328

*Description of changes:*


ChrisGabel23 (approved):

argush3 (approved):

legal-api/src/legal_api/core/filing.py
```diff
@@ -79,6 +79,7 @@ class FilingTypes(str, Enum):
         SPECIALRESOLUTION = 'specialResolution'
         TRANSITION = 'transition'
         PUTBACKON = 'putBackOn'
+        RESTORATION = 'restoration'
```
thorwolpert, line 82: Could you put this in alpha order, and fix the putBackOn too?
It'll make the longterm maintenance and debugging easier

thorwolpert (commented):
We've been moving to SQL Enums, and this is a departure from that. Would we do both, transition to all lookup tables and add tests to keep the enums & lookups in sync?

vysakh-menon-aot:
> We've been moving to SQL Enums, and this is a departure from that. Would we do both, transition to all lookup tables and add tests to keep the enums & lookups in sync?

We haven't migrated FilingTypes and LegalTypes to SQL Enums

PR #1938, opened 2023-02-21
Fix auth bug for restoration filings

argush3:
*Issue #:* /bcgov/entity#

*Description of changes:*

* Fix bug where authorization logic was only allowing restoration filings of sub-type `limitedRestorationExtension` and `limitedRestorationToFull` for a corp in `HISTORICAL` state.  These two types of restorations should be only available when the corp is `ACTIVE`.
* Update tests


jonathan-longe (approved):
Looks good.  I like how the business rules are in a configuration constant.

PR #1945, opened 2023-02-28
15479 Fix NR validation logic for restorations

argush3:
*Issue #:* /bcgov/entity#15479

*Description of changes:*

* Update restoration NR validation logic to not validate NR block when there is no `nrNumber` provided in filing json for a NR.  This follows the logic used for corrections.
* Added tests

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/restoration.py
+++ b/legal-api/src/legal_api/services/filings/validations/restoration.py
@@ -40,8 +40,13 @@ def validate(business: Business, restoration: Dict) -> Optional[Error]:
     elif restoration_type in ('fullRestoration', 'limitedRestorationToFull'):
         msg.extend(validate_relationship(restoration))
 
-    if restoration.get('filing', {}).get('restoration', {}).get('nameRequest', {}).get('nrNumber', None):
+    name_request = restoration.get('filing', {}).get('restoration', {}).get('nameRequest', {})
+    if name_request.get('nrNumber', None):
         msg.extend(validate_name_request(restoration, business.legal_type, filing_type))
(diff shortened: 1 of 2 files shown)


vysakh-menon-aot (approved):

legal-api/src/legal_api/services/filings/validations/restoration.py
```diff
@@ -40,7 +40,8 @@ def validate(business: Business, restoration: Dict) -> Optional[Error]:
     elif restoration_type in ('fullRestoration', 'limitedRestorationToFull'):
         msg.extend(validate_relationship(restoration))
 
-    msg.extend(validate_name_request(restoration, business.legal_type, filing_type))
+    if restoration.get('filing', {}).get('restoration', {}).get('nameRequest', {}).get('nrNumber', None):
```
severinbeauvais, line 43: Are there tests for the 3 cases?

argush3: no, name requests aren't checked currently for restoration validation tests

severinbeauvais: That would be good to do, since (1) the business name is required, and (2) we would catch it if a future change breaks something.

argush3: ok, will add some tests to restoration validator tests for NR block.

severinbeauvais: Thanks. Looks good!

severinbeauvais (approved):

PR #1959, opened 2023-03-16
Legal API: validate restoration NR request type code

kzdev420:
*Issue #:* /bcgov/entity#15445

*Description of changes:*
Add the validation for restoration NR request code type

### DIFF
--- a/legal-api/tests/unit/services/filings/validations/test_restoration.py
+++ b/legal-api/tests/unit/services/filings/validations/test_restoration.py
@@ -44,8 +44,7 @@ nr_response = {
         'name': legal_name,
         'state': 'APPROVED',
         'consumptionDate': ''
-    }],
-    'requestTypeCd': 'RCC'
+    }]
 }
 relationships = ['Heir or Legal Representative', 'Director']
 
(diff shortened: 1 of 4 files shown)


argush3:
@kzdev420 `alteration.py` is not the file that needs to be updated and the validation update needs to be specific to name requests and not things like court order, share structure and etc.  The GH link Severin provided was just an example of how a name request's request type code was validated for a different filing(_alteration_)  

You can start by looking at the name request restoration validation logic here as a starting point
https://github.com/bcgov/lear/blob/28c0820aef9eaee3dafe09a3dc1fa36219aa9aad/legal-api/src/legal_api/services/filings/validations/restoration.py#L45-L44

I think maybe you can add a `accepted_request_types` parameter to the [validate_name_request function](https://github.com/bcgov/lear/blob/a17e978ab72aafadca4a07a48639598864d36ea0/legal-api/src/legal_api/services/filings/validations/common_validations.py#L231-L234) and add a check on the `nr_response` object to validate that the returned `requestTypeCd` matches `RCC`, `RCR`, `BERE`, or `RUL`.

I think you'll also want to add a test after this test to test the new NR validation logic for restorations https://github.com/bcgov/lear/blob/28c0820aef9eaee3dafe09a3dc1fa36219aa9aad/legal-api/tests/unit/services/filings/validations/test_restoration.py#L223-L225

legal-api/tests/unit/services/filings/validations/test_restoration.py
```diff
@@ -201,26 +201,30 @@ def test_restoration_court_orders(session, test_status, file_number, expected_co
...
          'Legal name is missing in nameRequest.'),
-        ('FAIL_NR_AND_NO_NAME', 'limitedRestoration', ['BC', 'BEN', 'ULC', 'CC'], 'NR 1234567', None,
+        ('FAIL_NR_AND_NO_NAME', 'limitedRestoration', ['BC', 'BEN', 'ULC', 'CC'], 'NR 1234567', 'RUL', None,
          HTTPStatus.BAD_REQUEST, 'Legal name is missing in nameRequest.'),
+        ('FAIL_NO_NR_TYPE', 'limitedRestoration', ['BC', 'BEN', 'ULC', 'CC'], 'NR 1234567', '', 'new name',
+         HTTPStatus.BAD_REQUEST, 'The name type associated with the name request number entered cannot be used.'),
     ]
 )
-def test_restoration_nr(session, mocker, test_status, filing_sub_type, legal_types, nr_number, new_legal_name,
+def test_restoration_nr(session, mocker, test_status, filing_sub_type, legal_types, nr_number, nr_type, new_legal_name,
```
argush3, line 227: I think we should create a new test to test nr request type specifically.  It's probably possible to include it in this test but it's getting a little messy and we'll need to add some special cases.

I was envisioning the new test to look similar to this but with an additional for loop to loop through an array of nr request types.

With the current test, I would revert the code for this test and maybe update `nr_response` at the top of this `test_restoration.py` to have a valid restoration `requestTypeCd` value.

argush3: I was thinking something like this for the new test.  You'll of course need to look at `limitedRestoration` as well.

```
('SUCCESS_NR', 'fullRestoration', ['BC', 'BEN', 'ULC', 'CC'], 'NR 1234567', ['RCC', 'RCR', 'BERE', 'RUL'], None, None),
('FAIL_NR', 'fullRestoration', ['BC', 'BEN', 'ULC', 'CC'], 'NR 1234567', ['RCC', 'RCR', 'BERE', 'RUL'], None, None),
def test_restoration_nr_type(session, mocker, test_status, filing_sub_type, legal_types, nr_number, nr_type, expected_code, expected_msg):
```

kzdev420: I updated

ChrisGabel23 (approved):

PR #2050, opened 2023-05-19
16120 Update allowable actions - Continuation Out for Corporations

kzdev420:
*Issue #:* /bcgov/entity#16120

*Description of changes:*

- Corporation is in Good Standing
- Check the completed filings with consentContinuationOut


legal-api/src/legal_api/services/authz.py
```diff
@@ -155,6 +155,13 @@ def has_roles(jwt: JwtManager, roles: List[str]) -> bool:
                     'business': [BusinessBlocker.DEFAULT, BusinessBlocker.NOT_IN_GOOD_STANDING]
                 }
             },
+            'continuationOut': {
+                'legalTypes': ['BC', 'BEN', 'CC', 'ULC'],
+                'blockerChecks': {
+                    'business': [BusinessBlocker.NOT_IN_GOOD_STANDING],
+                    'completedFilings': ['consentContinuationOut']
```
argush3, line 162: We need code to actually do the check for `completedFilings`.  Take a look at the code used for `validStateFilings`.  It won't be the same but the pattern will be similar I think.

legal-api/src/legal_api/services/authz.py
```diff
@@ -533,6 +543,23 @@ def has_blocker_invalid_state_filing(state_filing: Filing, blocker_checks: dict)
     return has_state_filing(state_filing, state_filing_types)
 
 
+def has_blocker_completed_filing(completed_filing: Filing, blocker_checks: dict):
```
argush3, line 546: `completed_filing` parameter is not required.

legal-api/src/legal_api/services/authz.py
```diff
@@ -533,6 +543,23 @@ def has_blocker_invalid_state_filing(state_filing: Filing, blocker_checks: dict)
...
+    if not completed_filing:
+        return False
+
+    if not has_state_filing(completed_filing, complete_filing_types):
+        return False
+
+    if completed_filing.status == Filing.Status.COMPLETED.value:
+        return True
+
+    return False
```
argush3, line 560: The code here should be a check to see if the filing types provided(if any) in the `completedFilings` list exist for the business and have a filing status of `COMPLETED`.   If not return `False`.

legal-api/tests/unit/services/test_authorization.py
```diff
@@ -1541,6 +1561,159 @@ def mock_auth(one, two):  # pylint: disable=unused-argument; mocks of library me
...
+                          FilingKey.AR_CP,
+                          FilingKey.COA_CP,
+                          FilingKey.COD_CP,
+                          FilingKey.CORRCTN,
+                          FilingKey.COURT_ORDER,
+                          FilingKey.VOL_DISS,
+                          FilingKey.ADM_DISS,
+                          FilingKey.REGISTRARS_NOTATION,
+                          FilingKey.REGISTRARS_ORDER,
+                          FilingKey.SPECIAL_RESOLUTION])),
```
argush3, line 1580: At this point, I think we only need to do corps(BC, BEN, CC, ULC) tests for this test.  Consent continuation out filings won't be allowed for non-corps and I think we have this covered in other tests.  So this scenario is not possible.  

Can we remove all tests that don't involve corps?  In the future, we may have other filings and entity types using `completedFilings` but for now its just the `continuationOut` filing using it.

kzdev420: Yes, I will

legal-api/tests/unit/services/test_authorization.py
```diff
@@ -461,6 +471,16 @@ def mock_auth(one, two):  # pylint: disable=unused-argument; mocks of library me
...
+        ('staff_active_allowed', Business.State.ACTIVE, 'consentContinuationOut', None,
+         ['BC', 'BEN', 'ULC', 'CC'], 'staff', [STAFF_ROLE], True),
+        ('staff_active', Business.State.ACTIVE, 'consentContinuationOut', None,
+         ['CP', 'LLC'], 'staff', [STAFF_ROLE], False),
+
+        ('staff_active_allowed', Business.State.ACTIVE, 'continuationOut', None,
+         ['BC', 'BEN', 'ULC', 'CC'], 'staff', [STAFF_ROLE], False),
+        ('staff_active', Business.State.ACTIVE, 'continuationOut', None,
+         ['CP', 'LLC'], 'staff', [STAFF_ROLE], False),
+
```
argush3, line 483: please add this test for general user to confirm they can't file a `continuationOut`

PR #2066, opened 2023-05-26
16213 - unblock correction menu

Jxio:
*Issue #:*[ /bcgov/entity16213](https://github.com/bcgov/entity/issues/16213)

*Description of changes: unblock correction menu*

### DIFF
--- a/legal-api/src/legal_api/services/authz.py
+++ b/legal-api/src/legal_api/services/authz.py
@@ -168,8 +168,7 @@ ALLOWABLE_FILINGS: Final = {
             'correction': {
                 'legalTypes': ['CP', 'BEN', 'SP', 'GP', 'BC', 'ULC', 'CC'],
                 'blockerChecks': {
-                    'warningTypes': [WarningType.MISSING_REQUIRED_BUSINESS_INFO],
-                    'business': [BusinessBlocker.DEFAULT]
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/services/authz.py
```diff
@@ -168,8 +168,7 @@ def has_roles(jwt: JwtManager, roles: List[str]) -> bool:
             'correction': {
                 'legalTypes': ['CP', 'BEN', 'SP', 'GP', 'BC', 'ULC', 'CC'],
                 'blockerChecks': {
-                    'warningTypes': [WarningType.MISSING_REQUIRED_BUSINESS_INFO],
-                    'business': [BusinessBlocker.DEFAULT]
```
argush3, line 172: I don't think we should be removing `'business': [BusinessBlocker.DEFAULT]`.  

What is the motivation for removing the the default blockers(draft/pending filing + admin freeze) here?

The current allowable actions logic for corrections was previously reviewed via the [allowable actions spreadsheet](https://docs.google.com/spreadsheets/d/1sxsZ7vuODtD_ftjR27qLvgYErWDB-msvzGucjvBRQXQ/edit#gid=0).

image

Jxio: Thanks Argus! based on this doc, yes I think I can't remove it. I'll close this PR.

PR #2083, opened 2023-06-05
16539 update consent letter

vysakh-menon-aot:
*Issue #:* /bcgov/entity#16539

*Description of changes:*
 - include jurisdiction in letter of consent

### DIFF
--- a/legal-api/src/legal_api/models/consent_continuation_out.py
+++ b/legal-api/src/legal_api/models/consent_continuation_out.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """This model holds data for consent continuation out."""
 from __future__ import annotations
+
 from typing import Optional
 
 from .db import db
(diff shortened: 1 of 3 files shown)


argush3 (approved):

legal-api/src/legal_api/models/consent_continuation_out.py
```diff
@@ -49,3 +52,9 @@ def get_active_cco(business_id, expiry_date, foreign_jurisdiction=None, foreign_
                 ConsentContinuationOut.foreign_jurisdiction_region == foreign_jurisdiction_region.upper())
 
         return query.all()
+
+    @staticmethod
+    def get_by_filing_id(filing_id) -> Optional[ConsentContinuationOut]:
+        """Get consent continuation out by filing_id."""
+        return db.session.query(ConsentContinuationOut). \
+            filter(ConsentContinuationOut.filing_id == filing_id).one_or_none()
```
ChrisGabel23, line 60: What happens if this DB call returns null? Won't that cause issues in report.py?

vysakh-menon-aot: letter of consent is available only when filing status is COMPLETED, by then this table would be populated.

ChrisGabel23 (approved):

PR #2087, opened 2023-06-06
16658 using v2 save filings

vysakh-menon-aot:
*Issue #:* /bcgov/entity#16658

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -138,7 +138,7 @@ def post_businesses():
         return {'error': babel('Unable to create {0} Filing.'.format(Filing.FILINGS[filing_type]['title']))}, \
             HTTPStatus.SERVICE_UNAVAILABLE
 
-    return saving_filings(identifier=bootstrap.identifier)
+    return saving_filings(identifier=bootstrap.identifier)  # pylint: disable=no-value-for-parameter
 
 
 @bp.route('/search', methods=['POST'])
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -138,7 +138,7 @@ def post_businesses():
         return {'error': babel('Unable to create {0} Filing.'.format(Filing.FILINGS[filing_type]['title']))}, \
             HTTPStatus.SERVICE_UNAVAILABLE
 
-    return ListFilingResource.put(bootstrap.identifier, None)
+    return saving_filings(identifier=bootstrap.identifier)  # pylint: disable=no-value-for-parameter
```
argush3, line 141: I think this might be problematic if we do it this way?  

We bypass some checks that was originally done in the `PUT` call to the `ListingFilingResource`.  

Also, it seems like it was possible via this endpoint to do a `POST` and create the business in one go if the "draft" query param wasn't set?  I think that's what I'm seeing.  With this update, we lose the ability to do that.

vysakh-menon-aot: ListFilingResource.put in v1 and saving_filings in v2 are same. It won't bypass any checks

argush3: oh sorry, I got confused.  I was looking at the `ListFilingResource.save_filing`

argush3 (approved):

PR #2097, opened 2023-06-21
Fix emailer q_cli

argush3:
*Issue #:* /bcgov/entity###

*Description of changes:*

* Fix import order issue that was causing emailer `q_cli.py` to error out when called. 


vysakh-menon-aot (approved):

argush3:
It is related to 16730...the emailer retry ticket.  It was an update to just make it easier to test the emailer via the q_cli.  

@chenhongjing is still working on the retry code.  But we can deploy the q_cli.py updates.

PR #2102, opened 2023-06-23
16730 Dynamically configure number of email resending

chenhongjing:
*Issue #:* /bcgov/entity#16730

*Description of changes:*
-  Add a way to dynamically configure number of retries for emailer. It need to add a new variable `MSG_RETRY_NUM` at `.env`. Note that the default number of retries is 5.
- Update unit test

### DIFF
--- a/queue_services/entity-emailer/src/entity_emailer/worker.py
+++ b/queue_services/entity-emailer/src/entity_emailer/worker.py
@@ -57,8 +57,6 @@ from entity_emailer.email_processors import (
     restoration_notification,
 )
 
-from .message_tracker import tracker as tracker_util
-
 
 qsm = QueueServiceManager()  # pylint: disable=invalid-name
 APP_CONFIG = config.get_named_config(os.getenv('DEPLOYMENT_ENV', 'production'))
@@ -66,6 +64,7 @@ FLASK_APP = Flask(__name__)
 FLASK_APP.config.from_object(APP_CONFIG)
 db.init_app(FLASK_APP)
 
+from .message_tracker import tracker as tracker_util
 
 async def publish_event(payload: dict):
     """Publish the email message onto the NATS event subject."""
--- a/queue_services/entity-emailer/src/entity_emailer/worker.py
+++ b/queue_services/entity-emailer/src/entity_emailer/worker.py
@@ -59,6 +59,7 @@ from entity_emailer.email_processors import (
 
 from .message_tracker import tracker as tracker_util
 
+
 qsm = QueueServiceManager()  # pylint: disable=invalid-name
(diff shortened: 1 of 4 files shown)


queue_services/entity-emailer/src/entity_emailer/worker.py
```diff
@@ -57,15 +57,14 @@
...
-from .message_tracker import tracker as tracker_util
-
 
 qsm = QueueServiceManager()  # pylint: disable=invalid-name
 APP_CONFIG = config.get_named_config(os.getenv('DEPLOYMENT_ENV', 'production'))
 FLASK_APP = Flask(__name__)
 FLASK_APP.config.from_object(APP_CONFIG)
 db.init_app(FLASK_APP)
 
+from .message_tracker import tracker as tracker_util
```
chenhongjing, line 67: I'm not sure whether it's proper to change the order of import here. ❔  But it can avoid repeated importing and partial import error.

chenhongjing: Oops. It will make linter fail. 😨

queue_services/entity-emailer/tests/unit/test_tracker.py
```diff
@@ -430,11 +430,11 @@ async def test_should_correctly_track_retries_for_failed_processing(tracker_app,
 
     # mock out process_email function to throw exception to simulate failed scenario
     with patch.object(worker, 'process_email', side_effect=QueueException('Queue Error.')):
```
argush3, line 432: can we update this test to use `side_effect=EmailException('Queue Error.')` instead?  That will reflect what we are trying to test more accurately as it is one of the exceptions that will actually trigger retries.  

Also can we add/update a comment indicating that we are testing the max retry limit?

queue_services/entity-emailer/tests/unit/test_tracker.py
```diff
@@ -429,12 +429,19 @@ async def test_should_correctly_track_retries_for_failed_processing(tracker_app,
...
-        for x in range(5):
+    with patch.object(worker, 'process_email', side_effect=EmailException('Queue Error.')):
+        for _ in range(6):
+            with pytest.raises(EmailException):
+                await worker.cb_subscription_handler(mock_msg)
+
+    # mock out process_email function not to process since message reaches max retries
+    with patch.object(worker, 'process_email', side_effect=EmailException('Queue Error.')):
+        for _ in range(5):
             await worker.cb_subscription_handler(mock_msg)
```
argush3, line 440: Maybe we should just parametrize this test and test all the exception types(`OperationalError`, `EmailException`, `QueueException`, `Exception`) and verify it stops retrying after the retry limit is exceeded.

argush3 (approved):

PR #2108, opened 2023-06-27
16743 legal name db migration updates

argush3:
*Issue #:* /bcgov/entity#16743

*Description of changes:*

* Updated existing `legal_name_updates.sql` script to follow new versioning model.  Note this script was previously named `2023_may_25_legal_name_updates.sql`
* Added(step 6) migration logic to `legal_name_updates.sql` to update party ids in filing json to use newly created legal entity ids and colin entity ids.
* Added alembic scripts for misc updates and also to add `consent_continuation_outs` and `sent_to_gazette` tables
* Updated `transfer_to_new_lear.sql` to include `consent_continuation_outs` and `sent_to_gazette` tables
* Updated README
* Misc fixes to scripts

**Examples of firms after migrations scripts run against new LEAR model**

_SP example_
image

_GP example_
image

_GP example_
image


legal-api/migrations/versions/2fd104a5f3b5_.py
```diff
@@ -0,0 +1,42 @@
...
+# revision identifiers, used by Alembic.
+revision = '2fd104a5f3b5'
+down_revision = 'e7abbdeb252a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('party_roles', schema=None) as batch_op:
```
thorwolpert, line 21: Are we deleting party_roles or renaming it?

thorwolpert (approved):
Just the one question on party_roles
The rest looks really good.
Maybe at the end we should force the vacuum service to reorg the space, but the migration looks pretty nice.

legal-api/migrations/versions/2fd104a5f3b5_.py
argush3: We'll be deleting `party_roles` and `parties` table.  `entity_roles`, `legal_entities` and `colin_entities` will be used.  

The current migration flow/scripts transfer all the existing data into the new LEAR model first which is why I'm including the `change_filing_id` here.  From there, `legal_name_updates.sql` is used to move the  party related data into the new tables(`entity_roles`, `legal_entities` and `colin_entities`).

I think once we are confident where things are we can remove the `party_roles` and `parties` table.  IThe `signing_party_id` in the `resolutions` table will also need to be removed.

PR #2111, opened 2023-06-30
16901 Update db migration scripts to work with GCP loads

argush3:
*Issue #:* /bcgov/entity#16901

*Description of changes:*

* Updated scripts to add/drop constraints instead of disabling/enabling triggers.  Google Cloud SQL databases do not allow disabling of triggers easily.  
* Updated README such that `transfer_to_near_lear_after.sql` is run as last step.
* Changed ordering of table transfers in `transfer_to_new_lear.sql` to reduce adding/removal of constraints.
* Added additional tables to sequence updates as well as updated sequence updating logic to handle edge case.

Note that due to there being orphaned data in dev/test/prod databases, not all dropped constraints could be added back in the `transfer_to_new_lear_after.sql` script.  This issue will go away once the cleanup orphaned data script is created and applied via the work in #16902.  

As an example of the constraint issue, a successful test data load to GCP using a recent DEV LEAR db dump resulted in the following constraints remaining dropped.
image

### DIFF
--- a/legal-api/scripts/manual_db_scripts/legal_name_change/README.md
+++ b/legal-api/scripts/manual_db_scripts/legal_name_change/README.md
@@ -65,10 +65,10 @@ Transfer using 8 thread(s) sent_to_gazette ...                      1038 rows in
 Done
 
 ```
-8. Cleanup temporary artifacts and states created by data transfer scripts.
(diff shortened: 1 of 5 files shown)


vysakh-menon-aot (approved):

thorwolpert (approved):
good catch on the constraints.

PR #2139, opened 2023-07-14
first pass for draft

thorwolpert:
*Issue #:* /bcgov/entity#16407 (a part)

*Description of changes:*
- moved business model to it's own module
- trying to keep as slim as possible
- Flask is only in the tests


argush3 (approved):
Looks good.  thanks for the README!

python/common/business-registry-model/src/migrations/versions/b13796c7ba9a_.py
```diff
@@ -0,0 +1,73 @@
+"""empty message
+
+Revision ID: b13796c7ba9a
+Revises: 60d9c14c2b7f
+Create Date: 2023-07-13 23:03:01.394887
+
```
argush3, line 6: I think we can remove this file now.

PR #2164, opened 2023-07-26
model changes, move versions, update versioning delete

thorwolpert:
*Issue #:* n/a

*Description of changes:*
- updated model
- moved versioning into a module
- updated sql-versioning to store deletes


python/common/business-registry-model/src/business_model/models/legal_entity.py
```diff
@@ -753,3 +754,47 @@ def receive_before_change(mapper, connection, target):  # pylint: disable=unused
     LegalEntity.AssociationTypes.SP_SOLE_PROPRIETORSHIP.value: 'Sole Proprietorship',
     LegalEntity.AssociationTypes.SP_DOING_BUSINESS_AS.value: 'Sole Proprietorship (DBA)'
 }
+
+class LegalEntityType(str, Enum, metaclass=BaseMeta):
+    """The business type."""
+
+    COOPERATIVE = 'CP'
```
vysakh-menon-aot, line 761: missing `BC`?

python/common/business-registry-model/src/business_model_migrations/versions/b13796c7ba9a_.py
argush3, line 1: We should be able to delete this file now.

PR #2166, opened 2023-07-27
17147 update get version to use sql_versioning

vysakh-menon-aot:
*Issue #:* /bcgov/entity#17147

*Description of changes:*


legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -196,215 +192,145 @@ def get_company_details_revision(filing_id, legal_entity_id) -> dict:
...
-            .filter(Office.deactivated_date is null())
+            .filter(Office.deactivated_date == None)  # pylint: disable=singleton-comparison # noqa: E711,E501;
 
+        office_history = history_cls(Office)
         offices_historical = db.session.query(office_history) \
-            .filter(office_history.change_filing_id <= filing_id) \
+            .filter(office_history.change_filing_id == filing_id) \
             .filter(office_history.legal_entity_id == legal_entity_id)  \
-            .filter(Office.deactivated_date > filing.effective_date)
+            .filter(office_history.deactivated_date > filing.effective_date)
```
vysakh-menon-aot, line 267: @thorwolpert since we are filtering `change_filing_id == filing_id` does this condition make sense?

thorwolpert: i think your suggestion of rolling with `change_filing_id == filing_id` makes a lot of that logic go away.
good catch, and keep simplifying this as much as we can.

vysakh-menon-aot: we still need a field in each history table to track deleted or deactivated as part of a filing

legal-api/tests/unit/services/test_business_version.py
```diff
@@ -0,0 +1,210 @@
...
+            db.session.add(office)
+            if existing_address := office.addresses.all():
+                for address in existing_address:
+                    address.change_filing_id = alt_filing.id
+                    db.session.add(address)
+            db.session.commit()
+
+            if office.office_type == OfficeType.RECORDS:
+                db.session.delete(office)
+                db.session.commit()
```
vysakh-menon-aot, line 152: @thorwolpert are you following the same method in filer?

argush3 (approved):

PR #2195, opened 2023-09-12
17504 create proxy call colin api

kzdev420:
*Issue #:* /bcgov/entity#17504

*Description of changes:*

GET `legal-api-url/api/v2/colin/BC12322`
Response
```
{
        'business': {
                'identifier': 'BC123221',
                'legalName': 'TEST BUSINESS LTD',
                'legalType': 'COOP',
                'jurisdiction': 'BC',
                'homeIdentifier': 'BC234232'
          }
}
```

### DIFF
--- a/legal-api/src/legal_api/resources/v2/colin.py
+++ b/legal-api/src/legal_api/resources/v2/colin.py
@@ -1,0 +1,42 @@
+# Copyright © 2021 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Searching on a namerequest.
+
+Provides a proxy endpoint to retrieve corp's information.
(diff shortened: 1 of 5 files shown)


legal-api/src/legal_api/resources/v2/colin.py
```diff
@@ -0,0 +1,42 @@
...
+
+bp = Blueprint('COLIN2', __name__, url_prefix='/api/v2/colin')
+
+
+@bp.route('/<string:corp_num>', methods=['GET'])
+@cross_origin(origin='*')
+def get(corp_num):
+    """Return a JSON object."""
+    try:
+        nr_response = colin.find_by_corp_num(corp_num)
```
argush3, line 32: please rename `nr_response` to `response`

legal-api/src/legal_api/resources/v2/colin.py
```diff
@@ -0,0 +1,42 @@
...
+
+from legal_api.services import colin
+
+
+bp = Blueprint('COLIN2', __name__, url_prefix='/api/v2/colin')
+
+
+@bp.route('/<string:corp_num>', methods=['GET'])
+@cross_origin(origin='*')
+def get(corp_num):
```
argush3, line 29: This endpoint needs to be secure.  

Check that the api consumer for this endpoint has the `COLIN_SVC_ROLE`.

thorwolpert (commented):
Why are we proxying other services through the business-api service?
This will remove the default JWT check by the gateway.

legal-api/src/legal_api/resources/v2/colin.py
kzdev420: This endpoint will be call by NameX

PR #2196, opened 2023-09-13
17638 Made regions all optional for business office, business party a…

PaulGarewal:
…nd completing party

*Issue #:* /bcgov/entity#17638

*Description of changes:*

- added business party and completing party to all have optional region option.
- updated according unit tests 

### DIFF
--- a/legal-api/src/legal_api/services/warnings/business/business_checks/firms.py
+++ b/legal-api/src/legal_api/services/warnings/business/business_checks/firms.py
@@ -257,8 +257,9 @@ def check_address(address: Address,
                                                    address_type,
                                                    BusinessWarnings.NO_ADDRESS_POSTAL_CODE))
 
-    if referer not in (BusinessWarningReferers.BUSINESS_OFFICE, BusinessWarningReferers.BUSINESS_PARTY,
-                       BusinessWarningReferers.COMPLETING_PARTY) and not address.region:
+    if (referer == BusinessWarningReferers.BUSINESS_OFFICE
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/services/warnings/business/business_checks/firms.py
```diff
@@ -257,7 +257,8 @@ def check_address(address: Address,
                                                    address_type,
                                                    BusinessWarnings.NO_ADDRESS_POSTAL_CODE))
 
-    if referer != BusinessWarningReferers.BUSINESS_OFFICE and not address.region:
+    if referer not in (BusinessWarningReferers.BUSINESS_OFFICE, BusinessWarningReferers.BUSINESS_PARTY,
+                       BusinessWarningReferers.COMPLETING_PARTY) and not address.region:
```
argush3, line 261: I think it should be only adding a no address region warning if all 3 conditions below are met.  This would be the case where we expect the business office address to always have a non-null region value.  i.e. it should allways have a value of "BC"

1. referer == BusinessWarningReferers.BUSINESS_OFFICE
2. address_type == 'Address.DELIVERY
3. address.region is null

PaulGarewal: Makes way more sense to me now, thank you

argush3:
I think we need a test a `FAIL_NO_REGION` test for business delivery addresses.

image

argush3 (approved):

PR #2202, opened 2023-09-19
17447 Remove unnecessary logging

argush3:
*Issue #:* /bcgov/entity#17447

*Description of changes:*

* Remove most of the logging that was added/used to determine why calls to colin api with auth turned on was not working.  Issue turned out the  image pull policies in the cronjob.yaml files in update-colin-filings and update-colin-legal filings job were not set to "always".


chenhongjing (approved):
😺

PR #2232, opened 2023-10-12
display name properly on pdf when last name only

tshyun24:
*Issue #:* /bcgov/entity#18034

*Description of changes:*
Add if else statement to display first name, middle name, and last name properly on pdf when both middle name and first name are optional

### DIFF
--- a/legal-api/report-templates/template-parts/registration/completingParty.html
+++ b/legal-api/report-templates/template-parts/registration/completingParty.html
@@ -12,10 +12,19 @@
                         {% if party.officer.partyType == 'organization' %}
                            <span class="capitalize-text">{{ party.officer.organizationName }}</span>
                         {% else %}
-                           <span class="capitalize-text">{{ party.officer.lastName }}</span>,
-                           <span class="capitalize-text">{{ party.officer.firstName }}</span>
-                           {% if party.officer.middleName %}
-                              <span class="capitalize-text">{{ party.officer.middleName }}</span>
+                           {% if party.officer.firstName %}
+                              <span class="capitalize-text">{{ party.officer.lastName }}</span>,
+                              <span class="capitalize-text">{{ party.officer.firstName }}</span>
+                              {% if party.officer.middleName %}
+                                 <span class="capitalize-text">{{ party.officer.middleName }}</span>
+                              {% endif %}
+                           {% else %}
+                              {% if party.officer.middleName %}
+                                 <span class="capitalize-text">{{ party.officer.lastName }}</span>,
+                                 <span class="capitalize-text">{{ party.officer.middleName }}</span>
+                              {% else %}
+                                 <span class="capitalize-text">{{ party.officer.lastName }}</span>
+                              {% endif %}
                            {% endif %}
                         {% endif %}
                      </div>


legal-api/report-templates/template-parts/registration/completingParty.html
```diff
@@ -12,10 +12,19 @@
...
+                              {% if party.officer.middleName %}
+                                 <span class="capitalize-text">{{ party.officer.middleName }}</span>
+                              {% endif %}
+                           {% else %}
+                              {% if party.officer.middleName %}
+                                 <span class="capitalize-text">{{ party.officer.lastName }}</span>,
+                                 <span class="capitalize-text">{{ party.officer.middleName }}</span>
+                              {% else %}
+                                 <span class="capitalize-text">{{ party.officer.lastName }}</span>
+                              {% endif %}
```
argush3, line 27: I think you might be able to simplify this template by using [_get_party_name](https://github.com/bcgov/lear/blob/6e4130fda5c79c06ec576f5911e92dbb697eacb3/legal-api/src/legal_api/reports/report.py#L707-L717) in [_format_registration_data](https://github.com/bcgov/lear/blob/6e4130fda5c79c06ec576f5911e92dbb697eacb3/legal-api/src/legal_api/reports/report.py#L260-L259).  

Basically, loop through each party in the filing and just add a new field to the party to represent the display name for the party.  It won't be the exact code but you can use this piece of [code](https://github.com/bcgov/lear/blob/6e4130fda5c79c06ec576f5911e92dbb697eacb3/legal-api/src/legal_api/reports/report.py#L847-L846) as a reference too.

With this, I'm thinking you will only need to use something like `<span class="capitalize-text">{{ party.officer.fullName }}</span>`.  And it will also work for the case where partyType == 'organization' I'm thinking.

tshyun24: I pushed a new comment with adding fullName field, not sure whether the new function name are standards

legal-api/report-templates/template-parts/registration/completingParty.html
tshyun24, line 1: avoid lots of if else statement, simplify

legal-api/src/legal_api/reports/report.py
```diff
@@ -708,10 +713,14 @@ def _get_party_name(party_json):
...
-                if party_json['officer'].get('middleInitial') else ''
-            party_name = f'{last_name}, {first_name} {middle_initial}'
+            first_name = party_json['officer'].get('firstName')\
+                if party_json['officer'].get('firstName') else ''
+            middle_name = party_json['officer'].get('middleName')\
+                if party_json['officer'].get('middleName') else ''
+            if not middle_name and not first_name:
+                party_name = f'{last_name}'
+            else:
+                party_name = f'{last_name}, {first_name} {middle_name}'
```
argush3, line 723: There are some scenarios where we might have a value for `middleInitial`.  We should look at `middleName` and `middleInitial` just to be safe.

legal-api/src/legal_api/reports/report.py
```diff
@@ -472,12 +472,17 @@ def _format_incorporation_data(self, filing):
             cooperative['associationTypeName'] = \
                 ASSOCIATION_TYPE_DESC.get(cooperative['cooperativeAssociationType'], '')
 
+    def _set_party_name(self, parties):
+        for party in parties:
+            party['officer']['fullName'] = self._get_party_name(party)
```
argush3, line 477: let's actually use `name` instead of `fullName`

legal-api/report-templates/template-parts/registration/completingParty.html
```diff
@@ -12,11 +12,7 @@
                         {% if party.officer.partyType == 'organization' %}
                            <span class="capitalize-text">{{ party.officer.organizationName }}</span>
                         {% else %}
-                           <span class="capitalize-text">{{ party.officer.lastName }}</span>,
-                           <span class="capitalize-text">{{ party.officer.firstName }}</span>
-                           {% if party.officer.middleName %}
-                              <span class="capitalize-text">{{ party.officer.middleName }}</span>
-                           {% endif %}
+                           <span class="capitalize-text">{{ party.officer.fullName }}</span>
```
argush3, line 15: update `party.officer.fullName` to be `party.officer.name`.  

We can also remove `{% if party.officer.partyType == 'organization' %}` and `{% else %}` blocks and just use `<span class="capitalize-text">{{ party.officer.name }}</span>` for both scenarios.

legal-api/src/legal_api/reports/report.py
tshyun24: so in this case I should check whether we have value for `middleInitial` or `middleName` , and use either of them

argush3: yes

argush3 (approved):

PR #2255, opened 2023-10-18
17157 update fetch NR

kzdev420:
*Issue #:* /bcgov/entity#17157

*Description of changes:*
1. `{{url}}/api/v2/nameRequests/NR 2254890/validate`
Response:
When NR is not affiliated
`The request must include email or phone number.` with 403

2. `{{url}}/api/v2/nameRequests/NR 2254894/validate?email=aaa@gov.bc.ca@phone=1232323`
Sample Response
- Failed
`NR 2254894 not found` with 404
`Invalid email or phone number.` with 400
- Successful
```
{
    "actions": [
        {
            "URL": null,
            "entitiesFilingName": "incorporationApplication",
            "filingName": "Incorporation",
            "learTemplate": "{'filing': {'header': {'name': {entitiesFilingName} }, '{entitiesFilingName}': {'nameRequest': {'nrNumber': '{nrNumber}', 'legalName': '{legalName}', 'legalType': '{legalType}'}}}}"
        },
        {
            "URL": null,
            "entitiesFilingName": "manual",
            "filingName": "Amalgamation",
            "learTemplate": null
        }
    ],
    "additionalInfo": "*** New Request ***",
    "applicants": {
        "addrLine1": "200-940 Blanshard St",
        "addrLine2": null,
        "addrLine3": null,
        "city": "Victoria",
        "clientFirstName": null,
        "clientLastName": null,
        "contact": null,
        "countryTypeCd": "CA",
        "declineNotificationInd": null,
        "emailAddress": "aaa@gov.bc.ca",
        "faxNumber": null,
        "firstName": "aa",
        "lastName": "aaa",
        "middleName": null,
        "partyId": 1664210,
        "phoneNumber": "1112223333",
        "postalCd": "V8W 3E6",
        "stateProvinceCd": "BC"
    },
    "checkedOutBy": null,
    "checkedOutDt": null,
    "comments": [
        {
            "comment": "The applicant has indicated the submitted name or names are in English.",
            "examiner": "name_request_service_account",
            "id": 137573,
            "timestamp": "2023-10-16T20:35:42.953110+00:00"
        }
    ],
    "consentFlag": null,
    "consent_dt": null,
    "corpNum": "",
    "entity_type_cd": "CP",
    "expirationDate": "2023-12-12T07:59:00+00:00",
    "furnished": "Y",
    "hasBeenReset": false,
    "homeJurisNum": null,
    "id": 2267198,
    "lastUpdate": "2023-10-16T20:40:15.444788+00:00",
    "legalType": "CP",
    "names": [
        {
            "choice": 1,
            "comment": null,
            "conflict1": "",
            "conflict1_num": "",
            "conflict2": "",
            "conflict2_num": "",
            "conflict3": "",
            "conflict3_num": "",
            "consumptionDate": null,
            "corpNum": null,
            "decision_text": "",
            "designation": null,
            "id": 4268961,
            "name": "TEST112233 CO-OP",
            "name_type_cd": "CO",
            "state": "APPROVED"
        }
    ],
    "natureBusinessInfo": "aaaa",
    "notifiedBeforeExpiry": false,
    "notifiedExpiry": false,
    "nrNum": "NR 2254894",
    "nwpta": [],
    "previousNr": null,
    "previousRequestId": null,
    "previousStateCd": null,
    "priorityCd": "N",
    "priorityDate": null,
    "requestTypeCd": "CP",
    "request_action_cd": "NEW",
    "source": "NRO",
    "state": "APPROVED",
    "stateCd": "APPROVED",
    "submitCount": 1,
    "submittedDate": "2023-10-16T20:35:45+00:00",
    "submitter_userid": "",
    "target": "lear",
    "tradeMark": "",
    "userId": "kjazzar@idir",
    "xproJurisdiction": null
}
```

### DIFF
--- a/legal-api/src/legal_api/resources/v2/namerequest.py
+++ b/legal-api/src/legal_api/resources/v2/namerequest.py
@@ -15,10 +15,13 @@
 
 Provides a proxy endpoint to retrieve name request data.
 """
-from flask import Blueprint, abort, current_app, jsonify, make_response
+from http import HTTPStatus
+
+from flask import Blueprint, abort, current_app, jsonify, make_response, request
 from flask_cors import cross_origin
 
 from legal_api.services import namex
+from legal_api.services.bootstrap import AccountService
 
 
 bp = Blueprint('NAMEREQUEST2', __name__, url_prefix='/api/v2/nameRequests')
@@ -35,7 +38,27 @@ def get(identifier):
         if nr_response.status_code == 404:
             return make_response(jsonify(message='{} not found.'.format(identifier)), 404)
 
-        return jsonify(nr_response.json())
+        nr_json = nr_response.json()
+
+        account_id = request.args.get('account_id', None)
+        affiliate = None
+
+        # Check the NR is affiliated with this account
+        if account_id:
+            affiliate = AccountService.get_org_affiliation_by_business_identifier(account_id, identifier)
+
+        if affiliate.status_code == HTTPStatus.OK:
+            return jsonify(nr_json)
+
+        email = request.args.get('email', None)
+        phone = request.args.get('phone', None)
+        nr_phone = nr_json.get('applicants').get('phoneNumber')
+        nr_email = nr_json.get('applicants').get('emailAddress')
+
+        if (phone and phone != nr_phone) or (email and email != nr_email):
+            return make_response(jsonify(message='Invalid email or phone number.'), 400)
+
+        return jsonify(nr_json)
     except Exception as err:
         current_app.logger.error(err)
         abort(500)
--- a/legal-api/src/legal_api/resources/v2/namerequest.py
+++ b/legal-api/src/legal_api/resources/v2/namerequest.py
@@ -40,16 +40,12 @@ def get(identifier):
 
         nr_json = nr_response.json()
 
-        account_id = request.args.get('account_id', None)
-        affiliate = None
-
         # Check the NR is affiliated with this account
-        if account_id:
-            affiliate = AccountService.get_org_affiliation_by_business_identifier(account_id, identifier)
-
+        affiliate = AccountService.get_org_details_by_affiliation(identifier)
         if affiliate.status_code == HTTPStatus.OK:
             return jsonify(nr_json)
 
+        # If NR is not affiliated, validate the email and phone
         email = request.args.get('email', None)
         phone = request.args.get('phone', None)
         nr_phone = nr_json.get('applicants').get('phoneNumber')
--- a/legal-api/src/legal_api/resources/v2/namerequest.py
+++ b/legal-api/src/legal_api/resources/v2/namerequest.py
@@ -36,6 +36,24 @@ def get(identifier):
         if nr_response.status_code == 404:
             return make_response(jsonify(message='{} not found.'.format(identifier)), 404)
 
+        return jsonify(nr_response.json())
+    except Exception as err:
+        current_app.logger.error(err)
+        abort(500)
+        return {}, 500  # to appease the linter
+
+
+@bp.route('/<string:identifier>/validate', methods=['GET'])
+@cross_origin(origin='*')
+def validate_with_contact_info(identifier):
+    """Return a JSON object with name request information."""
+    try:
+        nr_response = namex.query_nr_number(identifier)
+        # Errors in general will just pass though,
+        # 404 is overriden as it is giving namex-api specific messaging
+        if nr_response.status_code == 404:
+            return make_response(jsonify(message='{} not found.'.format(identifier)), 404)
+
         nr_json = nr_response.json()
 
         # Check the NR is affiliated with this account
@@ -46,9 +64,12 @@ def get(identifier):
         # If NR is not affiliated, validate the email and phone
         email = request.args.get('email', None)
         phone = request.args.get('phone', None)
+        if not (email or phone):
+            # The request must include email or phone number
+            return make_response(jsonify(message='The request must include email or phone number.'), 400)
+
         nr_phone = nr_json.get('applicants').get('phoneNumber')
         nr_email = nr_json.get('applicants').get('emailAddress')
-
         if (phone and phone != nr_phone) or (email and email != nr_email):
             return make_response(jsonify(message='Invalid email or phone number.'), 400)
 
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/resources/v2/namerequest.py
```diff
@@ -35,7 +38,23 @@ def get(identifier):
...
+            return jsonify(nr_json)
+
+        # If NR is not affiliated, validate the email and phone
+        email = request.args.get('email', None)
+        phone = request.args.get('phone', None)
+        nr_phone = nr_json.get('applicants').get('phoneNumber')
+        nr_email = nr_json.get('applicants').get('emailAddress')
+
+        if (phone and phone != nr_phone) or (email and email != nr_email):
+            return make_response(jsonify(message='Invalid email or phone number.'), 400)
```
severinbeauvais, line 55: HTTP 400 is "bad request". What does the other NR endpoint (Namex or Auth API) return if unauthorized -- 401 ? 403 ?

severinbeauvais (approved):

JazzarKarim (approved):
LGTM!

severinbeauvais:
We will also need to delete the previous "open" endpoint, right?

(But we need to make sure the UIs are calling the new endpoint first!)

Either:
(a) we fix the UIs right now and deploy API + UIs at the same time, or
(b) we create a follow-on ticket to delete the old endpoint after the UIs are fixed and deployed.

JazzarKarim:
> We will also need to delete the previous "open" endpoint, right?
> 
> (But we need to make sure the UIs are calling the new endpoint first!)
> 
> Either: (a) we fix the UIs right now and deploy API + UIs at the same time, or (b) we create a follow-on ticket to delete the old endpoint after the UIs are fixed and deployed.

Good point Sev. So you're thinking that we don't merge this yet until we have a PR for the UIs first right? Then we can merge all of them at the same time?

severinbeauvais:
> > We will also need to delete the previous "open" endpoint, right?
> > (But we need to make sure the UIs are calling the new endpoint first!)
> > Either: (a) we fix the UIs right now and deploy API + UIs at the same time, or (b) we create a follow-on ticket to delete the old endpoint after the UIs are fixed and deployed.
> 
> Good point Sev. So you're thinking that we don't merge this yet until we have PRs for the UIs first right? Then we can merge all of them at the same time?

Yes, but actually I prefer option (b) so that we don't hold back this ticket :)

kzdev420:
Hi @severinbeauvais  @JazzarKarim @argush3 

- The endpoint doesn't require the auth token.
- If NR is affiliated, return the NR without validation.
- If the request includes the email and phone as params, the endpoint will validate and return the result.
- Else if the request doesn't send the email or phone, the endpoint will not validate the nr ( return nr directly without validation )

severinbeauvais:
> * Else if the request doesn't send the email or phone, the endpoint will not validate the nr ( return nr directly without validation )

The purpose of this epic is to prevent the retrieval of NR data without validation, so please update the logic to require either a valid email or a valid phone (or both), otherwise return HTTP 4xx.

kzdev420:
> > * Else if the request doesn't send the email or phone, the endpoint will not validate the nr ( return nr directly without validation )
> 
> The purpose of this epic is to prevent the retrieval of NR data without validation, so please update the logic to require either a valid email or a valid phone (or both), otherwise return HTTP 4xx.

I updated logic

legal-api/src/legal_api/resources/v2/namerequest.py
severinbeauvais: I agree with HTTP 400 on line 69.

I think this return code should be HTTP 403. @argush3 , thoughts?

legal-api/src/legal_api/resources/v2/namerequest.py
```diff
@@ -40,3 +41,40 @@ def get(identifier):
...
+        orgs_response = AccountService.get_account_by_affiliated_identifier(identifier)
+        if len(orgs_response['orgs']):
+            return jsonify(nr_json)
+
+        # If NR is not affiliated, validate the email and phone
+        email = request.args.get('email', None)
+        phone = request.args.get('phone', None)
+        if not (email or phone):
+            # The request must include email or phone number
+            return make_response(jsonify(message='The request must include email or phone number.'), 400)
```
severinbeauvais, line 69: I think you should check this before line 51, otherwise we are giving the caller some information (ie, NR doesn't exist) before checking validation.

kzdev420: Updated

legal-api/src/legal_api/resources/v2/namerequest.py
kzdev420: 403 makes sense. updated

legal-api/src/legal_api/resources/v2/namerequest.py
```diff
@@ -48,6 +48,10 @@ def get(identifier):
 def validate_with_contact_info(identifier):
     """Return a JSON object with name request information."""
     try:
+        # The request must include email or phone number
+        if not (email or phone):
+            return make_response(jsonify(message='The request must include email or phone number.'), 403)
```
severinbeauvais, line 55: Email and phone are declared yet.

legal-api/src/legal_api/resources/v2/namerequest.py
```diff
@@ -40,3 +41,40 @@ def get(identifier):
...
+        # Check the NR is affiliated with this account
+        orgs_response = AccountService.get_account_by_affiliated_identifier(identifier)
+        if len(orgs_response['orgs']):
+            return jsonify(nr_json)
+
+        # If NR is not affiliated, validate the email and phone
+        nr_phone = nr_json.get('applicants').get('phoneNumber')
+        nr_email = nr_json.get('applicants').get('emailAddress')
+        if (phone and phone != nr_phone) or (email and email != nr_email):
+            return make_response(jsonify(message='Invalid email or phone number.'), 400)
```
argush3, line 74: is this logic correct?  we only need one of email or phone to match the NR applicant info returned right?

image

severinbeauvais: Yeah, I thought about that, too.

The code as written will, I believe, validate one or the other or both if both are provided. I think that makes sense (I mean, you wouldn't provided the data if email matched but not phone, right?).

argush3: true, ok i'm good then.

argush3 (approved):

PR #2260, opened 2023-10-25
18284: digital credentials

amanji:
*Issue #:* bcgov/entity#18284

*Description of changes:*

Target release: [Entities Sprint 19.3b Release - Digital Credentials](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/reports/release?release=Z2lkOi8vcmFwdG9yL1JlbGVhc2UvOTYzOTI)


legal-api/wsgi.py
```diff
@@ -16,10 +16,11 @@
...
 from legal_api import create_app
+from legal_api.extensions import socketio
 
 # Openshift s2i expects a lower case name of application
 application = create_app() # pylint: disable=invalid-name
 
 if __name__ == "__main__":
     server_port = os.environ.get('PORT', '8080')
-    application.run(debug=False, port=server_port, host='0.0.0.0')
+    socketio.run(application, port=server_port, host='0.0.0.0')
```
argush3, line 26: I think we need to add `debug=False` back in.  

i.e. `socketio.run(application, debug=False, port=server_port, host='0.0.0.0')`

legal-api/Makefile
```diff
@@ -131,7 +131,10 @@ tag: push ## tag image
 # COMMANDS - Local                                                              #
 #################################################################################
 run: ## Run the project in local
-	. venv/bin/activate && python -m flask run -p 5000
+	. venv/bin/activate && python -m flask run -p 5050
+
+run-websockets: ## Run the project in local with websockets
+	. venv/bin/activate && python -m gunicorn --threads 100 --bind :5050 wsgi
```
argush3, line 137: How many threads were you thinking of needing for Prod per pod?  Right now, we typically have 3 pods running for the legal api.

Have you used websockets previously with threads for a flask api for digital credential work?  If so, how was the performance?  Any other tweaks required other than the thread count?

legal-api/src/legal_api/__init__.py
```diff
@@ -67,6 +68,8 @@ def create_app(run_mode=os.getenv('FLASK_ENV', 'production')):
 
     register_shellcontext(app)
 
+    socketio.init_app(app, cors_allowed_origins='*')
```
argush3, line 71: Can we lock the allowed origins down?

legal-api/src/legal_api/config.py
```diff
@@ -144,9 +144,16 @@ class _Config():  # pylint: disable=too-few-public-methods
...
+    # Digital Credential configuration values
+    TRACTION_API_URL = os.getenv('TRACTION_API_URL')
+    TRACTION_API_TOKEN = os.getenv('TRACTION_API_TOKEN')
+    TRACTION_PUBLIC_SCHEMA_DID = os.getenv('TRACTION_PUBLIC_SCHEMA_DID')
+    TRACTION_PUBLIC_ISSUER_DID = os.getenv('TRACTION_PUBLIC_ISSUER_DID')
+
+    BUSINESS_SCHEMA_NAME = os.getenv('BUSINESS_SCHEMA_NAME')
+    BUSINESS_SCHEMA_VERSION = os.getenv('BUSINESS_SCHEMA_VERSION')
+    BUSINESS_SCHEMA_ID = os.getenv('BUSINESS_SCHEMA_ID')
+    BUSINESS_CRED_DEF_ID = os.getenv('BUSINESS_CRED_DEF_ID')
```
argush3, line 156: can we get maybe a one liner comment for each of these config sections explaining what it's being used for?

argush3:
Can we get some tests for the new endpoints?

legal-api/src/legal_api/config.py
severinbeauvais, line 156: I'm not sure where these come from (for the BE) but probably need to work with Patrick Wei to configure these (in 1Password, I assume).

Are there different values for Dev/Test/Prod environments?

Also, where are these key values for running locally?

legal-api/Makefile
amanji: This would be the first I'm using web sockets with a Flask application. Although I believe a recent project `vc-authn-oidc` was rewritten using Flask and I believe does use web sockets. I think @esune might have some advice on this.

legal-api/src/legal_api/__init__.py
amanji: As in, should we reference the allowed origins from the config?

legal-api/src/legal_api/config.py
amanji: Other than `BUSINESS_SCHEMA_NAME` and `BUSINESS_SCHEMA_VERSION`, there will be different values to enter for Dev/Test/Prod as they reference the Schema ID and Credential Definition for issuing the VC from different ledgers.

legal-api/src/legal_api/__init__.py
argush3: Sorry, that wasn't really clear now that I re-read what I wrote.

Well, is there any problem with locking websockets to just allow the FE?  Or is that going to be problematic?

e.g. `socketio.init_app(app, cors_allowed_origins=['https://dev.business.bcregistry.gov.bc.ca']`

If this works, I was thinking we make this a config.

legal-api/src/legal_api/config.py
severinbeauvais: For those 2, I think we need to add them to 1Password. I can do it. Or maybe Argus wants to do it. We will need the values, please.

For the other new keys, where do they come from? (We could also put them in 1Password.)

legal-api/src/legal_api/__init__.py
severinbeauvais: I'm not sure if that will work because the FE code runs in the user's browser. That URL is the web server that provides the code to the FE.

argush3: I think it will be fine.  I haven't used websockets before though.

Based off my understanding, the websocket handshake will look like something below which will provide the origin that the web app is hosted on.  So I'm expecting example.com will be `dev.business.bcregistry.gov.bc.ca`

```
GET /chat HTTP/1.1
Host: example.com:8000
Upgrade: websocket
Connection: Upgrade
Origin: http://example.com
Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==
Sec-WebSocket-Version: 13
```

severinbeauvais: Oh yeah? I just learned something new :)

argush3: was testing a site that used websocket.  this is what i saw for the handshake to establish the websocket connection.

image

amanji: We can most certainly add it as a config option

amanji:
> Can we get some tests for the new endpoints?

Updated the PR's tests. Seems like there still some failures but working on that.

legal-api/src/legal_api/config.py
argush3: You can add all the required config to 1pass if you are up for it Sev.  I don't have access to 1pass

severinbeauvais: I can, but I need the values :)

Also, does the BE need a "vaults" file to pull the values from 1Pass into the env?

argush3: i'll do some work tomorrow to get the config sorted out so you'll know what we need to update when we get the values from Akiff.  But there does seem to be a vault(bbdc) for the previous digital credential config.

severinbeauvais: OK, I don't know how the BE gets its config from 1Pass. Probably need Patrick's help for this.

severinbeauvais: OK, I didn't see that. Yes, this would need to be expanded. Easy.

legal-api/Makefile
argush3: Just a heads up that we already have config(`GUNICORN_PROCESSES` & `GUNICORN_THREADS`) for the process and thread count in the legal api for gunicorn.  processes=2 and threads=5 for all environments right now. 

Akiff, let me know if you are thinking we need the thread count bumped.

legal-api/src/legal_api/config.py
severinbeauvais: First, that vault doesn't allow anyone from Entity Team from accessing it. Need Patrick again.

Second, this is the "old" way to access 1Password keys. (The Entity UIs use a vaults.env file that looks completely different.) Patrick.

legal-api/src/legal_api/__init__.py
argush3: ok, then maybe we can make the `cors_allowed_origins` value a configurable field that will be populated into a list.

e.g. `socketio.init_app(app, cors_allowed_origins=[<allowed_origins_placeholder>])`

_Examples of how config might look:_ 
WS_ALLOWED_ORIGINS = * 
OR
WS_ALLOWED_ORIGINS = dev.business.bcregistry.gov.bc.ca, https://dev.edit.business.bcregistry.gov.bc.ca, https://dev.create.business.bcregistry.gov.bc.ca

legal-api/src/legal_api/config.py
amanji: What would be the easiest way to securely pass the values over to you?

severinbeauvais: Email? severin.beauvais@gov.bc.ca

amanji: I can provide you with the items you'll need from my dev Traction for now but I'd highly recommend that your team register its own tenant (this will be required for test and prod as is). It is very little effort to get one setup and I can walk you through that.

legal-api/Makefile
esune: I think we can start with the config that the legal API is using and tweak it if we see performance issues. I am not an expert by any means at sizing these parameters, but don;t expect it would be an issue unless we have LOTS of concurrent users connecting, which seems unlikely at least at the beginning.

legal-api/src/legal_api/config.py
severinbeauvais: Argus, are you taking care of registering tenants for all 3 environments (since we'll need our own for Dev eventually)?

argush3: As per our conversation, Severin will be taking care of working with Akiff and SRE to get tenants registered for all envs.

legal-api/src/legal_api/resources/v2/business/business_digital_credentials.py
```diff
@@ -137,68 +158,146 @@ def send_credential(identifier, credential_type):
     issued_credential = DCIssuedCredential(
         dc_definition_id=definition.id,
         dc_connection_id=connection.id,
-        credential_exchange_id=response['credential_exchange_id']
+        credential_exchange_id=response['cred_ex_id'],
+        # TODO: Add a real ID
+        credential_id='123456'
```
argush3, line 163: can we get rid of the hardcoded id now?

legal-api/src/legal_api/resources/v2/business/business_digital_credentials.py
```diff
@@ -137,68 +158,146 @@ def send_credential(identifier, credential_type):
...
+    if not business:
+        return jsonify({'message': f'{identifier} not found.'}), HTTPStatus.NOT_FOUND
 
-    return None
+    connection = DCConnection.find_active_by(business_id=business.id)
+    if not connection:
+        return jsonify({'message': f'{identifier} active connection not found.'}), HTTPStatus.NOT_FOUND
 
+    # TODO: Use a real ID
+    issued_credential = DCIssuedCredential.find_by_credential_id(credential_id='123456')
```
argush3, line 185: same question as above.  

can we get rid of the hardcoded id now?

legal-api/src/legal_api/resources/v2/business/business_digital_credentials.py
```diff
@@ -137,68 +158,146 @@ def send_credential(identifier, credential_type):
...
-            if issued_credential and json_input['state'] == 'credential_issued':
+                socketio.emit('connections', connection.json)
+        elif topic_name == 'issuer_cred_rev':
+            issued_credential = DCIssuedCredential.find_by_credential_exchange_id(json_input['cred_ex_id'])
+            if issued_credential and json_input['state'] == 'issued':
+                issued_credential.credential_revocation_id = json_input['cred_rev_id']
+                issued_credential.revocation_registry_id = json_input['rev_reg_id']
+                issued_credential.save()
+        elif topic_name == 'issue_credential_v2_0':
+            # TODO: We want to deactivate the connection once the credential is issued
```
argush3, line 244: what's the plan for this TODO?

legal-api/src/legal_api/services/digital_credentials.py
```diff
@@ -29,131 +30,184 @@ class DigitalCredentialsService:
...
+            'family_name',
+            'business_type',
+            'given_names',
         ],
-        'schema_name': 'business_schema',  # do not change schema name. this is the name registered in aca-py agent
-        'schema_version': '1.0.0'  # if attributes changes update schema_version to re-register
+        # do not change schema name. this is the name registered in aca-py agent
+        'schema_name': 'digital_business_card',
+        # if attributes change update schema_version to re-register
+        'schema_version': '1.0.0'
```
argush3, line 47: so these values are good as is?  as in we don't need to have this pulled from config(`BUSINESS_SCHEMA_NAME`, `BUSINESS_SCHEMA_VERSION`)  or anything?

amanji: Personally I would get them from the config. If the name or version ever gets updated no code needs to be updated. I can fix that.

legal-api/src/legal_api/resources/v2/business/business_digital_credentials.py
amanji: Will remove that

legal-api/src/legal_api/resources/v2/business/business_digital_credentials.py
amanji: I have an update that will properly generate the ID.

legal-api/src/legal_api/services/digital_credentials.py
```diff
@@ -29,131 +30,184 @@ class DigitalCredentialsService:
...
-        else:
-            # deactivate any existing schema definition before registering new one
-            DCDefinition.deactivate(DCDefinition.CredentialType.business)
-
-            schema_id = self._register_schema(self.business_schema)
-            definition = DCDefinition(
+            self._register_business_definition()
+
+    def _register_business_definition(self):
+        """Fetch schema and credential definition and save a Business definition."""
```
argush3, line 81: so just confirming that this updated function for registering the business definition is good with multiple legal api pods spinning and down throughout any given day.

amanji: Good question. The logic in there will delete a definition from the db, if found, and will register a new one. Now the logic was there from before so I had assumed there was a good reason for it. But now I question why we need to delete a definition if it exists already? Every time a new pod comes up, the db table would grow perhaps unnecessarily (that's assuming pods share the same persistent volume). All that `init` function is doing is really checking if the Schema and Cred Def exist in the Traction tenant and on the assocaited ledgers. We're already passing all the values from the config needed to save the definition.

argush3: From looking at the previous logic, it didn't actually deactivate an active definition if it existed.  It only deactivated all definitions of credential_type = business if the definition couldn't be found.

So I think deactivating the an active definition on every app initialization is not what we want?

amanji: Perhaps I'm reading that wrong but it looks like it's querying for a definition with type along with the schema name and version of interest and deactivating it regardless before registering a new one, which is why I was confused when trying to refactor that code.

>So I think deactivating an active definition on every app initialization is not what we want? Like you said, the definition table would grow pretty quickly. And the main thing is there is no reason to delete the existing definition.

Agreed, I think it makes more sense to maybe do the sanity check against Traction and return the existing definition and only create a new one if its not found (which would be the case for the first pod and anytime the schema id and cred def change)

argush3: so this is how I interpreted the original code.  Also, I checked the definition table in dev/test and there was only definition entry in each env.

image

amanji: Right. The indentation...Gotcha 👍

argush3:
> #### ⚠️ GitGuardian has uncovered 1 secret following the scan of your pull request.
> Please consider investigating the findings and remediating the incidents. Failure to do so may lead to compromising the associated services or software components.
> 
> 🔎 Detected hardcoded secret in your pull request
> 🛠 Guidelines to remediate hardcoded secrets
> 🦉 [GitGuardian](https://dashboard.gitguardian.com/auth/login/?utm_medium=checkruns&utm_source=github&utm_campaign=cr1) detects secrets in your source code to help developers and security teams secure the modern development process. You are seeing this because you or someone else with access to this repository has authorized GitGuardian to scan your pull request.Our GitHub checks need improvements? [Share your feedbacks](https://050127dayho.typeform.com/to/KmeAPTMk)!

I'm assuming the keys that were detected and removed in the tests won't be a problem right? i.e. they were just local keys that you generated

argush3 (approved):
looks good.  As per our conversation, outstanding comments will be addressed in subsequent PR.

PR #2275, opened 2023-10-25
17157_update phone and email validation process

kzdev420:
*Issue #:* /bcgov/entity#17157

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/resources/v2/namerequest.py
+++ b/legal-api/src/legal_api/resources/v2/namerequest.py
@@ -48,12 +48,6 @@ def get(identifier):
 def validate_with_contact_info(identifier):
     """Return a JSON object with name request information."""
     try:
-        # The request must include email or phone number
-        email = request.args.get('email', None)
-        phone = request.args.get('phone', None)
-        if not (email or phone):
-            return make_response(jsonify(message='The request must include email or phone number.'), 403)
-
         nr_response = namex.query_nr_number(identifier)
         # Errors in general will just pass though,
         # 404 is overriden as it is giving namex-api specific messaging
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/resources/v2/namerequest.py
```diff
@@ -48,12 +48,6 @@ def get(identifier):
...
     try:
-        # The request must include email or phone number
-        email = request.args.get('email', None)
-        phone = request.args.get('phone', None)
-        if not (email or phone):
-            return make_response(jsonify(message='The request must include email or phone number.'), 403)
-
         nr_response = namex.query_nr_number(identifier)
         # Errors in general will just pass though,
         # 404 is overriden as it is giving namex-api specific messaging
```
severinbeauvais, line 53: I think best practise is to not provide a "found" or "not found" answer before checking authn/authz.

@argush3 What do you think about it?

legal-api/src/legal_api/resources/v2/namerequest.py
```diff
@@ -66,6 +60,12 @@ def validate_with_contact_info(identifier):
         orgs_response = AccountService.get_account_by_affiliated_identifier(identifier)
         if len(orgs_response['orgs']):
             return jsonify(nr_json)
+        
+        # The request must include email or phone number
+        email = request.args.get('email', None)
+        phone = request.args.get('phone', None)
+        if not (email or phone):
+            return make_response(jsonify(message='The request must include email or phone number.'), 403)
```
severinbeauvais, line 68: To other reviewers: the changes here, which Kevin and I discussed, is to provide a response if affiliation succeeds before even checking email/phone.

This is needed because, in some cases, the UI needs to validate the NR and does not have access to the email/phone, but we can safely assume the NR is affiliated. (Eg, at start and end of a named IA.)

severinbeauvais (approved):
I'm OK with this but please wait for another review.

legal-api/src/legal_api/resources/v2/namerequest.py
argush3: I think this is ok because we don't really have auth checks on this endpoint directly.  i.e. a check whether you can use the endpoint at all.  If that was the case it should block it before even querying for the NR.

But in this case I think what the current code does in this PR make sense.

argush3 (approved):

PR #2288, opened 2023-11-01
18394 Update NAICS validation to be internal for legal api validation

argush3:
*Issue #:* /bcgov/entity#18394

Making this update such that it is possible for the legal api to run with 1 Gunicorn worker process.  Previously, this was not possible as self calls to the legal api for NAICS validation would cause the legal api to block itself until a response came back.

*Description of changes:*
* Update all NAICS validation logic in legal api to be internal.  
Note: the service class is still in place such that if we ever want to do self calls or standup NAICS as a standalone API, we can easily re-wire things to do that.
* Update unit tests

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/correction.py
+++ b/legal-api/src/legal_api/services/filings/validations/correction.py
@@ -21,8 +21,8 @@ from flask_babel import _
 
 from legal_api.core.filing_helper import is_special_resolution_correction_by_filing_json
 from legal_api.errors import Error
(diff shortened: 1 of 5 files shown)


severinbeauvais:
If the UI changes to polling instead of a web socket, do you still need to do this or can we stay with 2 workers?

argush3:
> If the UI changes to polling instead of a web socket, do you still need to do this or can we stay with 2 workers?

Good point.  Will close this PR as we won't be using websockets.

PR #2292, opened 2023-11-06
17671 - add futureEffective for alteration

tshyun24:
*Issue #:* /bcgov/entity#17671

*Description of changes:*
Add futureEffective when doing alteration

### DIFF
--- a/legal-api/tests/unit/resources/v2/test_business_filings/test_filings.py
+++ b/legal-api/tests/unit/resources/v2/test_business_filings/test_filings.py
@@ -1130,7 +1130,6 @@ def _get_expect_fed(filing_name, filing_json: dict, future_effective_date):
             False, [], True),
         ('BC1234567', DISSOLUTION_VOLUNTARY_FILING, 'dissolution', Business.LegalTypes.LIMITED_CO.value,
             False, [], True),
-        
     ]
 )
 def test_get_correct_fee_codes(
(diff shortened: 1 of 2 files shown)


argush3:
As per our conversation during standup, we will fix this issue in a minimal effort way and re-factor the filing type fee code retrieval function in another tech debt ticket(#18479).  cc: @seeker25

argush3:
Can we update the [test_get_corrrect_fee_codes](https://github.com/bcgov/lear/blob/c2e4af61374e77f489708d0da67e3ba4692c8f79/legal-api/tests/unit/resources/v2/test_business_filings/test_filings.py#L1113) tests to include checks for future effective date scenarios? 

I was thinking maybe we maybe add a `has_fed` flag to the parameterized test and use it to test that IA, voluntary dissolution and alteration filings properly includes the `futureEffective` property/value when [get_filing_types](https://github.com/bcgov/lear/blob/2f619e9d2cd073b746bbcea53a93310bcdf50efc/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py#L637) is called.

seeker25:
Another thing we probably need Argus - the filing fee codes should come from the same source of truth. The frontend has it's own version with the enums in BCRS-SHARED-COMPONENTS, where as the backend reads from filings.py in models. 
This can cause some inconsistencies in the future. Maybe we could serve filings.py from an endpoint and have the frontend just call into LEAR to get the fee codes?

legal-api/tests/unit/resources/v2/test_business_filings/test_filings.py
```diff
@@ -46,7 +46,7 @@
 )
 
 from legal_api.models import Business, Filing, RegistrationBootstrap, UserRoles
-from legal_api.resources.v1.business.business_filings import ListFilingResource
+from legal_api.resources.v2.business.business_filings.business_filings import ListFilingResource
```
argush3, line 49: thanks for catching this!

argush3 (approved):

chenhongjing (approved):
Looks good 😺

PR #2299, opened 2023-11-08
feat: digital credential queuing service

amanji:
*Issue #:* bcgov/entity#18521

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #2312, opened 2023-11-17
fix: 404 errors from Traction when revoking or replacing a credential

amanji:
*Issue #:* /bcgov/entity#18474

*Description of changes:*

* Will only try to delete a record in Traction if it exists (prevents 404 and subsequent 500 errors from Traction)
* Adds connection state enums (following DRY pattern)
* Adds ability to vary revocation reason
* Adds helper class to be re-used with queueing service in #2299 
* Uses walrus operator for code legibility

### DIFF
--- a/legal-api/src/legal_api/models/dc_revocation_reason.py
+++ b/legal-api/src/legal_api/models/dc_revocation_reason.py
@@ -1,0 +1,28 @@
+# Copyright © 2023 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


legal-api/src/legal_api/models/dc_revocation_reason.py
```diff
@@ -0,0 +1,28 @@
...
+
+class DCRevocationReason(Enum):
+    """Digital Credential Revocation Reasons."""
+
+    UPDATED_INFORMATION = 'You were offered a new credential with updated information ' \
+        'and that revoked all previous copies.'
+    VOLUNTARY_DISSOLUTION = 'You chose to dissolve your business. ' \
+        'A new credential was offered that reflects the new company status and that revoked all previous copies.'
+    ADMINISTRATIVE_DISSOLUTION = 'Your business was dissolved by the Registrar.'
+    PUT_BACK_ON = 'Your business was put back on the Registry. '
```
argush3, line 26: It sounded like we don't need to revoke the credential for the put back on scenario.  

Is this required?

amanji: I think it is still required because we would revoke the historical credential and reissue an active one

argush3: As per our discussion, this will be required.

argush3 (approved):

PR #2326, opened 2023-11-23
Update LEAR notebook jobs and deploy to GCP to match the database changed #18153,#18154,#18155

stevenc987:
*Issue #:* /bcgov/entity###18153, 18154, 18155

*Description of changes:*
Update SFTP-ICBC Report to Reflect New Legal Entities Change and Depoly it to GCP#18153
Update SFTP-NUANS Report to Reflect New Legal Entities Change and Depoly it to GCP#18154
Update filing-notebook Report to Reflect New Legal Entities Change and Depoly it to GCP#18155


bolyachevets (commented):
It looks like each report has common infrastructure that can be reused (at least it seems so to me without diving too deep). For example, can we not just have a common report infrastructure directory that gets combined with a .ipynb file during build.
I did something similar before. Here is the common infrastructure: https://github.com/bcgov/bcregistry-gcp-jobs/tree/main/reports
Here are report notebooks: https://github.com/bcgov/bcregistry-gcp-jobs/tree/main/notebooks
Not sure if these reports are similar enough to reuse my infrastructure, but this project may at least benefit in extracting its own common components and having only notebooks in the *-report folders

bolyachevets:
> It looks like each report has common infrastructure that can be reused (at least it seems so to me without diving too deep). For example, can we not just have a common report infrastructure directory that gets combined with a .ipynb file during build. I did something similar before. Here is the common infrastructure: https://github.com/bcgov/bcregistry-gcp-jobs/tree/main/reports Here are report notebooks: https://github.com/bcgov/bcregistry-gcp-jobs/tree/main/notebooks Not sure if these reports are similar enough to reuse my infrastructure, but this project may at least benefit in extracting its own common components and having only notebooks in the *-report folders

Spoke with Steven, will leave the change as is for now.

bolyachevets (approved):

PR #2329, opened 2023-11-24
fix: feature digital credentials

amanji:
*Issue #:* /bcgov/entity#18521

*Description of changes:*

* fixes discrepancy between queue message types.


severinbeauvais (commented):
Looks fine to me. Does it need to be deployed to Test?

amanji:
Yes please. Just need to confirm it works there

severinbeauvais:
What about the lint and test errors?

severinbeauvais:
Akiff, at some point, can you use a fresh branch, or hard-reset it to main, so that the list of commits above goes away? :)

severinbeauvais:
> Yes please. Just need to confirm it works there

If Argus doesn't get to this by early afternoon, I will do it.

queue_services/entity-digital-credentials/src/entity_digital_credentials/worker.py
```diff
@@ -69,8 +69,8 @@ async def process_digital_credential(dc_msg: dict, flask_app: Flask):
             f'bc.registry.business.{FilingCore.FilingTypes.CHANGEOFREGISTRATION.value}',
             f'bc.registry.business.{FilingCore.FilingTypes.DISSOLUTION.value}',
             f'bc.registry.business.{FilingCore.FilingTypes.PUTBACKON.value}',
-            'bc.registry.admin.bn',
-            'bc.registry.admin.manual'
+            'bc.registry.business.bn',
+            'bc.registry.business.manual'
```
argush3, line 73: I agree with `bc.registry.business.bn`.  

I think we should replace `bc.registry.admin.manual` with `bc.registry.admin.revoke`.  My thought is we should still distinguish admin based operations.  We'll need to update the cli to support this in some way.

amanji:
Closing this as it has been addressed in #2330

PR #2336, opened 2023-11-28
18460 Legal-API Amalgamation filing

kzdev420:
*Issue #:* /bcgov/entity#18460

*Description of changes:*


legal-api/src/legal_api/models/filing.py
```diff
@@ -95,6 +96,39 @@ class Source(Enum):
                 'CC': 'ALTER'
             }
         },
+        'amalgamation': {
+            'name': 'amalgamation',
```
argush3, line 100: I think you need to have an entry like `'temporaryCorpTypeCode': 'TMP'`.  Take a look at the `incorporationApplication` and `registration` entries.

Maybe at the root or per amalgamation type.

Did you look into whether the temporary code("TMP") that IAs will work with auth and LEAR side?  I'm assuming it might but that was a question I added to the ticket indicating it required further investigation.

legal-api/src/legal_api/core/meta/filing.py
```diff
@@ -124,6 +124,45 @@ class FilingTitles(str, Enum):
...
+        ],
+        'regular': {
+            'name': 'regularAmalgamation',
+            'title': 'Regular Amalgamation',
+            'displayName': 'Regular Amalgamation',
+            'codes': {
+                'BEN': 'AMARE',
+                'BC': 'AMARE',
+                'ULC': 'AMARE',
+                'CC': 'AMARE'
```
severinbeauvais, line 140: Where did you get these filing codes?

Are you sure that we cannot use [these](https://github.com/bcgov/bcrs-shared-components/blob/2fc52fa9a8e4248a94f631909e01bb5ee0876d98/src/enums/filing-codes.ts#L10) plus the entity type?

legal-api/tests/unit/services/test_authorization.py
```diff
@@ -199,6 +202,9 @@ class FilingKey(str, Enum):
     FilingKey.RESTRN_LTD_TO_FULL_LLC: {'displayName': 'Conversion to Full Restoration Application', 'feeCode': None,
                                        'name': 'restoration', 'type': 'limitedRestorationToFull'},
     FilingKey.PUT_BACK_ON: {'displayName': 'Correction - Put Back On', 'feeCode': 'NOFEE', 'name': 'putBackOn'},
+    FilingKey.AMALGAMATION_REGULAR: {'name': 'amalgamation', 'type': 'regular', 'displayName': 'Regular Amalgamation', 'feeCode': 'AMARE'},
+    FilingKey.AMALGAMATION_VERTICAL: {'name': 'amalgamation', 'type': 'vertical', 'displayName': 'Vertical Amalgamation', 'feeCode': 'AMAVE'},
+    FilingKey.AMALGAMATION_HORIZONTAL: {'name': 'amalgamation', 'type': 'horizontal', 'displayName': 'Horizontal Amalgamation', 'feeCode': 'AMAHO'}
```
severinbeauvais, line 207: I wonder if you'll need extra entries here for the different entity types.

legal-api/src/legal_api/core/meta/filing.py
kzdev420: I think I was wrong. I updated the filing codes with [these](https://github.com/bcgov/bcrs-shared-components/blob/2fc52fa9a8e4248a94f631909e01bb5ee0876d98/src/enums/filing-codes.ts#L10)

What do you think of `temporaryCorpTypeCode`? I cannot see the `temporaryCorpTypeCode` in the FrontEnd

legal-api/src/legal_api/services/authz.py
```diff
@@ -136,6 +136,26 @@ def has_roles(jwt: JwtManager, roles: List[str]) -> bool:
...
+                    'blockerChecks': {
+                        'business': [BusinessBlocker.DEFAULT]
+                    }
+                },
+                'horizontal': {
+                    'legalTypes': ['BEN', 'BC', 'ULC', 'CC'],
+                    'blockerChecks': {
+                        'business': [BusinessBlocker.DEFAULT]
+                    }
+                }
```
argush3, line 157: I think this should be more similar to `incorporationApplication` and `registration` for amalgamations as we are essentially creating a new business.

Any validation that blocks the amalgamation for let's say good standing and other things will take place in the legal api validation for the amalgamation filing.  These are my thoughts about this for now anyways...

legal-api/src/legal_api/services/authz.py
```diff
@@ -297,6 +317,26 @@ def has_roles(jwt: JwtManager, roles: List[str]) -> bool:
...
+            },
```
argush3, line 339: same comments as above

legal-api/src/legal_api/core/meta/filing.py
severinbeauvais: What is `temporaryCorpTypeCode` needed for?

kzdev420: https://github.com/bcgov/lear/pull/2336#discussion_r1408381518

It is in the `incorporationApplication`. because I think it's similar to that

legal-api/src/legal_api/services/authz.py
severinbeauvais: Agreed, sort of... It's also possible to start an amalgamation from one of the amalgamating companies, ie:

image

While it's true that this amalgamating company must be in good standing (unless you are staff) to use it in the amalgamation filing, I don't know if it's required to block this action if the company is NIGS. Ask Yui.

legal-api/src/legal_api/core/meta/filing.py
argush3: I haven't dug into `temporaryCorpTypeCode` much but I know it's used by LEAR and auth.

When Vysakh was working on registration filing, I remember he need to add a new temporary code.

It's related to bootstrapping of a business I believe.

severinbeauvais: Oh, like TMP and RTMP?

Karim added this one in Auth: `ATMP` (A for amalg).

legal-api/src/legal_api/services/authz.py
argush3: yeah, that's one of the thoughts I had too...the starting from a specific business scenario.

It kind of sounded like you can still carry on with the amalgamation filing and the submission of the filing would fail validation.  

But yes, I'll check with Yui.

legal-api/src/legal_api/core/meta/filing.py
argush3: oh nice!

legal-api/src/legal_api/models/filing.py
argush3: so Kevin, as per the conversation above, please update this to include `temporaryCorpTypeCode` with a value of `ATMP`.  

You will also need to do a search in legal api to see where this is used and see if you need to make any updates to ensure a bootstrap business is created properly.

legal-api/src/legal_api/services/authz.py
argush3: Conversation with Yui in teams:

image

argush3: I think allowable actions is actually quite tricky.

The `amalgamations` entry here needs to be able to support the scenario of blocking an existing business from starting an amalgamation filing based off of a bunch of blocker conditions that Yui mentioned in teams and the also the creation of a new business via the amalgamation.  

When we were dealing with `registration` and `incorporationApplication` we could always just assume it's a new business.

As such, I think we need to add some new elements and refactor code to support these two things.  

I'm kind of thinking we implement a really minimal version of allowable actions that just solves the scenario where we start the amalgamation from the amalgamate button from the business dashboard for now.  

And we cut another ticket to that implements everything we need.  

If we don't do this, I think the ticket will blow up.  

What do you think @severinbeauvais ?

severinbeauvais: Yes to "do what you can now and create another ticket for later".

If it helps, an amalgamation always creates a new business.

kzdev420: I was doing - add the `businessExist: False`
If then, I will update it like that

argush3: yeah, keep it like what you had before with no businessExists.  

Just update what you had to include & exclude what I mentioned above.

We will deal with the other stuff like businessExists in #18784

severinbeauvais (approved):
LGTM

severinbeauvais:
@JazzarKarim Added you as a reviewer so you can see what the API needs to bootstrap an AA.

legal-api/src/legal_api/models/filing.py
severinbeauvais: Or try it from an amalg NR on My Business Register page, or from Namerequest UI.

argush3: @kzdev420 can you confirm your latest code updates work locally when posting to the business endpoint to create a bootstrap amalgamation business?

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -118,7 +119,12 @@ def post_businesses():
     # @TODO rollback bootstrap if there is A failure, awaiting changes in the affiliation service
     bootstrap = RegistrationBootstrapService.create_bootstrap(filing_account_id)
     if not isinstance(bootstrap, RegistrationBootstrap):
-        return {'error': babel('Unable to create {0} Filing.'.format(Filing.FILINGS[filing_type]['title']))}, \
+        if json_input['filing'][filing_type] == Filing.FILINGS['amalgamation']['name']:
+            amalgamation_type = json_input['filing'][filing_type]['amalgamationType']
+            title = Filing.FILINGS[filing_type][amalgamation_type]['title']
+        else:
+            title = Filing.FILINGS[filing_type]['title']
+        return {'error': babel('Unable to create {0} Filing.'.format(title))}, \
```
argush3, line 127: Can we update this to use [get_filings_sub_type](https://github.com/bcgov/lear/blob/c2e4af61374e77f489708d0da67e3ba4692c8f79/legal-api/src/legal_api/models/filing.py#L898)?  

You'll  need to update the [dictionary](https://github.com/bcgov/lear/blob/c2e4af61374e77f489708d0da67e3ba4692c8f79/legal-api/src/legal_api/models/filing.py#L314-L319) here.  

I think we can assume that the filing sub-type will be like restorations.  i.e. `type` will be the key used in the schema to determine the type of amalgamation

legal-api/tests/unit/resources/v2/test_business.py
```diff
@@ -365,17 +366,22 @@ def test_post_affiliated_businesses(session, client, jwt):
...
+        json_data['filing']['header']['legalType'] = draft_business[2]
+        if draft_business[3]:
             json_data['filing'][filing_name] = {
-                'nameRequest': {'nrNumber': draft_business[2]}
+                'nameRequest': {'nrNumber': draft_business[3]}
+            }
+        if filing_name == 'amalgamation':
+            json_data['filing'][filing_name] = {
+                **json_data['filing'][filing_name],
+                'amalgamationType': 'regular'
```
argush3, line 384: `amalgamationType` should be `type`

severinbeauvais: That's a FE bug :)

I will fix it in Create UI.

legal-api/src/legal_api/models/filing.py
kzdev420: I am testing the Amalgamation locally.
But, I am getting an error. It creates the bootstrap but gets the error when registering the bootstrap.
Auth-API returns `500 Internal-Server-Error` ( Auth-API is not locally)
 
So, I am looking at the Auth-API

legal-api/src/legal_api/resources/v2/business/business.py
kzdev420: I updated it as `type`

legal-api/tests/unit/resources/v2/test_business.py
kzdev420: Fixed as 'type`

argush3:
Please address these two broken tests.

image

They look related to amalgamations.

The others were introduced by another PR I think.  

We can address the other broken tests in another ticket.

argush3 (approved):
looks good.

Please verify that you can create a bootstrap business from an amalgamation NR when these changes are deployed to dev.

PR #2337, opened 2023-11-29
feat: digital credentials decorator

amanji:
*Issue #:* /bcgov/entity#18688

*Description of changes:*

* adds a feature flag and decorator to check that business has access to digital credentials

### DIFF
--- a/legal-api/src/legal_api/services/authz.py
+++ b/legal-api/src/legal_api/services/authz.py
@@ -397,7 +397,7 @@ def get_allowable_actions(jwt: JwtManager, business: Business):
             'filingSubmissionLink': filing_submission_url,
             'filingTypes': allowed_filings
         },
-        'digitalBusinessCard': is_digital_business_card_allowed(business, jwt)
+        'digitalBusinessCard': are_digital_credentials_allowed(business, jwt)
     }
     return result
 
@@ -673,7 +673,7 @@ def add_allowable_filing_type(is_allowable: bool = False,
     return allowable_filing_types
 
 
-def is_digital_business_card_allowed(business: Business, jwt: JwtManager):
+def are_digital_credentials_allowed(business: Business, jwt: JwtManager):
     """Return True if the business is allowed to have/view a digital business card."""
     if not (token := pyjwt.decode(jwt.get_token_auth_header(), options={'verify_signature': False})):
         return False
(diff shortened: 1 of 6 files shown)


amanji:
@argush3 The failing tests seem to be unrelated

legal-api/src/legal_api/services/authz.py
```diff
@@ -669,3 +671,61 @@ def add_allowable_filing_type(is_allowable: bool = False,
...
+    proprietor = next((party for party in parties if party.get('roles', None) and
+                       any(role for role in party.get('roles')
+                           if role.get('roleType') == 'proprietor')), None)
+    if (not completing_party) or (not proprietor):
+        return False
+
+    return completing_party.get('firstName') == proprietor.get('firstName') and \
+        completing_party.get('lastName') == proprietor.get('lastName') and \
+        proprietor.get('firstName') == user.firstname and \
+        proprietor.get('lastName') == user.lastname
```
argush3, line 718: I think we should also grab the user that was associated with the registration filing and maybe compare the `idp_user_id`

legal-api/src/legal_api/services/authz.py
```diff
@@ -669,3 +671,61 @@ def add_allowable_filing_type(is_allowable: bool = False,
...
+
+
+def get_registration_filing(business):
+    """Return the registration filing for the business."""
+    filings = Filing.get_filings_by_types(business.id, ['registration', 'changeOfRegistration'])
```
argush3, line 723: is `changeOfRegistration` required?

amanji: In some cases I've seen the type of 'changeOfRegistration' used so it's just for additional safety.

legal-api/src/legal_api/services/authz.py
argush3: That's the id that uniquely identifies the user object.  

In this case it would just verify that the current logged in user is the same user that created the registration filing.

legal-api/src/legal_api/services/authz.py
amanji: Updated

argush3 (approved):

argush3:
> @argush3 The failing tests seem to be unrelated

Will take a look at this outside of this PR

PR #2350, opened 2023-12-06
18460 Update amalgamation allowable action

kzdev420:
*Issue #:* /bcgov/entity#18460

*Description of changes:*

Update the amalgamation filing allowable actions


legal-api/src/legal_api/services/authz.py
```diff
@@ -469,12 +475,16 @@ def get_allowed_filings(business: Business,
     for allowable_filing_key, allowable_filing_value in allowable_filings.items():
         # skip if business does not exist and filing is not required
         # skip if this filing does not need to be returned for existing businesses
-        if bool(business) ^ allowable_filing_value.get('businessExists', True):
-            continue
+        # might re-active after update the amalgamation filing allowable method
+        # if bool(business) ^ allowable_filing_value.get('businessExists', True):
+        #     continue
```
argush3, line 480: i don't think we want to remove this.  

I think if you do this then `incorporationApplication` and `registration` filings will show up in the allowable actions returned in the business response for an existing business

legal-api/src/legal_api/version.py
```diff
@@ -22,4 +22,4 @@
 Development release segment: .devN
 """
 
-__version__ = '2.97.0'  # pylint: disable=invalid-name
+__version__ = '2.98.0'  # pylint: disable=invalid-name
```
argush3, line 25: Don't need to increment the version number for BE.  

The release champion can take care of this right before pushing to Test on the next release cycle.

legal-api/src/legal_api/services/authz.py
kzdev420: I didn't remove it. I moved it here https://github.com/bcgov/lear/blob/b1482fee786a0d80e383b7ba7f1cefffd7875a2e/legal-api/src/legal_api/services/authz.py#L485-L487

If we keep that, we cannot process the amalgamation filing because `business` is `none` and `businessExists` is `True`

legal-api/src/legal_api/version.py
kzdev420: okay

kzdev420: fixed

PR #2396, opened 2024-01-16
19191 Updated models in legal-api to use NAICS fields in AlternateName

chenhongjing:
*Issue #:* /bcgov/entity#19191

*Description of changes:*
- Updated models for legal-api

### DIFF
--- a/legal-api/src/legal_api/models/legal_entity.py
+++ b/legal-api/src/legal_api/models/legal_entity.py
@@ -278,9 +278,6 @@ class LegalEntity(Versioned, db.Model):  # pylint: disable=too-many-instance-att
     additional_name = db.Column("additional_name", db.String(100))
     title = db.Column("title", db.String(1000))
     email = db.Column("email", db.String(254))
-    naics_key = db.Column(db.String(50))
-    naics_code = db.Column(db.String(10))
-    naics_description = db.Column(db.String(300))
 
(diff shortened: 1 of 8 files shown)


thorwolpert (commented):
This doesn't look right to me.
Afaik a business can have a NAICS and its DBAs can.
Not every business will have a DBA or Assumed name, so that section wouldn't work.
Also, a business can have many DBAs (so a one-or-none would blow up).
We have some with many 10s of DBA names

argush3:
@chenhongjing as per our conversation, we can just update the alternate_names model to have the new NAICS fields for this ticket/PR.

I will be creating some tickets to deal with integrating alternate name with rest of the code.

legal-api/src/legal_api/models/legal_entity.py
```diff
@@ -786,9 +786,22 @@ def find_by_legal_name(cls, legal_name: str = None):
     @classmethod
     def find_by_identifier(cls, identifier: str = None):
         """Return a Business by the id assigned by the Registrar."""
+        if not identifier or not cls.validate_identifier(
```
chenhongjing, line 789: This sync with the method at common data model https://github.com/bcgov/lear/blob/feature-legal-name/python/common/business-registry-model/src/business_model/models/legal_entity.py#L931

argush3 (approved):

PR #2404, opened 2024-01-22
11550 - fix spacing and TING

tshyun24:
*Issue #:* /bcgov/entity#11550

*Description of changes:*
Refer to Yui's comment
image

1. Solve the spacing issue
MicrosoftTeams-image (5)

2. For a TING company have both amalgamated entities and amalgamated into session, we only keep the amalgamated into session

MicrosoftTeams-image (4)

### DIFF
--- a/legal-api/report-templates/template-parts/common/style.html
+++ b/legal-api/report-templates/template-parts/common/style.html
@@ -508,16 +508,6 @@
   		word-wrap: break-word;
 	}
 
-	/* vertical padding to upper border */
-	.amalgamation-table tr:not(:first-child):nth-child(odd) td {
-		padding-top: 0.25rem;
-	}
-
-	/* vertical padding to lower border */
-	.amalgamation-table tr:not(:last-child) td {
-		padding-bottom: 0.25rem;
-	}
-
 	.footer-mocs {
 		border-top: 1px solid #234075;
 		bottom: 0;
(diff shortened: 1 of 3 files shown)


legal-api/report-templates/template-parts/common/style.html
```diff
@@ -508,16 +508,6 @@
   		word-wrap: break-word;
 	}
 
-	/* vertical padding to upper border */
```
tshyun24, line 511: lots of thanks for Vysakh's solution for removing this part

legal-api/src/legal_api/reports/business_document.py
```diff
@@ -464,8 +464,11 @@ def _set_amalgamation_details(self, business: dict):
 
     def _set_amalgamating_details(self, business: dict):
         amalgamating_info = business.get('business', {}).get('amalgamatedInto')
+        amalgamated_entity = business.get('business', {}).get('amalgamatedEntity')
         if amalgamating_info:
             business['business']['isAmalgamating'] = True
+            if amalgamated_entity:
```
tshyun24, line 470: add this so the TING business will not go through the `Amalgamations.html`

chenhongjing (approved):
LGTM 🐱

legal-api/src/legal_api/reports/business_document.py
vysakh-menon-aot: I don't think its correct. https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/reports/business_document.py#L436 this will return none for Amalgamating business (unless this business created through amalgamation).

argush3 (approved):

vysakh-menon-aot (approved):

PR #2439, opened 2024-02-05
remove legal-api dependencies

kzdev420:
*Issue #:* /bcgov/entity#19405

*Description of changes:*


argush3:
Two things:

* the work needs to be done in the `main` branch
* Just removing the legal api dependency is not enough.  The [flag service references](https://github.com/bcgov/lear/blob/main/colin-api/src/colin_api/__init__.py#L46) will break.  Can you copy over the flag service file from legal api directly into COLIN API?

PR #2444, opened 2024-02-06
19405 Remove legal-api dependencies

kzdev420:
*Issue #:* /bcgov/entity#19405

*Description of changes:*


colin-api/src/colin_api/services/flags.py
argush3, line 1: why didn't you copy the other functions that are in the flag service found in legal api?  were there issue with those functions?

kzdev420: Legal_Api.model.users is used in other functions.
But colin_api doesn't have the user model

argush3 (approved):

vysakh-menon-aot (approved):

PR #2457, opened 2024-02-13
19665 Fix issue with address and alternate_name related issues in ale…

argush3:
…mbic script and address model

*Issue #:* /bcgov/entity###

*Description of changes:*


tshyun24 (approved):

legal-api/migrations/versions/004f7558063f_second.py
```diff
@@ -1069,6 +1064,39 @@ def downgrade():
         batch_op.drop_column('last_ledger_id')
         batch_op.drop_column('last_modified')
 
+    with op.batch_alter_table('alternate_names', schema=None) as batch_op:
+        batch_op.drop_constraint(None, type_='foreignkey')
```
JazzarKarim, line 1068: Just a learning question, why is this line repeated 6 times?

JazzarKarim (approved):

legal-api/migrations/versions/004f7558063f_second.py
argush3: It is just a way of dropping the 6 FK constraints created for `alternate_names` table.  I believe you can delete the FK by the FK name explicitly as well.

JazzarKarim: Noted, thanks!

PR #2481, opened 2024-02-28
19865-disallow-ting-restorations

jamespaologarcia:
*Issue #:* /bcgov/entity#19865

Disallow restorations for businesses in pending amalgamation

### DIFF
--- a/legal-api/src/legal_api/services/authz.py
+++ b/legal-api/src/legal_api/services/authz.py
@@ -46,7 +46,7 @@ class BusinessBlocker(str, Enum):
     DEFAULT = 'DEFAULT'
     BUSINESS_FROZEN = 'BUSINESS_FROZEN'
     DRAFT_PENDING = 'DRAFT_PENDING'
-    NOT_IN_GOOD_STANDING = 'NOT_IN_GOOD_STANDING',
+    NOT_IN_GOOD_STANDING = 'NOT_IN_GOOD_STANDING'
     AMALGAMATING_BUSINESS = 'AMALGAMATING_BUSINESS'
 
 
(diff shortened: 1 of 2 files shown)


jamespaologarcia:
Currently working on the test cases. Checking to see if I'm on the right track.

legal-api/src/legal_api/services/authz.py
```diff
@@ -615,6 +620,9 @@ def business_blocker_check(business: Business, is_ignore_draft_blockers: bool =
     if not business.good_standing:
         business_blocker_checks[BusinessBlocker.NOT_IN_GOOD_STANDING] = True
 
+    if business.is_pending_amalgamating_business:
```
argush3, line 623: Your tests fail because `is_pending_amalgamating_business` isn't an instance method.

Should be something like `if Business.is_pending_amalgamating_business(business.identifier):`

jamespaologarcia: Thank you. This fixed it. Working on adding unit tests now

vysakh-menon-aot (approved):

argush3:
Can we fix the broken tests?

legal-api/tests/unit/services/test_authorization.py
```diff
@@ -1151,6 +1151,58 @@ def mock_auth(one, two):  # pylint: disable=unused-argument; mocks of library me
...
+    with app.test_request_context():
+        monkeypatch.setattr('flask.request.headers.get', mock_auth)
+
+        for legal_type in legal_types:
+            business = None
+            identifier = (f'BC{random.SystemRandom().getrandbits(0x58)}')[:9]
+            business = factory_business(identifier=identifier,
+                                        entity_type=legal_type,
+                                        state=state)
+            with patch.object(type(business), 'is_pending_amalgamating_business', new_callable=PropertyMock) as is_ting:
```
argush3, line 1202: does this test work? 

from what i can see, you aren't using `is_pending_amalgamating_business` anymore for your allowable actions check.

vysakh-menon-aot:
> Can we fix the broken tests?

It doesn't look like related to this pr. It would be good if you can fix

legal-api/tests/unit/services/test_authorization.py
jamespaologarcia: Updated this to mock the amalgamating_businesses instead

argush3 (approved):

PR #2493, opened 2024-03-05
19827 Fix broken unit tests (validation services)

leodube-aot:
*Issue #:* /bcgov/entity#19827

*Description of changes:*
- Unit test outputs:
  - `services/business`: 6 passed
  - `services/filings`: 585 passed

### DIFF
--- a/legal-api/src/legal_api/models/legal_entity.py
+++ b/legal-api/src/legal_api/models/legal_entity.py
@@ -566,7 +566,7 @@ class LegalEntity(
     def find_by_identifier(cls, identifier: str = None, skip_identifier_validation: bool = False):
         """Return a Business by the id assigned by the Registrar."""
         if not identifier or (
-            skip_identifier_validation and not cls.validate_identifier(entity_type=None, identifier=identifier)
+            not skip_identifier_validation and not cls.validate_identifier(entity_type=None, identifier=identifier)
         ):
             return None
 
--- a/legal-api/src/legal_api/models/legal_entity.py
+++ b/legal-api/src/legal_api/models/legal_entity.py
@@ -565,8 +565,7 @@ class LegalEntity(
     @classmethod
     def find_by_identifier(cls, identifier: str = None):
         """Return a Business by the id assigned by the Registrar."""
-        if not identifier or not cls.validate_identifier(entity_type=None, identifier=identifier
-        ):
+        if not identifier or not cls.validate_identifier(entity_type=None, identifier=identifier):
             return None
 
         non_business_types = [
--- a/legal-api/src/legal_api/models/legal_entity.py
+++ b/legal-api/src/legal_api/models/legal_entity.py
@@ -563,10 +563,9 @@ class LegalEntity(
         return None
 
     @classmethod
-    def find_by_identifier(cls, identifier: str = None, skip_identifier_validation: bool = False):
+    def find_by_identifier(cls, identifier: str = None):
         """Return a Business by the id assigned by the Registrar."""
-        if not identifier or (
-            not skip_identifier_validation and not cls.validate_identifier(entity_type=None, identifier=identifier)
+        if not identifier or not cls.validate_identifier(entity_type=None, identifier=identifier
         ):
             return None
 
--- a/legal-api/src/legal_api/models/legal_entity.py
+++ b/legal-api/src/legal_api/models/legal_entity.py
@@ -563,9 +563,11 @@ class LegalEntity(
         return None
 
     @classmethod
-    def find_by_identifier(cls, identifier: str = None):
+    def find_by_identifier(cls, identifier: str = None, skip_identifier_validation: bool = False):
         """Return a Business by the id assigned by the Registrar."""
-        if not identifier or not cls.validate_identifier(entity_type=None, identifier=identifier):
+        if not identifier or (
+            skip_identifier_validation and not cls.validate_identifier(entity_type=None, identifier=identifier)
+        ):
             return None
 
         non_business_types = [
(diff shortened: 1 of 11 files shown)


legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
```diff
@@ -86,7 +86,15 @@ def validate_amalgamating_businesses(  # pylint: disable=too-many-branches,too-m
...
     for amalgamating_business_json in amalgamating_businesses_json:
         if identifier := amalgamating_business_json.get("identifier"):
-            if not (business := BusinessService.fetch_business(identifier)):
+            if (
+                identifier.startswith("A")
+                and (foreign_jurisdiction := amalgamating_business_json.get("foreignJurisdiction"))
+                and foreign_jurisdiction.get("country") == "CA"
+                and foreign_jurisdiction.get("region") == "BC"
+            ):
+                is_any_expro_a = True
```
leodube-aot, line 95: We no longer use `corpNumber`, so this section of code had to moved out of the `elif` statement below.

legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
```diff
@@ -132,8 +134,7 @@ def validate_amalgamating_businesses(  # pylint: disable=too-many-branches,too-m
                 if not _is_business_affliated(identifier, account_id):
                     msg.append(
                         {
-                            "error": f"{identifier} is not affiliated \
-                                with the currently selected BC Registries account.",
+                            "error": f"{identifier} is not affiliated with the currently selected BC Registries account.",  # noqa: E501
```
leodube-aot, line 137: The tab characters get added into the string when broken up like this.

legal-api/src/legal_api/models/legal_entity.py
```diff
@@ -563,9 +563,11 @@ def find_by_operating_name(cls, operating_name: str = None):
         return None
 
     @classmethod
-    def find_by_identifier(cls, identifier: str = None):
+    def find_by_identifier(cls, identifier: str = None, skip_identifier_validation: bool = False):
```
leodube-aot, line 566: Added this `skip_identifier_validation` flag to be used in amalgamations when fetching an expro business as the business identifier for expros does not pass the `validate_identifier` check.

legal-api/tests/unit/services/filings/validations/test_amalgamation_application.py
```diff
@@ -1620,6 +1616,7 @@ def mock_find_by_identifier(identifier):
         "legal_api.services.bootstrap.AccountService.get_account_by_affiliated_identifier",
         return_value={"orgs": [{"id": account_id}]} if test_status == "SUCCESS" else {},
     )
+    mocker.patch("legal_api.models.legal_entity.LegalEntity.validate_identifier", return_value=True)
```
leodube-aot, line 1619: We need to patch `validate_identifier` here as it is called when the identifier is set for a LegalEntity. In these tests we use `mock_find_by_identifier`, which creates a new LegalEntity with potentially an expro identifier (which will fail the `validate_identifier` check).

legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
```diff
@@ -86,7 +86,15 @@ def validate_amalgamating_businesses(  # pylint: disable=too-many-branches,too-m
...
-            if not (business := BusinessService.fetch_business(identifier)):
+            if (
+                identifier.startswith("A")
+                and (foreign_jurisdiction := amalgamating_business_json.get("foreignJurisdiction"))
+                and foreign_jurisdiction.get("country") == "CA"
+                and foreign_jurisdiction.get("region") == "BC"
+            ):
+                is_any_expro_a = True
+
+            if not (business := BusinessService.fetch_business(identifier, skip_identifier_validation=True)):
```
vysakh-menon-aot, line 97: these validations are updated recently in main branch.  this may not be needed

legal-api/src/legal_api/models/legal_entity.py
vysakh-menon-aot: the name says skip identifier validation but the below implementation is opposite, it validates if skip_identifier_validation is True

vysakh-menon-aot: @argush3 do we need to validate identifier in this function?

argush3: Maybe not.  

@thorwolpert do you have any opinions on this?  I think you added this piece of code.  

We already validate the identifier when we set the identifier property on LE.  

https://github.com/bcgov/lear/blob/95579b862cf27093747a9a4df7432b299c5b03fe/legal-api/src/legal_api/models/legal_entity.py#L245-L251

Or were you wanting to block invalid searches so it never hits the db?

argush3: I think we can just pull in the code from `main` to get things working and just remove `skip_identifier_validation` for now.

@JazzarKarim is working on syncing `main` with `feature-legal-name` for the BE right now so anything missing pieces should be pulled in soon too.

leodube-aot: Okay sounds good, yeah the main branch changes means we no longer need `skip_identifier_validation`. I can remove the flag. 

Should I copy the relevant main branch changes over in this PR or just wait for the rebase? The tests will pass without the main branch changes.

(Not sure what happened to my previous GH comment on this thread, but I was able to confirm the main branch changes work without this flag through testing).

argush3 (approved):

PR #2509, opened 2024-03-13
20121 Fix db data not being reset properly between test runs

leodube-aot:
*Issue #:* /bcgov/entity#20121

*Description of changes:*
- Update conftest to work with flask-sqlalchemy 3.x
- A couple small fixes to some of the unit tests
- This update improves the unit test success:
```
Latest test run on feature-legal-name == 382 failed, 1466 passed, 44 skipped in 303.10s (0:05:03)
Current test run in this PR: =========== 182 failed, 1666 passed, 44 skipped in 308.61s (0:05:08)
```
- There's still failing unit tests but from a cursory look the failing tests are due to different issues unrelated to improper transaction rollbacks. 

### What caused the bug
In the legal name feature branch, we upgraded the `flask-sqlalchemy` package from 2.5.1 to ^3.0.3. Part of this upgrade means moving from SqlAlchemy 1.x to SqlAlchemy 2.x (confirmed in [poetry.lock](https://github.com/bcgov/lear/blob/feature-legal-name/legal-api/poetry.lock#L799)). This seems to have [caused issues with rollback in pytest](https://github.com/pallets-eco/flask-sqlalchemy/discussions/1179). 

I think what is happening is that when we commit to the nested session, the SAVEPOINT is not getting restarted correctly. We end up with pending changes getting commited to the outermost transaction, and are not getting rolled back within the scope of the savepoint. This can cause queries that expect `one_or_none()` to fail, like in [`legal_entity.find_by_identifier`](https://github.com/bcgov/lear/blob/feature-legal-name/legal-api/src/legal_api/models/legal_entity.py#L575-L579) since there are multiple legal entities created within the outermost transaction with the same identifier.

### Proposed solution
The SQLAlchemy 2.x docs outlines a new approach to [joining a session into an external transaction](https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#session-external-transaction). This uses the new [Session.join_transaction_mode](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.params.join_transaction_mode) parameter with the setting "create_savepoint", which "_indicates that new SAVEPOINTs should be created in order to implement BEGIN/COMMIT/ROLLBACK for the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session), which will leave the external transaction in the same state in which it was passed._"

### DIFF
--- a/legal-api/tests/conftest.py
+++ b/legal-api/tests/conftest.py
@@ -25,6 +25,8 @@ from flask_migrate import Migrate, upgrade
 from ldclient.integrations.test_data import TestData
 from sqlalchemy import event, inspect, text
 from sqlalchemy.schema import DropConstraint, MetaData
+from sqlalchemy.orm import scoped_session, sessionmaker
+
 
 from legal_api import create_app
 from legal_api import jwt as _jwt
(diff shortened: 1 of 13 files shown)


legal-api/tests/conftest.py
```diff
@@ -267,25 +268,16 @@ def session(app, db):  # pylint: disable=redefined-outer-name, invalid-name
         txn = conn.begin()
 
         try:
-            options = dict(bind=conn, binds={})
-            # sess = db.create_scoped_session(options=options)
-            sess = db._make_scoped_session(options=options)
+            sess = scoped_session(
```
seeker25, line 271: Looks good, much better than what I wrote lol

seeker25: https://github.com/bcgov/sbc-pay/blob/06673215d07ac896fdc84e134f424de4d89c87b5/pay-api/tests/conftest.py#L84

I'll give it a go in pay

argush3 (approved):
Looks good!  

And thanks for the detailed explanation!

PR #2517, opened 2024-03-14
20276 Fixed change of registration PUT request

JazzarKarim:
*Issue #:* /bcgov/entity#20276

*Description of changes:*
- Fixed the doing a PUT request for the change of registration.
- This should also hopefully fix all PUT requests for filings done for entries in the Alternate Names.

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -615,7 +615,9 @@ class ListFilingResource:
                 )
                 rv = (
                     db.session.query(business_model, Filing)
-                    .join(Filing, filing_entity_id == business_model.id)  # join Filing table based on the dynamic column
+                    .join(
+                        Filing, filing_entity_id == business_model.id
+                    )  # join Filing table based on the dynamic column
                     .filter(business_model.identifier == business_identifier)
                     .filter(Filing.id == filing_id)
                     .one_or_none()
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -615,7 +615,7 @@ class ListFilingResource:
                 )
                 rv = (
                     db.session.query(business_model, Filing)
-                    .filter(business_model.id == filing_entity_id)
+                    .join(Filing, filing_entity_id == business_model.id)  # join Filing table based on the dynamic column
                     .filter(business_model.identifier == business_identifier)
                     .filter(Filing.id == filing_id)
                     .one_or_none()
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -547,7 +547,6 @@ class ListFilingResource:
             current_app.logger.error("Business:%s unable to post to queue, err=%s", identifier, err)
             return {"errors": {"message": "unable to publish for post processing"}}, HTTPStatus.BAD_REQUEST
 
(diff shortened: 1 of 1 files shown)


JazzarKarim:
Here's an example/verification:

For the POST Request:
post prev

Now for the PUT Request, previously we had:
put prev

Now, we're getting:
put now

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -596,13 +606,22 @@ def save_filing(  # pylint: disable=too-many-return-statements,too-many-branches
...
+                        .one_or_none()
+                    )
+                else:
+                    rv = (
+                        db.session.query(LegalEntity, Filing)
+                        .filter(LegalEntity.id == Filing.legal_entity_id)
+                        .filter(LegalEntity.identifier == business_identifier)
+                        .filter(Filing.id == filing_id)
+                        .one_or_none()
+                    )
```
argush3, line 625: Think we can simplify and make it one query.

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -537,6 +547,7 @@ def process_colin_filing(identifier: str, filing: Filing, business: any) -> Tupl
             current_app.logger.error("Business:%s unable to post to queue, err=%s", identifier, err)
             return {"errors": {"message": "unable to publish for post processing"}}, HTTPStatus.BAD_REQUEST
 
+    # pylint: disable=too-many-statements
```
JazzarKarim, line 550: `pylint` run was failing. This is to suppress that.

JazzarKarim: Reverted this change. The two extra if statements are no longer there so pylint is fine now (49/50 statements).

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
JazzarKarim: Updated, and verified!
202 put

Let me know if you're OK with the variable names. My brain went blank on what to name them 😆

leodube-aot (approved):
LGTM!

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -596,10 +606,17 @@ def save_filing(  # pylint: disable=too-many-return-statements,too-many-branches
...
+                filing_entity_id = (
+                    Filing.alternate_name_id if business.is_alternate_name_entity else Filing.legal_entity_id
+                )
                 rv = (
-                    db.session.query(LegalEntity, Filing)
-                    .filter(LegalEntity.id == Filing.legal_entity_id)
-                    .filter(LegalEntity.identifier == business_identifier)
+                    db.session.query(business_model, Filing)
+                    .filter(business_model.id == filing_entity_id)
+                    .filter(business_model.identifier == business_identifier)
```
argush3, line 619: think there is a problem with `.filter(business_model.id == filing_entity_id)`.  These are both columns and the filter requires a column to be compared against a value I believe.

Maybe we need something like

```python
rv = (
        db.session.query(business_model, Filing)
        .join(Filing, filing_entity_id == business_model.id)  # join Filing table based on the dynamic column
        .filter(business_model.identifier == business_identifier)
        .filter(Filing.id == filing_id)
        .one_or_none()
    )
```

JazzarKarim: Argus, I've made this change and verified:
For Firms:
202 put worksafter comment

For a BEN, I tried doing a PUT by making an alteration:
alteration PUT

argush3 (approved):

PR #2533, opened 2024-03-20
20320 Get entity-filer tests running

leodube-aot:
*Issue #:* /bcgov/entity#20320

*Description of changes:*
- Some setup work for the upcoming filer updates. This PR updates the common models and gets the unit tests running (but does not necessarily fix any).
- Updated the common models to match legal api models. Also updated the common model migrations to match legal api.
- Fixed some conftest and config issues that were affecting the test db setup and connection.
- There is currently a dependency issue with the pyproject.toml file that should get resolved when this PR gets merged (since common models now uses registry-schemas to 2.18.19 but it is still 2.18.10 on the feature branch).

Note: The migration part of the db setup in conftest takes very long on each run (a couple minutes), which does not seem to happen for the legal api tests. I'm wondering why that might be the case?


chenhongjing (approved):
LGTM 😸

argush3 (approved):
Thanks!  This will be super helpful for the other devs doing work in the filer too.

Interesting note about the slow tests.  Wonder if it's minio related or something.  Anyways, something we may need to look into.

kzdev420 (approved):

PR #2538, opened 2024-03-22
20360 Update reg and change of reg outputs with alt name

kzdev420:
*Issue #:* /bcgov/entity#20360

*Description of changes:*

Screenshot 2024-03-22 at 11 26 24 AM

Screenshot 2024-03-22 at 11 26 49 AM


argush3:
So have you tested registration and change of registration outputs against GPs and the different flavors of SP to see if they load? Can you give me a summary of scenarios you've tested?

Also, for change of registration, did you verify the amended registration statement output?
image

argush3:
You'll need to rebase against legal name branch as bunch of new changes have been merged.

legal-api/src/legal_api/models/alternate_name.py
```diff
@@ -281,6 +281,11 @@ def alias_json(self):
         alias = {"id": str(self.id), "name": self.name, "type": self.name_type.name}
         return alias
 
+    @property
+    def founding_date(self):
+        """Return the founding_date."""
+        return self.start_date
+
```
argush3, line 288: Don't think we want to do this.  `start_date` can be used for name translations and other future name types which isn't a founding date.

legal-api/src/legal_api/reports/report.py
```diff
@@ -295,6 +295,10 @@ def _handle_special_resolution_filing_data(self, filing):
             self._format_special_resolution_application(filing, "alteration")
 
     def _set_completing_party(self, filing):
+        business = BusinessService.fetch_business_by_filing(self._filing)
+        if business.is_alternate_name_entity and (business.entity_type == LegalEntity.EntityTypes.SOLE_PROP.value):
+            return
+
```
argush3, line 301: I believe a SP still has a completing party

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -538,13 +538,14 @@ def party_role_revision_json(filing, party_role_revision, is_ia_or_after) -> dic
     @staticmethod
     def get_party_revision(filing, party_id) -> dict:
```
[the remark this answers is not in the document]
argush3: so which column doesn't it like specifically?  the `changed` column?

If that's the case, can you dynamically add the columns and exclude the `changed` column for whichever business model(LE or AN table) get matched?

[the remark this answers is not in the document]
kzdev420: Yeah. history table has the `changed` column more.
I will add the columns dynamically for LE and AN

legal-api/src/legal_api/reports/report.py
kzdev420: I think we return the party info for the only `proprietor` role.

https://github.com/kzdev420/lear/blob/1d171cb28fc5044cbe40014ea89a3e8eb20a4266/legal-api/src/legal_api/models/alternate_name.py#L206-L219

legal-api/src/legal_api/models/alternate_name.py
kzdev420: Then, which value should I use as founding_date for Alt_names
I referred to this code, line 297

https://github.com/kzdev420/lear/blob/1d171cb28fc5044cbe40014ea89a3e8eb20a4266/legal-api/src/legal_api/models/alternate_name.py#L284-L306

legal-api/src/legal_api/reports/report.py
argush3: No, this is specifically for the completing party.  The proprietor party info is different.  Also, the proprietor is not necessarily the same as the completing party if that's what you were thinking.

I think it may work as is if you just remove the new lines of code you added.  

image

legal-api/src/legal_api/models/alternate_name.py
argush3: You can just use `start_date` directly.  

If you want to really add the `founding_date` property, then add it to `BusinessCommon`.  But you will need to update the LE model's `founding_date` property to be `_founding_date` so we an set the value.

legal-api/src/legal_api/reports/report.py
kzdev420: Okay

legal-api/src/legal_api/models/alternate_name.py
kzdev420: Got it

kzdev420: I used the `start_date` directly

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -583,15 +584,17 @@ def party_role_revision_json(filing, party_role_revision, is_ia_or_after) -> dic
...
+            db.session.query(business_attr)
+            .filter(business_attr.change_filing_id == filing.id)
+            .filter(business_attr.id == party_id)
         )
 
-        party_version = history_cls(LegalEntity)
+        party_version = history_cls(business_attr)
         party_history = (
             db.session.query(party_version)
+            .options(defer(party_version.changed))  # TODO: remove this after update history_cls
```
argush3, line 597: I don't think we need this todo comment?  how were you planning on changing the `history_cls`?

Also, have you confirm this actually works?  i.e. are history records properly included in the union?

kzdev420: I thought the `changed` column could be removed in the future because it is optional. but, I was wrong.
I will remove the comments.

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -812,6 +815,9 @@ def resolution_json(resolution_revision) -> dict:
     @staticmethod
     def business_revision_json(business_revision, business_json):
         """Return the business revision as a json object."""
+        if not business_revision:
+            return business_json
```
argush3, line 819: This is not required.  I believe you should also have a non-null value for `business_revision`.

You should be updating this [function](https://github.com/bcgov/lear/blob/1b40e6d07ba8b1d090b4ffd21ca3aaef102f1533/legal-api/src/legal_api/services/business_details_version.py#L222-L238) as it's incorrect.

legal-api/src/legal_api/models/alternate_name.py
argush3: where are you using `start_date` directly?  can you link me?

legal-api/src/legal_api/reports/report.py
```diff
@@ -869,19 +869,18 @@ def _format_change_of_registration_data(
...
-            )
-            filing["offices"]["businessOffice"]["deliveryAddress"]["changed"] = self._compare_address(
-                business_office.get("deliveryAddress"), offices_json["businessOffice"]["deliveryAddress"]
-            )
-            filing["offices"]["businessOffice"]["changed"] = (
-                filing["offices"]["businessOffice"]["mailingAddress"]["changed"]
-                or filing["offices"]["businessOffice"]["deliveryAddress"]["changed"]
-            )
+            offices_json = VersionedBusinessDetailsService.get_office_revision(prev_completed_filing, self._business)
+            if offices_json:
```
argush3, line 873: this is the same situation as https://github.com/bcgov/lear/blob/1b40e6d07ba8b1d090b4ffd21ca3aaef102f1533/legal-api/src/legal_api/services/business_details_version.py#L216-L219 I believe.

legal-api/src/legal_api/models/alternate_name.py
kzdev420: https://github.com/bcgov/lear/blob/1b40e6d07ba8b1d090b4ffd21ca3aaef102f1533/legal-api/src/legal_api/reports/report.py#L466-L468

argush3:
so I had already asked above but never got my [comment](https://github.com/bcgov/lear/pull/2538#issuecomment-2015853942) addressed.

I need the following from you:
* A list of OCP vs GCP outputs for the scenarios you tested
* Did you verify the amended registration statement output?
* The registration output you attached in the description has no value for "business name".  Does your most recent code provide that now?  If so, please provided updated OCP vs GCP outputs

legal-api/src/legal_api/reports/report.py
```diff
@@ -315,7 +315,7 @@ def _set_completing_party(self, filing):
             self._filing.id, datetime.utcnow(), EntityRole.RoleTypes.completing_party.name
         )
         if completing_party_role:
-            filing["completingParty"] = completing_party_role[0].party.json
+            filing["completingParty"] = completing_party_role[0].json
```
argush3, line 318: Are we sure about this?  The original code went through `party` which is now a `legal_entity` record.  

Can we go through the `related_entity_id` and call `party_json` on the legal entity model?  

image

legal-api/src/legal_api/services/business_details_version.py
argush3: Ok, but otherwise, you've confirmed using `defer` still brings back history records properly?

kzdev420: Yes, I confirmed

JazzarKarim (commented):
From what I can see, LGTM as long as Argus' comments get resolved 👍

legal-api/src/legal_api/reports/report.py
kzdev420: Updated

legal-api/src/legal_api/reports/report.py
kzdev420: Updated

legal-api/src/legal_api/services/business_details_version.py
kzdev420: I tested with `FM1064603`.
But, I cannot see any revision for `FM1064603` in the `Alternate_names_history` table
Screenshot 2024-04-10 at 3 42 37 AM

legal-api/src/legal_api/reports/report.py
```diff
@@ -903,6 +901,8 @@ def _format_change_of_registration_data(
                     prev_party_json = VersionedBusinessDetailsService.party_revision_json(
                         prev_completed_filing, prev_party, True
                     )
+                    # current_party type is person, but prev_party type is organization
+                    # so has issue at _has_party_name_change()
                     if self._has_party_name_change(prev_party_json, party):
```
kzdev420, line 906: `FM1064603`, filing_id: `148571`:
In this case, prev_party is organization. current_party is person.
So `_has_party_name_change` function returns the error. We should fix this logic

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -308,33 +311,36 @@ def get_office_revision(filing, business) -> dict:  # pylint: disable=too-many-l
...
-                db.session.query(address_history)
+                db.session.query(*add_history_columns)
                 .filter(address_history.change_filing_id <= filing_id)
                 .filter(address_history.office_id == office.id)
             )
 
-            addresses_result = addresses_current.union(addresses_historical).all()
-            for address_result in addresses_result:
-                address, _ = address_result
+            addresses = addresses_current.union(addresses_historical).all()
```
kzdev420, line 342: `FM1000406`, filing_id: `146925`
In this case, business_id is 691 ( alternate_name_id )

But, there are no records for this business in Address and Address_history table ( my local DB )
So, it returns `None`, and will make the error

legal-api/src/legal_api/reports/report.py
argush3: We can fix this in another ticket.  We need to think about this more when we get back to this work.

I'll create a new ticket.

Can you find a business to test with that doesn't have party type changing?

kzdev420: I am trying to find it

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -583,17 +589,32 @@ def party_role_revision_json(filing, party_role_revision, is_ia_or_after) -> dic
...
-            db.session.query(LegalEntity)
-            .filter(LegalEntity.change_filing_id == filing.id)
-            .filter(LegalEntity.id == party_id)
+            db.session.query(business_attr)
+            .filter(business_attr.change_filing_id == filing.id)
+            .filter(business_attr_id == int(party_id))
         )
 
-        party_version = history_cls(LegalEntity)
+        columns_to_select = VersionedBusinessDetailsService.select_revision_columns(party_version)
```
argush3, line 613: I'm not sure if you are selecting the right columns for the party in all cases.  For example, it looks like if it's a SP person, SP DBA LEAR or SP DBA COLIN, you are selecting the alternate name columns.  I think you should be selecting the columns for the LE or AN record.

legal-api/src/legal_api/services/business_details_version.py
argush3: so whic output are you testing?

kzdev420: Sorry, it should be removed. I forgot it

argush3:
Just tested your PR against FM1064603 for the statement of registration output.  It's still missing the business name.

image

kzdev420:
> Just tested your PR against FM1064603 for the statement of registration output. It's still missing the business name. I've mentioned this a few times now...
> 
> image

Sorry, I forgot to push the changes for this. Just pushed

legal-api/src/legal_api/reports/report.py
argush3: Ignore my previous comment.  This is a code issue, not a data issue I think.  

You are pulling back an AN record when the code calls `get_party_revision`.

image

It should be pulling back the LE person history record.

image

I'm assuming this is related to your `_has_party_name_change` error.

legal-api/src/legal_api/services/business_details_version.py
argush3: Can you make this change.  I think it will fix your problem with retrieving a business_revision

image

kzdev420: Yes, I will do

legal-api/src/legal_api/reports/report.py
kzdev420: Okay, if business.legal_entity_id is null and colin_entity_id is True, then we should pulling back the Colin_entity history?

argush3: yes, but i'm not entirely sure if that will work as is.  you may need to do some work to get it working.  maybe put a todo and ignore that scenario for now.  just say we need to still take care of that scenario in the future.

legal-api/src/legal_api/services/business_details_version.py
```diff
@@ -572,7 +572,7 @@ def party_role_revision_json(filing, party_role_revision, is_ia_or_after) -> dic
     def get_party_revision(filing, party_id) -> dict:
         """Consolidates all party changes up to the given transaction id."""
         business = BusinessService.fetch_business_by_filing(filing)
-        if filing.legal_entity_id:
+        if filing.legal_entity_id or business.legal_entity_id:
```
argush3, line 578: i don't think this change is required?

kzdev420: FM1064603, filing_id: 148571.

In this case, the filing doesn't have the legal_entity. it has the alternate_name_entity.
And alternate_name_entity has the legal_entity.

I think you mentioned that filing pull back the legal_entity in this case

argush3 (approved):
Approving.  As per our conversation, outstanding issues that still need to be addressed will be picked up in #20903

legal-api/src/legal_api/services/business_details_version.py
argush3: Yes, I mentioned that.  But I think the logic needs to be more precise.  Anyways, we can pick this up in the new ticket.

PR #2543, opened 2024-03-25
20327 Filer - Update registration filing processor to work with alternate name changes

chenhongjing:
*Issue #:* /bcgov/entity#20327

*Description of changes:*
- Update work.py to work with alternate names
- Update registration processor to work with alternate names
- Update unit tests
- Import busines_service.py directly from legal-api for now (any better way?)
- Misc updates


queue_services/entity-filer/src/entity_filer/filing_processors/registration.py
```diff
@@ -116,38 +112,43 @@ def process(
...
+    alternate_name.state = BusinessCommon.State.ACTIVE
 
     if nr_number := business_info_obj.get("nrNumber", None):
         # TODO check how this is getting used, may need operating name?
         filing_meta.registration = {
             **filing_meta.registration,
             **{
                 "nrNumber": nr_number,
-                "legalName": business_info_obj.get("legalName", None),
+                "operatingName": business_info_obj.get("legalName", None),
```
chenhongjing, line 144: Do we need to change `legalName` to `operatingName` here? Or just keep it?

queue_services/entity-filer/src/entity_filer/resources/worker.py
```diff
@@ -80,6 +80,7 @@
 )
 from entity_filer.filing_processors.filing_components import name_request
 from entity_filer.services import queue
+from entity_filer.services.business_service import BusinessService
```
leodube-aot, line 83: I think for now this is the correct approach (at least it's the one I also decided to take), which is to copy the business service instead of importing from legal api.

queue_services/entity-filer/src/entity_filer/resources/worker.py
```diff
@@ -432,7 +439,11 @@ def process_filing(
         #     )
 
     if any("registration" in x for x in legal_filings):
-        filing_submission.legal_entity_id = business.id
+        if alternate_name.entity_type == BusinessCommon.EntityTypes.SOLE_PROP:
```
leodube-aot, line 442: Is it possible for `alternate_name` to be `None` here? If so you may need to also check it exists before getting the entity type.

queue_services/entity-filer/src/entity_filer/filing_processors/registration.py
leodube-aot: From what I can tell, the filing meta data is mainly used by the reports in legal api. I couldn't actually find anywhere where `legalName` is used from the filing meta (except maybe [here for continuation out](https://github.com/bcgov/lear/blob/feature-legal-name/legal-api/src/legal_api/reports/business_document.py#L447)). I think it's okay to leave as `legalName`, especially since this is in reference to the name request,  which [still uses legalName](https://github.com/bcgov/business-schemas/blob/feature-legal-name/src/registry_schemas/schemas/name_request.json#L19).

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/legal_entity_info.py
```diff
@@ -38,7 +38,7 @@ def set_corp_type(legal_entity: LegalEntity, legal_entity_info: Dict) -> Dict:
     try:
         entity_type = legal_entity_info.get("legalType")
         if entity_type:
-            legal_entity.entity_type = entity_type
+            legal_entity._entity_type = entity_type
```
argush3, line 41: this has actually reverted to `entity_type` now I believe.  Might need to pull in the latest LE model.

queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/legal_entity_info.py
```diff
@@ -63,19 +63,19 @@ def set_association_type(legal_entity: LegalEntity, association_type: String) ->
...
+        legal_entity._legal_name = f"{corp_num[2:]} {numbered_legal_name_suffix}"
 
 
 def update_legal_entity_info(corp_num: str, legal_entity: LegalEntity, legal_entity_info: Dict, filing: Filing):
     """Format and update the legal_entity entity from incorporation filing."""
     if corp_num and legal_entity and legal_entity_info and filing:
         set_legal_name(corp_num, legal_entity, legal_entity_info)
         legal_entity.identifier = corp_num
-        legal_entity.entity_type = legal_entity_info.get("legalType", None)
+        legal_entity._entity_type = legal_entity_info.get("legalType", None)
```
argush3, line 78: Can be reverted back to `entity_type`

queue_services/entity-filer/src/entity_filer/filing_processors/registration.py
```diff
@@ -116,38 +112,43 @@ def process(
...
         case _:
             # Default and failed
             # Based on the above checks, this should never happen
             raise DefaultException(f"registration {filing_rec.id} had no valid Firm type.")
 
     # Assuming we should not reset this from a filing
-    if not business.tax_id:
-        business.tax_id = registration_filing.get("business", {}).get("taxId", None)
+    # if not business.tax_id:
+    #     business.tax_id = registration_filing.get("business", {}).get("taxId", None)
```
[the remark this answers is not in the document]
argush3: I think we might be able to delete these 2 lines altogether.

I believe the `legal_entity.tax_id` or `alternate_name.bn15` only ever get set via entity-bn listener service or update-legal-filings job.

Will double check...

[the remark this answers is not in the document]
argush3: Actually not sure...seems like it was added intentionally.  Asking Vysakh right now...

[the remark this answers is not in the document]
argush3: Talked to Vysakh...conversation below.

Basically, I think we should do the following:

**SP person**
- if registration filing json has tax_id populated in business block, assign tax_id(bn9) value to `legal_entities.tax_id` and `alternate_names.bn15`

**SP DBA LEAR/COLIN**
- if registration filing json has tax_id populated in business block, add tax_id(bn9)  alternate_name record

**GP**
- if registration filing json has tax_id populated in business block, assign tax_id(bn9) value to `legal_entities.tax_id`

image

queue_services/entity-filer/src/entity_filer/filing_processors/registration.py
```diff
@@ -116,25 +112,24 @@ def process(
...
             # Default and failed
             # Based on the above checks, this should never happen
             raise DefaultException(f"registration {filing_rec.id} had no valid Firm type.")
 
-    # Assuming we should not reset this from a filing
-    if not business.tax_id:
-        business.tax_id = registration_filing.get("business", {}).get("taxId", None)
-
-    business.state = LegalEntity.State.ACTIVE
+    business.state = BusinessCommon.State.ACTIVE
```
argush3, line 131: so `business` here can potentially be colin entity or a LE person record right?  

Don't think we need to set the state for such scenarios.  Only the alternate name record will need the state set.  

And in the case of colin entity, I think it will error out.

chenhongjing: Yes it could be. Thanks for catching it! 😸 So only update SP & GP themselves right?

chenhongjing: Will remove this line and the line below since it's set `ACTIVE` by default.

argush3 (approved):

PR #2565, opened 2024-04-03
20148 remove operating name

JazzarKarim:
*Issue #:* /bcgov/entity#20148

*Description of changes:*
- Removed `operatingName` from everywhere because we're using `name` now.

Verification/Testing:
removed operatingname

### DIFF
--- a/legal-api/src/legal_api/models/alternate_name.py
+++ b/legal-api/src/legal_api/models/alternate_name.py
@@ -333,7 +333,7 @@ class AlternateName(Versioned, db.Model, BusinessCommon):
                         if self.business_start_date
                         else None
                     ),
-                    "nameType": self.name_type.name
+                    "nameType": self.name_type.name,
                 }
             ],
         }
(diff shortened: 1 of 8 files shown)


legal-api/src/legal_api/models/alternate_name.py
```diff
@@ -333,8 +333,7 @@ def _slim_json(self):
                         if self.business_start_date
                         else None
                     ),
-                    "nameType": self.name_type.name,
-                    "operatingName": self.name,  # will be removed in the future
+                    "nameType": self.name_type.name
```
chenhongjing, line 336: Just a comment not a big issue, this may trigger linting failure without comma, but why does it not have any CI check?

chenhongjing (approved):

legal-api/src/legal_api/models/alternate_name.py
JazzarKarim: Good catch! I might be because when I created my PR, I had main set as the base by mistake. After creating it, I changed the base branch so maybe that prevented the CI from triggering.

argush3 (approved):

PR #2567, opened 2024-04-04
Fix issue where none was being returned for DBAs where owner is GP

argush3:
*Issue #:* /bcgov/entity###

*Description of changes:*
* Updated business_service class functions for returning a business to work for scenario where GP is owner of a SP.  Previously, I believe we had to do some hacky logic when `alternate_name.entity_type` wasn't an actual column in the table.  Don't think we need this anymore.

### DIFF
--- a/legal-api/src/legal_api/services/business_service.py
+++ b/legal-api/src/legal_api/services/business_service.py
@@ -30,17 +30,8 @@ class BusinessService:
             return None
         if legal_entity := LegalEntity.find_by_identifier(identifier):
             return legal_entity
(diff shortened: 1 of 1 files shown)


JazzarKarim (approved):

tshyun24 (approved):
LGTM, just the linting issue

chenhongjing (approved):
Good but linting failed. ~And do you also fix it in common data model in this PR?~

vysakh-menon-aot (approved):

PR #2577, opened 2024-04-10
20334 Fix outstanding legal entity model tests

leodube-aot:
*Issue #:* /bcgov/entity#20334

*Description of changes:*
- Fixed outstanding legal entity model tests
- Updated references to `sqlalchemy_continuum` to work with `sql_versioning` instead
- WIP: one test remains broken: `test_amalgamated_into_business_json`


legal-api/src/legal_api/models/business_common.py
```diff
@@ -187,7 +187,7 @@ def legal_name(self):
         """
         from . import ColinEntity, LegalEntity
 
-        if self.entity_type == self.EntityTypes.SOLE_PROP:
+        if self.is_alternate_name_entity:
```
leodube-aot, line 190: This was causing the below error to occur for GPs:

`AttributeError: 'AlternateName' object has no attribute '_legal_name'. Did you mean: 'legal_name'?`

The reason is that `entity_type` used to [return SP for alternate name entities](https://github.com/bcgov/lear/pull/2535/files#diff-a0b8a49b352e95f68f25b22078477bcbfc2a1d9c34ddc1baab448da949d6dfdcL134), but that functionality was removed when we started storing `entity_type` in the `alternate_names` table.

legal-api/src/legal_api/models/amalgamation.py
```diff
@@ -19,11 +19,9 @@
 
 from enum import auto
 
-from sql_versioning import Versioned
-from sqlalchemy import or_
-from sqlalchemy_continuum import version_class
+from sql_versioning import Versioned, history_cls
```
leodube-aot, line 22: Updated `amalgamating_business` and `amalgamation` to use `sql_versioning` instead of `sqlalchemy_continuum`.

legal-api/src/legal_api/models/amalgamation.py
```diff
@@ -82,33 +79,26 @@ def json(self):
         }
 
     @classmethod
-    def get_amalgamation_revision_obj(cls, transaction_id, business_id):
+    def get_amalgamation_revision_obj(cls, filing, legal_entity_id):
         """Get amalgamation for the given transaction id."""
         # pylint: disable=singleton-comparison;
-        amalgamation_version = version_class(Amalgamation)
+        amalgamation_version = history_cls(Amalgamation)
```
leodube-aot, line 85: I'm stuck on fixing one last test and I am not sure why it is failing. The test failing is:

`tests/unit/models/test_legal_entity.py::test_amalgamated_into_business_json[EXIST-2] - AttributeError: 'NoneType' object has no attribute 'amalgamation_date'`

I can trace it back to this function, `get_amalgamation_revision_obj`, which returns `None` from the below query. From what I can tell, the `amalgamations_history` table does not get populated during this test (checking the count confirms this), so the query returns None.

Does anyone know how the history table works, and why it isn't updating for the `test_amalgamated_into_business_json` unit test? I'm sure I am missing something in unit test itself.

kzdev420 (approved):
LGTM

legal-api/src/legal_api/models/amalgamation.py
argush3: I think it's because the amalgamation record would not be in the `amalgamation_history` table in your unit test scenario  

The new versioning works in a way where the latest changes only reside in the parent table which is not how the old versioning worked.  The old versioning model always cut a row for latest change even for the versioning table.

In  the failing test scenario, there will only one entry the `amalgamation` table with no row in `amalgamation_history`.

An update update like this should work I think
image

argush3 (commented):
Looks good.  Just the one last fix on the unit test.

legal-api/src/legal_api/models/amalgamation.py
leodube-aot: Ahh I see, that makes sense. That works perfectly thanks for the help Argus!

argush3 (approved):

PR #2608, opened 2024-04-19
Gazette Messaging for Incorporation Application Filings#6870

stevenc987:
*Issue #:* /bcgov/entity### 6870

*Description of changes:*

When a BEN, LTD, CCC or ULC is created, we need to publish in the government electronic newsletter "The Gazette" the name, number and incorporation date. For these 4 entity types, we are moving this functionality from CPRD and into the Modernization realm.


stevenc987:
According to Thor, we need to hold this till at least Oct. So the codes were check-in to feature-legal-name branch in LEAR and we can continue at that time.


image

PR #2611, opened 2024-04-22
20777 - Update logic for legalName field in business response

chenhongjing:
*Issue #:* /bcgov/entity#20777

*Description of changes:*
- Add `business_legal_name` hybrid property in business model to return correct legal name for SP/GPs when `enable-legal-name-fix` feature flag is on. Note that `legal_name` is frequently used in the outputs as business name so I use another name to prevent incorrect use.
- Fix some potential issues in the outputs where `legalName` in `business.json()` is used
- Update unit tests

*Some testing examples when flag is on*
- SP Individual - FM1028594
image

- SP DBA - FM1054864
image

- GP (2 partners) - FM1044962
image

- GP (>2 partners)
image

- Non firm - BC1218875
image

### DIFF
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -23,18 +23,22 @@ from flask import current_app
 from sqlalchemy.exc import OperationalError, ResourceClosedError
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import backref
+from sqlalchemy.sql import func
 
 from legal_api.exceptions import BusinessException
 from legal_api.utils.base import BaseEnum
 from legal_api.utils.datetime import datetime, timezone
 from legal_api.utils.legislation_datetime import LegislationDatetime
 
+from .amalgamation import Amalgamation  # noqa: F401, I001, I003 pylint: disable=unused-import
 from .db import db  # noqa: I001
+from .party import Party
 from .share_class import ShareClass  # noqa: F401,I001,I003 pylint: disable=unused-import
+
 
 from .address import Address  # noqa: F401,I003 pylint: disable=unused-import; needed by the SQLAlchemy relationship
 from .alias import Alias  # noqa: F401 pylint: disable=unused-import; needed by the SQLAlchemy relationship
-from .amalgamation import Amalgamation  # noqa: F401, I001, I003 pylint: disable=unused-import;
+
 from .filing import Filing  # noqa: F401, I003 pylint: disable=unused-import; needed by the SQLAlchemy backref
 from .office import Office  # noqa: F401 pylint: disable=unused-import; needed by the SQLAlchemy relationship
 from .party_role import PartyRole  # noqa: F401 pylint: disable=unused-import; needed by the SQLAlchemy relationship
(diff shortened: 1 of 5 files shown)


legal-api/src/legal_api/reports/business_document.py
```diff
@@ -122,6 +122,7 @@ def _get_template_data(self, get_json=False):
             # get document data
             business_json['reportType'] = self._document_key
             business_json['business'] = self._business.json()
+            business_json['business']['legalName'] = self._business.legal_name  # legal name easy fix
```
vysakh-menon-aot, line 125: Does this required? business model handled it (above ⬆️)

legal-api/src/legal_api/reports/report.py
```diff
@@ -465,6 +465,7 @@ def _populate_business_info_to_filing(filing: Filing, business: Business):
             business_json = VersionedBusinessDetailsService.get_business_revision(filing.transaction_id, business)
         else:
             business_json = business.json()
+            business_json['legalName'] = business.legal_name  # legal name easy fix
```
vysakh-menon-aot, line 468: same as above

legal-api/src/legal_api/reports/report.py
```diff
@@ -465,6 +465,7 @@ def _populate_business_info_to_filing(filing: Filing, business: Business):
             business_json = VersionedBusinessDetailsService.get_business_revision(filing.transaction_id, business)
```
vysakh-menon-aot, line 465: We get version data when we filing is completed. Might need to handle this scenario.

Do we need this change in NOA? if needed check line no 240

legal-api/src/legal_api/reports/business_document.py
chenhongjing: Yes. Outputs should use the old legal name. That is, it should be operating name for firms. But json now returns the correct legal name, which is not the operating name.

legal-api/src/legal_api/reports/report.py
chenhongjing: It will finally be updated [here](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/business_details_version.py#L609). So there's no need to change.

legal-api/src/legal_api/models/business.py
```diff
@@ -267,6 +271,45 @@ def identifier(self, value: str):
...
+        Return correct legal name based on feature flag.
+
+        Legal Name Easy Fix
+        """
+        from legal_api.services import flags  # pylint: disable=import-outside-toplevel
+
+        flag_on = flags.is_on('enable-legal-name-fix')
+        if self.legal_type in [
+            Business.LegalTypes.PARTNERSHIP.value,
+            Business.LegalTypes.SOLE_PROP.value
```
argush3, line 286: we can use `self.is_firm` I believe.

legal-api/src/legal_api/models/business.py
```diff
@@ -267,6 +271,42 @@ def identifier(self, value: str):
...
+                func.coalesce(Party.last_name+' ', '') +
+                func.coalesce(Party.first_name+' ', '') +
+                func.coalesce(Party.middle_initial, '')
+            )
+
+            parties_query = self.party_roles.join(Party).filter(
+                func.lower(PartyRole.role).in_([
+                    func.lower(value) for value in [
+                        PartyRole.RoleTypes.PARTNER.value,
+                        PartyRole.RoleTypes.PROPRIETOR.value
```
leodube-aot, line 296: Just curious why `func.lower` is needed here. Aren't these enum values already lower case?

legal-api/tests/unit/models/test_business.py
```diff
@@ -473,3 +476,66 @@ def test_amalgamated_into_business_json(session, test_name, existing_business_st
...
+
+@pytest.mark.parametrize('test_name, legal_type, flag_on', [
+    ('GP_FLAG_ON', 'GP', True),
+    ('SP_FLAG_ON', 'SP', True),
+    ('GP_FLAG_OFF', 'GP', False),
+    ('SP_FLAG_OFF', 'SP', False),
+    ('NON_FIRM_FLAG_ON', 'BC', True),
+    ('NON_FIRM_FLAG_OFF', 'BC', False),
+])
+def test_firm_business_json(session, test_name, legal_type, flag_on):
```
argush3, line 489: can we update this test so we can verify the scenario where there are more than 2 partners so can verify 'et al` is properly appended to the legal name?

legal-api/src/legal_api/models/business.py
chenhongjing: Sure, let me check 😸

argush3 (approved):

leodube-aot (approved):
Looks good!

tshyun24 (approved):
LGTM!

PR #2632, opened 2024-05-01
20996 Added alternateNames to slim json

JazzarKarim:
*Issue #:* /bcgov/entity#20996 

*Description of changes:*
Added alternateNames to slim json


legal-api/src/legal_api/models/business.py
```diff
@@ -469,6 +469,7 @@ def _slim_json(self):
         """Return a smaller/faster version of the business json."""
         d = {
             'adminFreeze': self.admin_freeze or False,
+            'alternateNames': self.get_alternate_names() or [],
```
argush3, line 472: if you add this here, can you remove it from `_extend_json` below?  And also make sure to include the FF check.

JazzarKarim:
Weird, the unit test passes locally:
test business json passing

PR #2635, opened 2024-05-03
19926 certificateOfNameCorrection

vysakh-menon-aot:
*Issue #:* /bcgov/entity#19926

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -274,7 +274,7 @@ class Report:  # pylint: disable=too-few-public-methods, too-many-lines
             self._format_registration_data(filing)
         elif self._report_key == 'changeOfRegistration':
             self._format_change_of_registration_data(filing, 'changeOfRegistration')
-        elif self._report_key == 'certificateOfNameChange':
+        elif self._report_key in ['certificateOfNameChange', 'certificateOfNameCorrection']:
             self._format_name_change_data(filing)
         elif self._report_key == 'certificateOfRestoration':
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/reports/report.py
```diff
@@ -1313,6 +1313,10 @@ class ReportMeta:  # pylint: disable=too-few-public-methods
             'filingDescription': 'Certificate of Name Change',
             'fileName': 'certificateOfNameChange'
         },
+        'certificateOfNameCorrection': {
+            'filingDescription': 'Certificate of Name Correction',
+            'fileName': 'certificateOfNameChange'
+        },
```
severinbeauvais, line 1319: Do you have a UI example of this?

severinbeauvais: Does the list of outputs look any different in Filings UI?

severinbeauvais: It is intentional that the fileName is the same as certificate of name change?

vysakh-menon-aot: filieName you see here is the filie name of the html template (using the same template) not the output file name

vysakh-menon-aot: image

severinbeauvais: If you think the output name and link (in Filings UI) work fine, go ahead and resolve this conversation.

severinbeauvais (approved):

argush3 (approved):

PR #2674, opened 2024-05-07
21025 create configuration endpoint

kzdev420:
*Issue #:* /bcgov/entity#21025

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/models/configuration.py
+++ b/legal-api/src/legal_api/models/configuration.py
@@ -82,12 +82,6 @@ class Configuration(db.Model):  # pylint: disable=too-many-instance-attributes
                 int(self.val)
             except ValueError as exc:
                 raise ValueError(f'Value for key {self.name} must be an integer') from exc
-            num_dissolutions_allowed = self.val if self.name == 'NUM_DISSOLUTIONS_ALLOWED'\
-                else Configuration.find_by_name('NUM_DISSOLUTIONS_ALLOWED').val
-            max_dissolutions_allowed = self.val if self.name == 'MAX_DISSOLUTIONS_ALLOWED'\
-                else Configuration.find_by_name('MAX_DISSOLUTIONS_ALLOWED').val
-            if int(num_dissolutions_allowed) > int(max_dissolutions_allowed):
-                raise ValueError('NUM_DISSOLUTIONS_ALLOWED cannot be greater than MAX_DISSOLUTIONS_ALLOWED.')
         elif self.name in bool_names:
             if self.val not in {'True', 'False'}:
                 raise ValueError(f'Value for key {self.name} must be a boolean')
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/models/configuration.py
```diff
@@ -64,6 +82,12 @@ def validate_value(self):
                 int(self.val)
             except ValueError as exc:
                 raise ValueError(f'Value for key {self.name} must be an integer') from exc
+            num_dissolutions_allowed = self.val if self.name == 'NUM_DISSOLUTIONS_ALLOWED'\
+                else Configuration.find_by_name('NUM_DISSOLUTIONS_ALLOWED').val
+            max_dissolutions_allowed = self.val if self.name == 'MAX_DISSOLUTIONS_ALLOWED'\
+                else Configuration.find_by_name('MAX_DISSOLUTIONS_ALLOWED').val
+            if int(num_dissolutions_allowed) > int(max_dissolutions_allowed):
+                raise ValueError('NUM_DISSOLUTIONS_ALLOWED cannot be greater than MAX_DISSOLUTIONS_ALLOWED.')
```
argush3, line 90: I don't think you should be doing this type of validation here.

We should do it further up where the resource is called.

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -0,0 +1,63 @@
...
+        return ({'message': 'Request body cannot be blank'}), HTTPStatus.BAD_REQUEST
+
+    configurations = json_input['configurations']
+
+    for data in configurations:
+        if data['name']:
+            configuration = Configuration.find_by_name(data['name'])
+            configuration.val = data.get('value', configuration.val)
+            configuration.full_description = data.get('full_description', configuration.full_description)
+            configuration.short_description = data.get('short_description', configuration.short_description)
```
argush3, line 57: for now we only support updating the value.  you can remove the description fields.

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -0,0 +1,63 @@
...
+        'configurations': [
+            configuration.json for configuration in configurations
+        ]
+    }), HTTPStatus.OK
+
+
+@bp.route('', methods=['PUT'])
+@cross_origin(origin='*')
+@jwt.has_one_of_roles([UserRoles.staff])
+def update_configurations():
```
argush3, line 43: can we put the core logic of this endpoint into a service class?

argush3, line 43: you'll need to make sure you have duplicate key validation and any other validation you haven't addressed that was mentioned in the ticket desc.

legal-api/src/legal_api/resources/v2/configuration.py
kzdev420: Removed

legal-api/src/legal_api/models/configuration.py
kzdev420: Updated

argush3:
@kzdev420 this PR can be closed right?

kzdev420:
> @kzdev420 this PR can be closed right?

Yes, This can be close

PR #2680, opened 2024-05-09
21025 create PUT configuration endpoint

kzdev420:
*Issue #:* /bcgov/entity#21025

*Description of changes:*

Screenshot 2024-05-09 at 1 08 54 PM


legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +35,69 @@ def get_configurations():
...
+                return {'message': f'Invalid value for {name}.'}, HTTPStatus.BAD_REQUEST
+
+        config = Configuration.find_by_name(name)
+        config.val = str(value)
+        try:
+            config.save()
+        except ValueError as e:
+            return {'message': str(e)}, HTTPStatus.BAD_REQUEST
+
+    return {'message': 'Configurations updated successfully'}, HTTPStatus.OK
```
argush3, line 74: Your response should be basically the updated items returned

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +35,69 @@ def get_configurations():
...
+    for data in input_data:
+        name = data.get('name')
+        if name not in valid_names:
+            return False, f'{name} is an invalid key.'
+        if name_count[name] == 1:
+            return False, f'{name} is duplicated.'
+        else:
+            name_count[name] += 1
+
+    return True, None
```
argush3, line 103: i think you should have a generic validation function that wraps the core validation logic which can call other independent validation functions if required.  

Also, II don't think we need to validate in such a hardcoded way for this scenario.  I think you should be able to make a db call to validate this.

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +35,69 @@ def get_configurations():
...
+        value = config_data.get('value')
+
+        if name in numeric_names:
+            if not is_valid_numeric_value(name, value):
+                return {'message': f'Invalid value for {name}.'}, HTTPStatus.BAD_REQUEST
+
+        config = Configuration.find_by_name(name)
+        config.val = str(value)
+        try:
+            config.save()
```
argush3, line 70: As per our conversation, it would be ideal if we only save entries only add the end. 

If there are any validation errors, we would roll back all updates to updated configuration.

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +35,69 @@ def get_configurations():
...
+
+    configurations = json_input.get('configurations', [])
+    if not configurations:
+        return {'message': 'Configurations list cannot be empty'}, HTTPStatus.BAD_REQUEST
+
+    valid_names, msg = has_validate_names(configurations)
+    if not valid_names:
+        return ({'message': msg}), HTTPStatus.BAD_REQUEST
+
+    numeric_names = {'NUM_DISSOLUTIONS_ALLOWED', 'MAX_DISSOLUTIONS_ALLOWED'}
```
argush3, line 57: I think we may want to update the model to have this there.  We may want to update some of Leo's code for the **validate_value** function to use these too.

legal-api/src/legal_api/resources/v2/configuration.py
kzdev420: Updated

legal-api/src/legal_api/resources/v2/configuration.py
kzdev420: Updated

legal-api/src/legal_api/resources/v2/configuration.py
kzdev420: Updated

legal-api/src/legal_api/resources/v2/configuration.py
kzdev420: Updated.
But, I am not sure it works well for you.

legal-api/src/legal_api/resources/v2/configuration.py
chenhongjing: I think it will be good to make [these lines](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/configuration.py#L76-L78) enums or something similar and also update the line here if they will be used frequently. 🤔

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +36,60 @@ def get_configurations():
...
+
+def is_valid_numeric_value(name, value):
+    """Check if the numeric value is valid."""
+    if name == 'NUM_DISSOLUTIONS_ALLOWED':
+        max_value = Configuration.find_by_name('MAX_DISSOLUTIONS_ALLOWED').val
+        return int(value) <= int(max_value)
+    elif name == 'MAX_DISSOLUTIONS_ALLOWED':
+        min_value = Configuration.find_by_name('NUM_DISSOLUTIONS_ALLOWED').val
+        return int(value) >= int(min_value)
+    return False
```
argush3, line 95: this should still be taken care of by the model.  

All other types of validation can happen outside of the model I think.

argush3: actually...i see what you are trying to do.  Ignore my comment here for now.

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +36,60 @@ def get_configurations():
             configuration.json for configuration in configurations
         ]
     }), HTTPStatus.OK
+
+
+@bp.route('', methods=['PUT'])
+@cross_origin(origin='*')
+# @jwt.has_one_of_roles([UserRoles.staff])
```
vysakh-menon-aot, line 43: Do we not need to authenticate?

chenhongjing, line 43: Just a reminder to uncomment it before it can be merged. 👀

chenhongjing: It should be staff only.

kzdev420: updated

kzdev420: We need, updated

legal-api/src/legal_api/models/configuration.py
```diff
@@ -88,6 +88,8 @@ def validate_value(self):
         elif self.name in cron_names:
             if not croniter.is_valid(self.val):
                 raise ValueError(f'Value for key {self.name} must be a cron string')
+        else:
+            raise ValueError(f'{self.name} is an invalid key.')
```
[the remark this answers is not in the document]
kzdev420: I thought the request body can have the wrong key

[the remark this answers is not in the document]
chenhongjing: Is it possible to update configuration with a key that does not exist?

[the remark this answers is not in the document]
argush3: we should remove this.

I haven't thought this out entirely but my initial thought on configuration validation is you should construct an initial query to pull out all the key names in the request body and use that for a query against the configuration.  If the count of the results coming back from the query coming back doesn't match the number of keys in the request body, then you return a validation error that indicates there is a duplicate or invalid key error.

And I think you'll also be able to just use the configuration items that come back for the eventual updates maybe.

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +36,60 @@ def get_configurations():
...
+            config.val = str(value)
+            db.session.add(config)
+            response.append(config.json)
+        db.session.commit()
+    except ValueError as e:
+        # Rollback transaction
+        db.session.rollback()
+        return {'message': str(e)}, HTTPStatus.BAD_REQUEST
+
+    return {'results': response}, HTTPStatus.OK
```
argush3, line 84: `results` should be `configurations`

kzdev420: updated

argush3:
@kzdev420 here's an [example draft PR ](https://github.com/bcgov/lear/pull/2687)that covers some of the core structure of how I was thinking to structure things.  Things can be refined and I having tested this thoroughly.  You'll want to refine things when you move code over to  your branch.

One thing to note is we'll want to add something to the model to represent the column names that we need to reference in the validation and etc.  Maybe an enum for each data type that gets validated.

legal-api/src/legal_api/models/configuration.py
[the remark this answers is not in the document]
argush3: resolving, reference draft PR for resolution

legal-api/src/legal_api/resources/v2/configuration.py
argush3: resolving, reference draft PR for resolution

legal-api/src/legal_api/resources/v2/configuration.py
kzdev420: Removed this

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +36,114 @@ def get_configurations():
...
+
+def validate_configurations(configurations):
+    """Validate the configurations before updating."""
+    if not configurations:
+        return 'Configurations list cannot be empty'
+
+    # Extract names from the requested configuration updates
+    names = [config['name'] for config in configurations]
+    if len(names) != len(set(names)):
+        return 'Duplicate names error.'
```
argush3, line 81: i think there is a scenario where duplicates won't be picked up here...not sure.  think we may want to uppercase all the entries in `names`

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +36,114 @@ def get_configurations():
...
+    error_message = validate_configurations(configurations)
+    if error_message:
+        return {'message': error_message}, HTTPStatus.BAD_REQUEST
+
+    response = []
+    try:
+        for config_data in configurations:
+            name = config_data.get('name')
+            value = config_data.get('value')
+            config = Configuration.find_by_name(name)
```
argush3, line 60: Please update `Configuration.find_by_name` to be case insensitive.

legal-api/src/legal_api/models/configuration.py
```diff
@@ -73,21 +73,26 @@ def find_by_name(cls, config_name: str) -> Configuration:
...
+        """Ensure the value is the correct type before insert or update."""
         int_names = {'NUM_DISSOLUTIONS_ALLOWED', 'MAX_DISSOLUTIONS_ALLOWED'}
         bool_names = {'DISSOLUTIONS_ON_HOLD'}
         cron_names = {'NEW_DISSOLUTIONS_SCHEDULE'}
 
-        if self.name in int_names:
+        if name in int_names:
             try:
-                int(self.val)
+                int(val)
```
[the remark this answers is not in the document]
kzdev420: Added

legal-api/src/legal_api/resources/v2/configuration.py
kzdev420: Updated

kzdev420:
> @kzdev420 here's an [example draft PR ](https://github.com/bcgov/lear/pull/2687)that covers some of the core structure of how I was thinking to structure things. Things can be refined and I haven't tested this thoroughly. You'll want to refine things when you move code over to your branch.
> 
> One thing to note is we'll want to add something to the model to represent the column names that we need to reference in the validation and etc. Maybe an enum for each data type that gets validated. You'll need to do some analysis to see what makes sense.

Created the enum for names

legal-api/tests/unit/resources/v2/test_configuration.py
```diff
@@ -49,3 +50,65 @@ def test_get_configurations_with_invalid_user(app, session, client, jwt):
...
+     'Duplicate names error.', HTTPStatus.BAD_REQUEST),
+    ('dissolution_hold', {'configurations': [{'name': 'DISSOLUTIONS_ON_HOLD','value': '1'}]},
+     'Value for key DISSOLUTIONS_ON_HOLD must be a boolean', HTTPStatus.BAD_REQUEST),
+    ('dissolution_schedule', {'configurations': [{'name': 'NEW_DISSOLUTIONS_SCHEDULE','value': '1'}]},
+     'Value for key NEW_DISSOLUTIONS_SCHEDULE must be a cron string', HTTPStatus.BAD_REQUEST),
+    ('blank_request_body', None, 'Request body cannot be blank', HTTPStatus.BAD_REQUEST),
+    ('request_body_without_configuration_list', {'configurations': []}, 'Configurations list cannot be empty', HTTPStatus.BAD_REQUEST),
+])
+def test_put_configurations_with_invalid_data(app, session, client, jwt, test_name, input_data, message, status_code):
+    """Assert that failure update."""
```
argush3, line 106: can we add a test to test the negative integer scenario?

legal-api/src/legal_api/resources/v2/configuration.py
argush3: Can we make this change in the `find_by_name` function instead?

kzdev420: I made it at find_by_name too.
Screenshot 2024-05-15 at 10 13 52 AM

legal-api/tests/unit/resources/v2/test_configuration.py
kzdev420: Added

argush3:
looks good!  just need to fix the linting

argush3 (approved):

chenhongjing (approved):
LGTM 😺

PR #2681, opened 2024-05-09
21202 Integrate disable-nr-check feature flag with NR validate code

argush3:
*Issue #:* /bcgov/entity#21202

*Description of changes:*
**Note: this feature flag is only intended to disable NR check in the sandbox.**

* Integrate `disable-nr-check` feature flag with namex service's `nr_validate` function.  Always returns that NR is valid when flag is on.
* Fix broken tests
* Add tests to test valid NR is always returned when flag is on

### DIFF
--- a/legal-api/flags.json
+++ b/legal-api/flags.json
@@ -3,6 +3,7 @@
         "string-flag": "a string value",
         "bool-flag": true,
         "integer-flag": 10,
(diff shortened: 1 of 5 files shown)


thorwolpert (approved):
thanks for adding the comments around it as well.

PR #2687, opened 2024-05-13
Example Put configuration implementation

argush3:
*Issue #:* /bcgov/entity###

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/resources/v2/configuration.py
+++ b/legal-api/src/legal_api/resources/v2/configuration.py
@@ -66,11 +66,10 @@ def update_configurations():
             config = Configuration.find_by_name(name)
             if not config:
                 raise ValueError(f'{name} is an invalid key.')
-            
+
             if name_count.get(name, None) == 1:
                 raise ValueError(f'{name} is duplicated.')
-            else:
-                name_count = {name: 1, **name_count}
+            name_count = {name: 1, **name_count}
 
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/services/utils.py
```diff
@@ -86,3 +86,21 @@ def get_int(filing: Dict, path: str) -> str:
...
+        key_value = data.get(key_attr)
+
+        if key_value in key_count:
+            if key_count[key_value] == 1:
+                duplicated_keys.append(key_value)
+            key_count[key_value] += 1
+        else:
+            key_count[key_value] = 1
+
+    return duplicated_keys
```
argush3, line 106: we can remove this now

legal-api/src/legal_api/resources/v2/configuration.py
```diff
@@ -35,3 +37,121 @@ def get_configurations():
...
+            value = config_data.get('value')
+            config = Configuration.find_by_name(name)
+            config.val = str(value)
+            db.session.add(config)
+            response.append(config.json)
+        db.session.commit()
+    except ValueError as e:
+        # Rollback transaction
+        db.session.rollback()
+        return {'message': str(e)}, HTTPStatus.BAD_REQUEST
```
argush3, line 69: we need to catch all other exception types as well

PR #2691, opened 2024-05-15
21196 Misc configurations table updates and other updates

chenhongjing:
*Issue #:* /bcgov/entity#21196

*Description of changes:*
- Create new alembic script to update `configurations` table
- Update data model
- Fix unit tests


legal-api/src/legal_api/models/configuration.py
```diff
@@ -23,6 +24,13 @@
 from .db import db
 
 
+EMAIL_PATTERN = (r'^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|'
+                 r'(".+"))@'
+                 r'((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|'
+                 r'(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
+                 )
```
chenhongjing, line 31: This is the pattern that I grabbed from FE, but I'm not sure if we need to validate email pattern before inserting or updating.

kzdev420 (approved):

legal-api/src/legal_api/models/configuration.py
argush3: we can just use this for now

argush3 (approved):

JazzarKarim:
Please let me know when this is merged as I need to make FE changes to accommodate the changes here.

PR #2693, opened 2024-05-16
21202 Add additional checks to disable NR related functionality when disable-nr-check flag is on

argush3:
*Issue #:* /bcgov/entity#21202

*Description of changes:*
* Update to return with no NR validation errors for all top level functions that validate NRs when NR check flag is off
* Update get_approved_name in namex service to return an approved or conditionally approved name regardless of NR's state when NR check flag is off.  This is called separately in some places in code base.
* Update/add tests


thorwolpert (approved):
thanks!

PR #2698, opened 2024-05-17
21123 add in dissolution property to business model

eason-pan-bc:
*Issue #:* /bcgov/entity#21123

*Description of changes:*
- Add in_dissolution property to business model
  - checks that a business belongs to a batches entry where status is not COMPLETED
  - checks that a business has a batch_processing entry that matches on business_id and status is not COMPLETED or WITHDRAWN
- Add unit tests for the logic added above

### DIFF
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -420,19 +420,17 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
     def in_dissolution(self):
         """Return true if in dissolution, otherwise false."""
         # check a business has a batch_processing entry that matches business_id and status is not COMPLETED
-        find_in_batch_processing = db.session.query(BatchProcessing).filter(
-            BatchProcessing.business_id == self.id,
-            BatchProcessing.status != BatchProcessing.BatchProcessingStatus.COMPLETED,
-            BatchProcessing.status != BatchProcessing.BatchProcessingStatus.WITHDRAWN,
-        ).one_or_none()
-        if not find_in_batch_processing:
-            return False
-        # check the business belongs to a batch in batches where status is not COMPLETED
-        check_batches = db.session.query(Batch). \
-            filter(Batch.id == find_in_batch_processing.batch_id). \
-            filter(Batch.status != Batch.BatchStatus.COMPLETED). \
+        find_in_batch_processing = db.session.query(BatchProcessing, Batch).\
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/models/business.py
```diff
@@ -414,6 +416,24 @@ def good_standing(self):
                 return date_cutoff.replace(tzinfo=pytz.UTC) > datetime.utcnow()
         return True
 
+    @property
+    def in_dissolution(self):
```
chenhongjing, line 420: Can you simplify this to one query?
For example,
``` python
      check = db.session(Batch, BatchProcessing).\
          filter(Batch.id == BatchProcessing.batch_id).\
          filter(Batch.batch_type == Batch.BatchType.INVOLUNTARY_DISSOLUTION.value).\
          filter(<MORE CONDITIONS>).\
          one_or_none()
```

legal-api/src/legal_api/models/business.py
```diff
@@ -414,6 +416,24 @@ def good_standing(self):
...
+        """Return true if in dissolution, otherwise false."""
+        # check a business has a batch_processing entry that matches business_id and status is not COMPLETED
+        find_in_batch_processing = db.session.query(BatchProcessing).filter(
+            BatchProcessing.business_id == self.id,
+            BatchProcessing.status != BatchProcessing.BatchProcessingStatus.COMPLETED,
+            BatchProcessing.status != BatchProcessing.BatchProcessingStatus.WITHDRAWN,
```
chenhongjing, line 426: Also, this can be simplified too 😺 
```
BatchProcessing.status.notin_([BatchProcessing.BatchProcessingStatus.COMPLETED.value,
                                                  BatchProcessing.BatchProcessingStatus.WITHDRAWN.value])
```

leodube-aot (approved):
Looks good once Hongjing's comments have been addressed!

legal-api/src/legal_api/models/business.py
eason-pan-bc: Thank you for the feedback, I'm going to simplify the query. 
Just out of curiosity, Is the batch_type check for the future? Since currently we have only one type in the batch_type Enum.

chenhongjing: Yes

eason-pan-bc:
adjusted the improvements pointed out by @chenhongjing

kzdev420 (approved):

argush3 (approved):

PR #2706, opened 2024-05-22
21262 Added inDissolution to slim json

JazzarKarim:
*Issue #:* /bcgov/entity#21262

*Description of changes:*
Moved the ~warnings array~ inDissolution to slim json

Currently, in Auth Web, the search endpoint for all the businesses returns the slim json:
something like this serach

So, as of right now, there's no way of actually knowing that the business is the process of being dissolved (D1 state for example).

Now that the inDissolution is in the slim json, I will be able to use it in AUTH WEB. And In doing so, I can actually add the alert when inDissolution is true.

### DIFF
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -465,7 +465,6 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
             'foundingDate': self.founding_date.isoformat(),
             'hasRestrictions': self.restriction_ind,
             'complianceWarnings': self.compliance_warnings,
-            'warnings': self.warnings,
             'lastAnnualGeneralMeetingDate': datetime.date(self.last_agm_date).isoformat() if self.last_agm_date else '',
             'lastAnnualReportDate': datetime.date(self.last_ar_date).isoformat() if self.last_ar_date else '',
             'lastLedgerTimestamp': self.last_ledger_timestamp.isoformat(),
@@ -496,7 +495,8 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
             'identifier': self.identifier,
             'legalName': self.business_legal_name,
             'legalType': self.legal_type,
-            'state': self.state.name if self.state else Business.State.ACTIVE.name
+            'state': self.state.name if self.state else Business.State.ACTIVE.name,
+            'warnings': self.warnings
         }
 
         if flags.is_on('enable-legal-name-fix'):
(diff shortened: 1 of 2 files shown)


chenhongjing:
I'm confused and I don't think `no_dissolution` is helpful for showing the alert. Because it is used as a flag for excluding businesses from selecting to dissolution when it is set True.

JazzarKarim:
> I'm confused and I don't think `no_dissolution` is helpful for showing the alert. Because it is used as a flag for excluding businesses from selecting to dissolution when it is set True.

Yep, you're absolutely right. I changed it!

AimeeGao (approved):
Clean and looks great!

eason-pan-bc:
If the warning shows the specific stage of a dissolution process, maybe consider having the _step_ information from the _batch_processing_ table (the enum is `BatchProcessingStep` in code)? Since it's directly indicating the specific step (_WARNING_LEVEL_1 / WARNING_LEVEL_2 / DISSOLUTION_)

argush3 (approved):

chenhongjing:
> > I'm confused and I don't think `no_dissolution` is helpful for showing the alert. Because it is used as a flag for excluding businesses from selecting to dissolution when it is set True.
> 
> Yep, you're absolutely right. I changed it!

Another question, `in_dissolution` cannot distinguish D1 & D2 states for now. Do you think it is okay to use it or you may have a combined check with others?

JazzarKarim:
> If the warning shows the specific stage of a dissolution process, maybe consider having the _step_ information from the _batch_processing_ table (the enum is `BatchProcessingStep` in code)? Since it's directly indicating the specific step (_WARNING_LEVEL_1 / WARNING_LEVEL_2 / DISSOLUTION_). Otherwise, looks great!

Good point! From the designs, the alert will not indicate the stage. It will just show "This business is in the process of being dissolved.". So whether it's in D1 or D2, I don't think it'll matter much for the UI.

JazzarKarim:
> > > I'm confused and I don't think `no_dissolution` is helpful for showing the alert. Because it is used as a flag for excluding businesses from selecting to dissolution when it is set True.
> > 
> > 
> > Yep, you're absolutely right. I changed it!
> 
> Another question, `in_dissolution` cannot distinguish D1 & D2 states for now. Do you think it is okay to use it or you may have a combined check with others?

Yep, as per my reply to Eason, the UI doesn't distinguish between both states as well. The text/alert will be the same regardless of the state.

eason-pan-bc (approved):

chenhongjing:
> > > > I'm confused and I don't think `no_dissolution` is helpful for showing the alert. Because it is used as a flag for excluding businesses from selecting to dissolution when it is set True.
> > > 
> > > 
> > > Yep, you're absolutely right. I changed it!
> > 
> > 
> > Another question, `in_dissolution` cannot distinguish D1 & D2 states for now. Do you think it is okay to use it or you may have a combined check with others?
> 
> Yep, as per my reply to Eason, the UI doesn't distinguish between both states as well. The text/alert will be the same regardless of the state.

Got it. Thanks for clarifying!

chenhongjing (approved):

kzdev420 (approved):

PR #2722, opened 2024-05-29
21455 add criterion to exclude frozen businesses from being selected to involuntary dissolution

eason-pan-bc:
*Issue #:* /bcgov/entity#21455

*Description of changes:*
- Add a criterion to exclude frozen businesses by checking _Business.admin_freeze_ is not True
- Update unit tests, expecting business eligibility returns false when the business is admin frozen
- Passed all unit tests in _test_involuntary_dissolution.py_ including the updated one

### DIFF
--- a/legal-api/src/legal_api/services/involuntary_dissolution.py
+++ b/legal-api/src/legal_api/services/involuntary_dissolution.py
@@ -59,7 +59,7 @@ class InvoluntaryDissolutionService():
                                   Batch.batch_type == Batch.BatchType.INVOLUNTARY_DISSOLUTION)
 
         query = db.session.query(Business).\
-            filter(not_(Business.admin_freeze == True)).\
+            filter(not_(Business.admin_freeze.is_(True))).\
             filter(Business.state == Business.State.ACTIVE).\
             filter(Business.legal_type.in_(eligible_types)).\
             filter(Business.no_dissolution.is_(False)).\
--- a/legal-api/src/legal_api/services/involuntary_dissolution.py
+++ b/legal-api/src/legal_api/services/involuntary_dissolution.py
@@ -59,6 +59,7 @@ class InvoluntaryDissolutionService():
                                   Batch.batch_type == Batch.BatchType.INVOLUNTARY_DISSOLUTION)
 
         query = db.session.query(Business).\
+            filter(not_(Business.admin_freeze == True)).\
             filter(Business.state == Business.State.ACTIVE).\
             filter(Business.legal_type.in_(eligible_types)).\
             filter(Business.no_dissolution.is_(False)).\
--- a/legal-api/tests/unit/services/test_involuntary_dissolution.py
+++ b/legal-api/tests/unit/services/test_involuntary_dissolution.py
@@ -296,10 +296,7 @@ def test_get_businesses_eligible_query_xpro_from_nwpta(session, test_name, juris
 def test_exclude_admin_frozen_businesses(session, test_name, admin_freeze, eligible):
     """Assert service returns eligible business excluding admin frozen businesses"""
     identifier = 'BC1234567'
-    business = factory_business(identifier=identifier, admin_freeze=admin_freeze, entity_type=Business.LegalTypes.COMP.value)
-    if not eligible:
-        business.no_dissolution = True
-        business.save()
+    factory_business(identifier=identifier, admin_freeze=admin_freeze, entity_type=Business.LegalTypes.COMP.value)
 
     check_query = InvoluntaryDissolutionService._get_businesses_eligible_query().\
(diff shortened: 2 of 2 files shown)


legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -59,6 +59,7 @@ def _get_businesses_eligible_query():
                                   Batch.batch_type == Batch.BatchType.INVOLUNTARY_DISSOLUTION)
 
         query = db.session.query(Business).\
+            filter(not_(Business.admin_freeze.is_(True))).\
```
chenhongjing, line 62: This can be simplified to use `Business.admin_freeze.is_(False)`.

legal-api/tests/unit/services/test_involuntary_dissolution.py
```diff
@@ -39,15 +39,16 @@
...
-            ('TEST_ELIGIBLE', True),
-            ('TEST_INELIGIBLE', False)
+        'test_name, isAdminFreeze, eligible', [
+            ('TEST_ELIGIBLE', False, True),
+            ('TEST_INELIGIBLE', False, False),
+            ('TEST_INELIGIBLE_ADMIN_FREEZE', True, False)
         ]
 )
-def test_check_business_eligibility(session, test_name, eligible):
+def test_check_business_eligibility(session, test_name, isAdminFreeze, eligible):
```
chenhongjing, line 48: Can you revert the changes here and write separate tests for your work? This test I wrote is to check the general functionality of method `check_business_eligibility` instead of the underlying query. The query will be used in more than one method. It's better to write new & seperate tests to check whether frozen buisnesses are excluded.

You can use this [test](https://github.com/bcgov/lear/blob/main/legal-api/tests/unit/services/test_involuntary_dissolution.py#L118) as reference.

legal-api/src/legal_api/services/involuntary_dissolution.py
eason-pan-bc: I checked the db, there are some entries, actually a lot, with `null` value in the admin_freeze column. Should I ignore these?
image

legal-api/tests/unit/services/test_involuntary_dissolution.py
eason-pan-bc: Thanks for letting me know, going to work on it.

legal-api/src/legal_api/services/involuntary_dissolution.py
chenhongjing: Yep, should be good to keep your change. Ignore my comment.

legal-api/tests/unit/services/test_involuntary_dissolution.py
argush3, line 48: can we use the standard convention of using an underscore between word for param and variable naming?  i.e. `isAdminFreeze` -> `is_admin_freeze`

eason-pan-bc: thank you for pointing out, working on it

eason-pan-bc: created new unit tests to check query results excluded the business with admin_freeze is true, along testing eligibility returns the results as expected.

eason-pan-bc: fixed, please review

legal-api/tests/unit/services/test_involuntary_dissolution.py
```diff
@@ -285,3 +285,25 @@ def test_get_businesses_eligible_query_xpro_from_nwpta(session, test_name, juris
...
+        ('BUSINESS_ADMIN_FREEZE', True, False)
+    }
+)
+def test_exclude_admin_frozen_businesses(session, test_name, admin_freeze, eligible):
+    """Assert service returns eligible business excluding admin frozen businesses"""
+    identifier = 'BC1234567'
+    business = factory_business(identifier=identifier, admin_freeze=admin_freeze, entity_type=Business.LegalTypes.COMP.value)
+    if not eligible:
+        business.no_dissolution = True
+        business.save()
```
argush3, line 302: Do you need this?  I think it makes it hard to tell if the admin_freeze exclusion is really working.

eason-pan-bc: make sense! going to remove and test it again

eason-pan-bc: updated and passed, that's a very good point

argush3 (approved):

JazzarKarim (approved):
Nice! LGTM

PR #2714, opened 2024-05-24
21092 Dissolution Job - add logic to initiate dissolution process for businesses (first pass)

leodube-aot:
*Issue #:* /bcgov/entity#21092

*Description of changes:*
- Added logic to initiate dissolution process for businesses 
- First pass at unit tests
- This is meant to be an initial PR to unblock others while I iron out some issues with it locally.

*Notes:*

_Edit: fixed!_

I cannot seem to get the requirements installed correctly on my machine, and so I am unable to run the job or tests locally to verify they work. The issue I am getting is:

Screenshot 2024-05-24 at 10 44 48 AM

However, it does install the requirements without issue on Docker, but I get a different error when trying to run it on Docker:

Screenshot 2024-05-24 at 10 45 27 AM

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #2736, opened 2024-06-04
21331 Dissolutions Job - Implement stage 2

chenhongjing:
*Issue #:* /bcgov/entity#21331

*Description of changes:*
- Implemented stage 2 for dissolutions job
- Add unit tests
- Misc updates

### DIFF
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -140,8 +140,8 @@ def initiate_dissolution_process(app: Flask):  # pylint: disable=redefined-outer
 
 def stage_2_process(app: Flask):
     """Run dissolution stage 2 process for businesses meet moving criteria."""
-    if not (stage_2_delay := app.config.get('STAGE_2_DELAY', None)):
-        app.logger.debug('Skipping stage 2 run since config STAGE_2_DELAY is missing.')
+    if not (stage_1_delay := app.config.get('STAGE_1_DELAY', None)):
+        app.logger.debug('Skipping stage 2 run since config STAGE_1_DELAY is missing.')
         return
 
     batch_processings = (
@@ -156,7 +156,7 @@ def stage_2_process(app: Flask):
             BatchProcessing.step == BatchProcessing.BatchProcessingStep.WARNING_LEVEL_1
         )
         .filter(
-            BatchProcessing.created_date + text(f"""INTERVAL '{stage_2_delay} DAYS'""")
+            BatchProcessing.created_date + text(f"""INTERVAL '{stage_1_delay} DAYS'""")
             <= func.timezone('UTC', func.now())
         )
         .all()
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -220,7 +220,9 @@ async def run(loop, application: Flask = None):  # pylint: disable=redefined-out
             # check if batch can be run today
             stage_1_valid, stage_2_valid, stage_3_valid = check_run_schedule()
             if not any([stage_1_valid, stage_2_valid, stage_3_valid]):
-                application.logger.debug('Skipping job run since current day of the week does not match any cron schedule.')  # noqa: E501
+                application.logger.debug(
+                    'Skipping job run since current day of the week does not match any cron schedule.'
+                )
                 return
 
             if stage_1_valid:
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -25,7 +25,7 @@ from legal_api.models import Batch, BatchProcessing, Configuration, db  # noqa: 
 from legal_api.services.flags import Flags
 from legal_api.services.involuntary_dissolution import InvoluntaryDissolutionService
 from sentry_sdk.integrations.logging import LoggingIntegration
-from sqlalchemy import Date, cast
+from sqlalchemy import Date, cast, func, text
 
 import config  # pylint: disable=import-error
 from utils.logging import setup_logging  # pylint: disable=import-error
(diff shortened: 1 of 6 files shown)


jobs/involuntary-dissolutions/Makefile
```diff
@@ -39,7 +39,7 @@ clean-test: ## clean test files
 build-req: clean ## Upgrade requirements
 	test -f venv/bin/activate || python3.8 -m venv  $(CURRENT_ABS_DIR)/venv ;\
 	. venv/bin/activate ;\
-	pip install --upgrade pip ;\
```
chenhongjing, line 42: I reverted the changes in Makefile back to previous version because it resolved the `ModuleNotFoundError: No module named 'involuntary_dissolutions'` I got when running tests.

jobs/involuntary-dissolutions/involuntary_dissolutions.py
```diff
@@ -73,7 +74,7 @@ def shell_context():
 
 
 def initiate_dissolution_process(app: Flask):  # pylint: disable=redefined-outer-name
-    """Initiate dissolution process for new businesses that meet dissolution criteria"""
+    """Initiate dissolution process for new businesses that meet dissolution criteria."""
     try:
         # check if batch has already run today
         batch_today = Batch.find_by(batch_type=Batch.BatchType.INVOLUNTARY_DISSOLUTION, start_date=datetime.today())
```
leodube-aot, line 80: You can remove the 'start_date' parameter here to fix the linting error.

I had originally planned to update the 'find_by' function in the 'batch' model but I think I will just use a query here like you do for 'batch_processing' on line 116. I'll update the logic part in my PR.

jobs/involuntary-dissolutions/involuntary_dissolutions.py
```diff
@@ -116,13 +180,18 @@ async def run(loop, application: Flask = None):  # pylint: disable=redefined-out
...
-            new_dissolutions_schedule_config = Configuration.find_by_name(config_name='DISSOLUTIONS_STAGE_1_SCHEDULE')
-            tz = pytz.timezone('US/Pacific')
-            cron_valid = croniter.match(new_dissolutions_schedule_config.val, tz.localize(datetime.today()))
-            if cron_valid:
+            cron_valids = check_run_schedule()
+            if not any(cron_valids):
+                application.logger.debug('Skipping job run since current day of the week does not match any cron schedule.')  # noqa: E501
+                return
+
+            if cron_valids[0]:
```
leodube-aot, line 188: Do you think it would make sense to return a `dict` from `check_run_schedule` instead of a list to improve readability? 

We could then check this as `cron_valids['stage_1']`. I'm not sure how we handle similar situations elsewhere in the code.

Otherwise I really like the new `check_run_schedule` function!

jobs/involuntary-dissolutions/Makefile
leodube-aot: Hmm, weird. The `--upgrade` change [was a workaround for a dependency error](https://github.com/bcgov/lear/pull/2714#issuecomment-2135395063) that was happening on my mac (and maybe Argus' mac too), but obviously we don't want it breaking the unit tests.

I've been messing around with this a bit to see if I can find a middle ground where the requirements get installed correctly and the unit tests still run. I've found that `pip install pip==21.0` works for both cases but throws a bunch of deprecation warnings during the tests.

I plan to keep experimenting a bit but I will make the change in my PR if I find a version that works for both cases.

jobs/involuntary-dissolutions/requirements.txt
```diff
@@ -26,5 +26,5 @@ six==1.15.0
 urllib3==1.26.11
 requests==2.25.1
 cachelib==0.1.1
-git+https://github.com/bcgov/lear.git#egg=legal_api&subdirectory=legal-api
+legal_api[nats] @ git+https://github.com/bcgov/lear.git#egg=legal_api&subdirectory=legal-api
```
chenhongjing, line 29: @leodube-aot This should work now if you run into problem on installing all legal-api dependencies. 👀

jobs/involuntary-dissolutions/involuntary_dissolutions.py
argush3: My thoughts are maybe `check_run_schedule` is a little overloaded.

Maybe have a function that checks if a given cron string can run today.

And `check_run_schedule` can still check the 3 schedules.

To make things more readable as Leo suggested, maybe call `check_run_schedule` with `stage_1_valid, stage_2_valid, stage_3_valid = check_run_schedule()`.

We can also use a dataclass to return the validity of the stages.

leodube-aot (approved):
LGTM!

jobs/involuntary-dissolutions/involuntary_dissolutions.py
```diff
@@ -138,6 +138,76 @@ def initiate_dissolution_process(app: Flask):  # pylint: disable=redefined-outer
...
+        )
+        if eligible:
+            batch_processing.step = BatchProcessing.BatchProcessingStep.WARNING_LEVEL_2
+        else:
+            batch_processing.status = BatchProcessing.BatchProcessingStatus.WITHDRAWN
+
+            batch_processing.notes = (
+                batch_processing.notes + ', ' if batch_processing.notes else ''
+            )
+            batch_processing.notes += 'Moved back into good standing'
```
argush3, line 179: I think maybe we can just replace whatever is notes entirely for now.

jobs/involuntary-dissolutions/involuntary_dissolutions.py
```diff
@@ -148,13 +218,20 @@ async def run(loop, application: Flask = None):  # pylint: disable=redefined-out
         application.logger.debug(f'enable-involuntary-dissolution flag on: {flag_on}')
         if flag_on:
             # check if batch can be run today
-            new_dissolutions_schedule_config = Configuration.find_by_name(config_name='DISSOLUTIONS_STAGE_1_SCHEDULE')
-            tz = pytz.timezone('US/Pacific')
-            cron_valid = croniter.match(new_dissolutions_schedule_config.val, tz.localize(datetime.today()))
-            if cron_valid:
+            stage_1_valid, stage_2_valid, stage_3_valid = check_run_schedule()
```
argush3, line 221: can we add some logging to display the stage validity values returned by `check_run_schedule`?  

Think it will help with debugging in the near future.

jobs/involuntary-dissolutions/tests/unit/test_involuntary_dissolutions.py
```diff
@@ -84,3 +104,107 @@ def test_initiate_dissolution_process(app, session):
...
+        status=status,
+        step=step,
+        created_date=created_date,
+        last_modified=last_modified
+    )
+
+    stage_2_process(app)
+
+    if found:
+        assert batch_processing.last_modified != last_modified
```
argush3, line 173: can we also assert it gets updated to warning level 2?

argush3 (approved):

PR #2746, opened 2024-06-07
21475 - Add trigger_date to batch_processing table

chenhongjing:
*Issue #:* /bcgov/entity#21475

This is the first PR for the ticket. More changes will come in the subsequent PRs.

*Description of changes:*
- Added alembic script
- Updated data model

*batch_processing table after the change*

### DIFF
--- a/legal-api/src/legal_api/models/batch_processing.py
+++ b/legal-api/src/legal_api/models/batch_processing.py
@@ -14,12 +14,18 @@
 """This module holds data for batch processing."""
 from enum import auto
 
+from datedelta import datedelta
 from sqlalchemy.dialects.postgresql import JSONB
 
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/models/batch_processing.py
```diff
@@ -22,6 +23,11 @@
 from .db import db
 
 
+def default_trigger_date():
+    """Set default value for trigger_date."""
+    return datetime.utcnow() + datedelta(days=42)
```
argush3, line 28: Don't think we should have a default trigger date.

The batch related tables may be used for other batch related processes in the future.

argush3 (approved):

PR #2749, opened 2024-06-10
21330 Filer - Add involuntary dissolution unit tests

leodube-aot:
*Issue #:* /bcgov/entity#21330

*Description of changes:*
- Added unit tests for involuntary dissolutions to filer


chenhongjing:
> The CI is failing due to the new `testcontainers==4.5.1` requirement, which is not compatible with python 3.8 ([it requires python >= 3.9](https://pypi.org/project/testcontainers/)). This causes the requirements in `requirements/dev.txt` to not install, resulting in the dependency error.
> 
> Screenshot 2024-06-11 at 7 57 05 AM
> I tried downgrading to the latest version that supports python 3.8 `testcontainers==3.7.1`, and was able to get the linting step to work, but it still fails on the testing step ([see the CI logs for that commit here](https://github.com/bcgov/lear/actions/runs/9467161578/job/26080566894)).
> 
> With the downgraded package I was able to get the unit tests I added to run locally by running `PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python pytest tests/unit/filing_processors/test_dissolution.py`
> 
> Screenshot 2024-06-11 at 8 31 33 AM
> I have reverted the package downgrade for now since I am unsure if we plan to use specific functionality from later versions of `testcontainers`.
> 
> CC: @chenhongjing Looks like you've ran into the same CI issue as me on your filer draft PR, thought I'd share what I've been able to figure out so far about the problem.

Thanks for your sharing😸!  I got almost the same issues like yours. Here are my findings.

Apart from the `testcontainers` error. I got this error when I ran filer. It's the same error that is shown in CI logs.
- My PC
```
line 194, in _run_module_as_main (Current frame)
    return _run_code(code, main_globals, None,
TypeError: Descriptors cannot be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
```
- CI logs

image

From my understanding, there's dependency conflicts between GCP stuff & NATS. Then, I downgraded protobuf to `protobuf==3.19.5`. This at least made filer & unit tests run so that I could verify my code changes. But I still got conflict message.

Pasted image 20240610091759


Besides, as far as I know, changes to integrate GCP pub/sub stuff are still undegoing. We need to wait to see what will happen.

chenhongjing (approved):

argush3:
Looks good and thanks for the info on the dependency issues.  We'll need to get that sorted out at some point.

The one thing I was unsure of was the amalgamation code here https://github.com/bcgov/lear/blob/ff6bb08b6eb20575da031c0fc052ac8176d1cd3f/queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py#L59-L62.

But after talking with Vysakh, it sounds like we don't to worry about it for involuntary dissolutions.

>i dont think you need to do that. admin dissolve assumes that this amalgamation is void and need to putback the TINGS back to active.  
>
>since involuntary dissolutions is similar to voluntary dissolutions its not needed

argush3 (approved):
Approved assuming the conflicts being flagged are resolved.  

I think we can merge and sort out the dependency issues as the updates for GCP pub/sub pay integration are sorted out

vysakh-menon-aot (approved):

PR #2751, opened 2024-06-11
21475 Dissolutions job - Update stage 1 & stage 2 processes

chenhongjing:
*Issue #:* /bcgov/entity#21475

*Description of changes:*
- Updated stage 1 & stage 2 processes for dissolutions job.
Now `batch_processing.trigger_date` is set in stage 1, and stage 2 & 3 can use this to check whether the business can be processed.
- Fix unit tests

### DIFF
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -98,15 +98,8 @@ def initiate_dissolution_process(app: Flask):  # pylint: disable=redefined-outer
             return
 
         # get stage_1 & stage_2 delay configs
-        if (stage_1_delay_config := app.config.get('STAGE_1_DELAY', 0)):
-            stage_1_delay = timedelta(days=int(stage_1_delay_config))
-        else:
-            stage_1_delay = timedelta(days=0)
-
-        if (stage_2_delay_config := app.config.get('STAGE_2_DELAY', 0)):
-            stage_2_delay = timedelta(days=int(stage_2_delay_config))
-        else:
-            stage_2_delay = timedelta(days=0)
+        stage_1_delay = timedelta(days=app.config.get('STAGE_1_DELAY'))
+        stage_2_delay = timedelta(days=app.config.get('STAGE_2_DELAY'))
 
         # create new entry in batches table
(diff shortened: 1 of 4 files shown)


jobs/involuntary-dissolutions/involuntary_dissolutions.py
```diff
@@ -77,8 +77,7 @@ def initiate_dissolution_process(app: Flask):  # pylint: disable=redefined-outer
     """Initiate dissolution process for new businesses that meet dissolution criteria."""
     try:
         # check if batch has already run today
-        tz = pytz.timezone('US/Pacific')
-        today_date = tz.localize(datetime.today()).date()
+        today_date = datetime.utcnow().date()
```
chenhongjing, line 80: Think it's better to use UTC time instead. 💭

jobs/involuntary-dissolutions/involuntary_dissolutions.py
```diff
@@ -98,11 +97,22 @@ def initiate_dissolution_process(app: Flask):  # pylint: disable=redefined-outer
...
+        # get stage_1 & stage_2 delay configs
+        if (stage_1_delay_config := app.config.get('STAGE_1_DELAY', 0)):
+            stage_1_delay = timedelta(days=int(stage_1_delay_config))
+        else:
+            stage_1_delay = timedelta(days=0)
+
+        if (stage_2_delay_config := app.config.get('STAGE_2_DELAY', 0)):
+            stage_2_delay = timedelta(days=int(stage_2_delay_config))
+        else:
+            stage_2_delay = timedelta(days=0)
```
argush3, line 109: can we update config.py so our delay config entries are like the following?

https://github.com/bcgov/lear/blob/f61c84b50887c3c5e894169743ee0cf3bfdfdde7/queue_services/entity-bn/src/entity_bn/config.py#L78.  I think we can include the default fallback values in code too just in case it doesn't pick it up from 1pass for whatever reason.

Doing this way, we also won't need to provide the fallback outside of config.py.

https://github.com/bcgov/lear/blob/c5d21d3900c1100b35799314514d589d6ef7704c/queue_services/entity-bn/src/entity_bn/bn_processors/change_of_registration.py#L61

argush3 (approved):

leodube-aot (approved):

PR #2761, opened 2024-06-17
21737 - Create alembic script & model for furnishings table

chenhongjing:
*Issue #:* /bcgov/entity#21737

*Description of changes:*
- Added alembic script to create furnishings table
- Added data model for furnishing
- Added unit tests
- Misc updates for others

*db after upgrading*
image

### DIFF
--- a/legal-api/migrations/versions/01b28a2bb730_furnishings.py
+++ b/legal-api/migrations/versions/01b28a2bb730_furnishings.py
@@ -1,0 +1,63 @@
+"""furnishings
+
+Revision ID: 01b28a2bb730
+Revises: 3083d361616e
+Create Date: 2024-06-12 20:10:44.723041
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+# revision identifiers, used by Alembic.
+revision = '01b28a2bb730'
+down_revision = '3083d361616e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table('furnishings',
+        sa.Column('id', sa.Integer(), nullable=False),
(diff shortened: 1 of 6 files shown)


legal-api/migrations/versions/01b28a2bb730_furnishings.py
```diff
@@ -0,0 +1,63 @@
...
+            'INTENT_TO_DISSOLVE',
+            'INTENT_TO_DISSOLVE_XPRO',
+            'CORP_DISSOLVED',
+            'CORP_DISSOLVED_XPRO',
+            name='furnishing_name'
+            ),
+            nullable=False
+        ),
+        sa.Column('batch_id', sa.Integer(), nullable=False),
+        sa.Column('grouping_identifier', sa.Integer(), nullable=True),
```
argush3, line 38: Can we add a sequence that we can use for this field when it's required?

Thinking something like this.  Can start at 1.

https://github.com/bcgov/lear/blob/c8572bcc07bcbe8c1be41b54dd080463a43ec96d/legal-api/migrations/versions/3d8eb786f4c2_add_coop_sequence.py#L19-L24

legal-api/src/legal_api/models/address.py
```diff
@@ -32,7 +32,8 @@ class Address(db.Model):  # pylint: disable=too-many-instance-attributes
 
     MAILING = 'mailing'
     DELIVERY = 'delivery'
-    ADDRESS_TYPES = [MAILING, DELIVERY]
+    FURNISHING = 'furnishing'
+    ADDRESS_TYPES = [MAILING, DELIVERY, FURNISHING]
```
argush3, line 36: I know i have this in the design but I'm actually still thinking about how we will model the address used at the time the dissolution letters need to go out.

Can we leave this out for now?

argush3:
Looks good.  Just the few comments.

leodube-aot (approved):
LGTM!

argush3 (approved):

PR #2772, opened 2024-06-19
Add Reset by event ID endpoint for colin api requiring colin svc id

BrandonSharratt:
*Issue #:* /bcgov/entity###

*Description of changes:*
Add Reset by event id endpoint to colin api. Requires the api service id. Currently only resets events caused by Business AR api calls (business filings)


colin-api/src/colin_api/models/reset.py
```diff
@@ -239,3 +239,49 @@ def reset_filings(cls, start_date: str = None, end_date: str = None, identifiers
...
+            if event_ids:
+                # setup db connection
+                con = DB.connection
+                con.begin()
+                cursor = con.cursor()
+
+                # reset data in oracle for events
+                # The commented out events do not seem to happen for AR so they are commented out.
+                new_corps = cls._get_incorporations_by_event(cursor, event_ids)
+                #Party.reset_dirs_by_events(cursor=cursor, event_ids=event_ids)
```
argush3, line 269: So there is no need to reset party info?

From an AR for a BC company I pulled up in COLIN, I can see that there are `CORP_PARTY` entries

image

colin-api/src/colin_api/models/reset.py
```diff
@@ -239,3 +239,49 @@ def reset_filings(cls, start_date: str = None, end_date: str = None, identifiers
...
+                #Office.reset_offices_by_events(cursor=cursor, event_ids=event_ids)
+                #Business.reset_corp_states(cursor=cursor, event_ids=annual_report_events)
+                Business.reset_corporations(cursor=cursor, event_info=events_info, event_ids=event_ids)
+                #ShareObject.delete_shares(cursor, event_ids)
+                cls._delete_filing_user(cursor=cursor, event_ids=event_ids)
+                #cls._delete_ledger_text(cursor=cursor, event_ids=event_ids)
+                #cls._delete_corp_name(cursor=cursor, event_ids=list(new_corps.values()))
+                cls._delete_corp_state(cursor=cursor, corp_nums=list(new_corps.keys()))
```
argush3, line 277: I'm not entirely sure about this but I'm assuming there is a need to delete all corp state entries that exists for corps that had filed an AR from BAR?

Just confirming that this is needed.

colin-api/src/colin_api/models/reset.py
```diff
@@ -239,3 +239,49 @@ def reset_filings(cls, start_date: str = None, end_date: str = None, identifiers
             if con:
                 con.rollback()
             raise err
+
+    @classmethod
+    def reset_filings_by_event(cls, event_ids: list = []):
```
argush3, line 244: Besides the one entry for corp_party(`Party.reset_dirs_by_events(cursor=cursor, event_ids=event_ids`) that I was wondering about, I think all the other commented out reset calls are not relevant to ARs from what I can see.

Can we remove these lines from the code?

colin-api/src/colin_api/resources/reset.py
```diff
@@ -55,3 +56,41 @@ def post():
...
+        """Reset filing(s) based on the provided event_id, or array of event_ids.
+        This is only tested to work on Annual Reports, ymmv"""
+        try:
+            
+            event_ids = API.payload.get('event_ids', None)
+
+            for event in event_ids:
+                Reset.reset_filings_by_event(
+                  event_ids=event_ids
+                )
```
argush3, line 89: is this for loop needed?  

It looks like ` Reset.reset_filings_by_event(event_ids=event_ids)` can handle `event_ids` as a list already?

argush3 (commented):
Think this is pretty close.

Just a few comments.

colin-api/src/colin_api/models/reset.py
BrandonSharratt: I don't think it's needed from the filing I did no event id was logged in BAR that corresponded to a CORP_PARTY entry in COLIN. That said because of that ther eis no harm in adding it so I'll add it in

colin-api/src/colin_api/models/reset.py
BrandonSharratt: That one does indeed generate a state change, as they get marked as having filed their AR so without that it'll continue to think it was filed.

colin-api/src/colin_api/models/reset.py
BrandonSharratt: Sure

colin-api/src/colin_api/resources/reset.py
BrandonSharratt: Good catch

colin-api/src/colin_api/models/reset.py
argush3: ok, resolving.

argush3 (approved):

severinbeauvais:
Is there a ticket for this change? If so, please tag it to [Entities Sprint 22.1a Release](https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/reports/release?release=Z2lkOi8vcmFwdG9yL1JlbGVhc2UvMTAzMDc4). Thanks.

PR #2777, opened 2024-06-21
21769 - Furnishings job - stage 1 (email flow)

chenhongjing:
*Issue #:* /bcgov/entity#21769

*Description of changes:*
- Implemented stage 1 logic (email flow only)
- Added unit tests
- Updated configs
- Updated dependency requirements (nats related)
- Updated furnishing code model

*Some local testing*
- New furnishing entry is created and email message that emailer received is shown as follows.


chenhongjing:
I've changed the code model of `BatchProcessing` in this PR. Though the job CI testing failed because it still refers to the old model, it succeeded at my local env.

image

jobs/furnishings/tests/unit/test_furnish.py
```diff
@@ -0,0 +1,93 @@
...
+            if test_name == 'NO_EMAIL':
+                assert email is None
+            else:
+                assert email == 'test@no-reply.com'
+
+
+@pytest.mark.asyncio
+async def test_stage_1_process_first_notification_email(app, session):
+    """Assert that email furnishing entry is created correctly."""
+    business = factory_business(identifier='BC1234567')
```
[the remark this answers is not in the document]
chenhongjing: The code of transition filing check in dissolution doesn't support Xpros for now. We'll leave it until we need to make changes. And I added the other test cases.

leodube-aot (approved):

jobs/furnishings/furnish.py
```diff
@@ -65,11 +76,190 @@ def shell_context():
...
+                    _, eligible_details = InvoluntaryDissolutionService.check_business_eligibility(
+                        batch_processing.business_identifier, False
+                        )
+                    if eligible_details:
+                        new_furnishing = create_new_furnishing(
+                            batch_processing,
+                            eligible_details,
+                            Furnishing.FurnishingType.EMAIL,
+                            email
+                            )
```
argush3, line 192: I thought about the design a bit more and I think for every all furnishing entries of a given furnishing type should have a `grouping_identifier` assigned.  

Can you generate one and pass it into each `create_new_furnishing` call here?

jobs/furnishings/tests/unit/test_furnish.py
```diff
@@ -0,0 +1,119 @@
...
+        business_id=business.id,
+        identifier=business.identifier,
+    )
+    if 'TRANSITION' in test_name:
+        factory_completed_filing(business, RESTORATION_FILING, filing_type='restoration')
+
+    qsm = MagicMock()
+    with patch('furnish.get_email_address_from_auth', return_value='test@no-reply.com'):
+        with patch('furnish.send_email', return_value=None):
+            await stage_1_process(app, qsm)
```
argush3, line 111: can we also add an assertion to check that `furnish.send_email`?

jobs/furnishings/tests/unit/test_furnish.py
```diff
@@ -0,0 +1,119 @@
...
+            Business.LegalTypes.COMP.value,
+            Furnishing.FurnishingName.DISSOLUTION_COMMENCEMENT_NO_TR
+        ),
+        (
+            'XP_AR_OVERDUE',
+            Business.LegalTypes.EXTRA_PRO_A.value,
+            Furnishing.FurnishingName.DISSOLUTION_COMMENCEMENT_NO_AR_XPRO
+        )
+    ]
+)
```
argush3, line 95: can we add a test case where if there is no email, no email furnishing entry is created and no call to `furnish.send_email` is made?

argush3 (commented):
Looks good!

Just a few comments.

jobs/furnishings/src/furnishings/stage_processors/stage_one.py
```diff
@@ -0,0 +1,187 @@
...
+
+        grouping_identifier = self._get_grouping_identifier(furnishing_type)
+
+        new_furnishing = Furnishing(
+            furnishing_type=furnishing_type,
+            furnishing_name=furnishing_name,
+            batch_id=batch_processing.batch_id,
+            business_id=batch_processing.business_id,
+            business_identifier=batch_processing.business_identifier,
+            processed_date=datetime.utcnow(),
```
argush3, line 101: I was thinking to make this `null` and the emailer can populate it when it processes it - whether it fails or succeeds in calling the notify api

argush3 (approved):
Thanks for the refactor!  

This should help with keeping things clean as we probably still have lots of code to add to stage 1-3.

PR #2787, opened 2024-06-25
21506 Legal API - Prioritize businesses eligible for involuntary dissolution by transition filing

AimeeGao:
*Issue #:* /bcgov/entity#21506

*Description of changes:*

- Added sorting logic to prioritize businesses by "transition filing not filed" criterion first, followed by the oldest restoration filing date in ascending order.
- Added and updated unit tests to verify the new sorting logic.

image

### DIFF
--- a/legal-api/src/legal_api/services/involuntary_dissolution.py
+++ b/legal-api/src/legal_api/services/involuntary_dissolution.py
@@ -149,7 +149,8 @@ class InvoluntaryDissolutionService():
                     _is_limited_restored(),
                     _is_xpro_from_nwpta()
                 )
-            )
+            ).\
+            order_by(no_transition_filed_after_restoration.desc())
 
         return query
 
--- a/legal-api/src/legal_api/services/involuntary_dissolution.py
+++ b/legal-api/src/legal_api/services/involuntary_dissolution.py
@@ -16,7 +16,7 @@
 from dataclasses import dataclass
 from typing import Final, Tuple
 
-from sqlalchemy import and_, exists, func, not_, or_, text
+from sqlalchemy import and_, exists, func, not_, or_, text, select
 from sqlalchemy.orm import aliased
 
 from legal_api.models import Batch, BatchProcessing, Business, Filing, db
@@ -118,9 +118,9 @@ class InvoluntaryDissolutionService():
                 Batch.batch_type == Batch.BatchType.INVOLUNTARY_DISSOLUTION
             )
         )
-        specific_filing_overdue = _has_specific_filing_overdue()
-        no_transition_filed_after_restoration = _has_no_transition_filed_after_restoration()
-        oldest_restoration_filing_date = _get_oldest_restoration_filing_date_subquery()
+        specific_filing_overdue = _has_specific_filing_overdue() < func.timezone('UTC', func.now())
+        no_transition_filed_after_restoration = func.coalesce((_has_no_transition_filed_after_restoration()
+                                                               <= func.timezone('UTC', func.now())), False)
 
         query = db.session.query(
             Business,
@@ -153,7 +153,9 @@ class InvoluntaryDissolutionService():
             ).\
             order_by(
                 no_transition_filed_after_restoration.desc(),
-                oldest_restoration_filing_date.asc(),
+                _has_no_transition_filed_after_restoration().asc(),
+                specific_filing_overdue.desc(),
+                _has_specific_filing_overdue().asc()
             )
 
         return query
@@ -182,7 +184,7 @@ def _has_specific_filing_overdue():
 
     latest_date_cutoff = latest_date + text("""INTERVAL '26 MONTHS'""")
 
-    return latest_date_cutoff < func.timezone('UTC', func.now())
+    return latest_date_cutoff
 
 
 def _has_no_transition_filed_after_restoration():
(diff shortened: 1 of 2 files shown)


legal-api/tests/unit/services/test_involuntary_dissolution.py
```diff
@@ -308,3 +308,39 @@ def test_exclude_admin_frozen_businesses(session, test_name, admin_freeze, eligi
 
     check_eligibility, eligibility_details = InvoluntaryDissolutionService.check_business_eligibility(identifier)
     assert check_eligibility == eligible
+
+
+@pytest.mark.parametrize(
```
eason-pan-bc, line 313: Given that there is only one test case, is parameterizing this test necessary? Just a thought!

legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -141,6 +142,10 @@ def _get_businesses_eligible_query(exclude_in_dissolution=True):
                     _is_limited_restored(),
                     _is_xpro_from_nwpta()
                 )
+            ).\
+            order_by(
+                no_transition_filed_after_restoration.desc(),
+                oldest_restoration_filing_date.asc(),
```
eason-pan-bc, line 148: Just wondering, what will happen when there are businesses have the same effective date? then by filing id or random?

legal-api/tests/unit/services/test_involuntary_dissolution.py
AimeeGao: Thanks for the feedback. Mark this PR as a draft to make further updates.

AimeeGao:
Mark this PR as a draft to make further updates. Not ready for review now.

legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -141,6 +142,12 @@ def _get_businesses_eligible_query(exclude_in_dissolution=True):
                     _is_limited_restored(),
                     _is_xpro_from_nwpta()
                 )
+            ).\
+            order_by(
+                no_transition_filed_after_restoration.desc(),
+                _has_no_transition_filed_after_restoration().asc(),
```
eason-pan-bc, line 148: how do we handle the situation when there are multiple businesses have the same effective date? e.g. 
image

Let's say, we are going to select 100 businesses to put into Involuntary Dissolution process, and the 100th - 101st businesses have the same effective date. Then which one should we pick? Or this doesn't actually matter?

Other than these, LGTM

AimeeGao: Thank you for bringing this up, Eason. 

Right now, we don't have a specific requirement for this case, maybe the order between them doesn't matter.

legal-api/tests/unit/services/test_involuntary_dissolution.py
```diff
@@ -308,3 +308,67 @@ def test_exclude_admin_frozen_businesses(session, test_name, admin_freeze, eligi
...
+    # create business that is neither ar_overdue nor transition_overdue
+    business_no_overdue = factory_business(identifier='BC1122334', entity_type=Business.LegalTypes.COMP.value)
+    factory_completed_filing(business_no_overdue, RESTORATION_FILING, filing_type='restoration')
+    factory_completed_filing(business_no_overdue, TRANSITION_FILING_TEMPLATE, filing_type='transition')
+    business_no_overdue.last_ar_date = datetime.utcnow()
+    business_no_overdue.save()
+
+    result = InvoluntaryDissolutionService._get_businesses_eligible_query().all()
+    if exclude:
+        assert not result
```
chenhongjing, line 370: Is `exclude` a necessary parameter in this test?

chenhongjing (commented):
Looks good!
Just a comment. 👀

chenhongjing (approved):
LGTM 😸

eason-pan-bc (approved):
LGTM 👍

argush3:
can you rebase your latest changes?  

Kevin made some recent updates that affect this code I think.

AimeeGao:
> can you rebase your latest changes?
> 
> Kevin made some recent updates that affect this code I think.

Sure.

legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -149,6 +150,12 @@ def _get_businesses_eligible_query(eligibility_filters: EligibilityFilters = Eli
...
+                specific_filing_overdue.desc(),
+                _has_specific_filing_overdue().asc()
```
argush3, line 158: does this `order_by` code break if any of the clauses returns `None`?

AimeeGao: no_transition_filed_after_restoration and specific_filing_overdue will only return true or false. 

_has_no_transition_filed_after_restoration() might return None when we can't find the expected result. For this case, SQLAlchemy will handle None values by putting them at the end of the result (cause we use ascending order).

The result after ordering would be:
 1. rows sorted based on no_transition_filed_after_restoration. True values appear before False values(In descending order).
 2. The rows with the same no_transition_filed_after_restoration value would be prioritized in ascending order with the Non-None value, and None values will be placed after them.
 3. specific_filing_overdue and _has_specific_filing_overdue have the same ordering logic.
image

argush3: great, thanks for the explanation!

argush3 (approved):

PR #2799, opened 2024-07-02
21777 add exclude future effective filing flag

kzdev420:
*Issue #:* /bcgov/entity#21777

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/services/involuntary_dissolution.py
+++ b/legal-api/src/legal_api/services/involuntary_dissolution.py
@@ -46,7 +46,7 @@ class InvoluntaryDissolutionService():
         transition_overdue: bool
 
     @dataclass
-    class ExcludeDetails:
+    class ExclusionSettings:
         """Details about the exclude of a business for dissolution."""
 
         exclude_in_dissolution: bool = True
@@ -54,7 +54,7 @@ class InvoluntaryDissolutionService():
 
     @classmethod
     def check_business_eligibility(
-        cls, identifier: str, exclude_details: ExcludeDetails
+        cls, identifier: str, exclude_settings: ExclusionSettings = ExclusionSettings()
     ) -> Tuple[bool, EligibilityDetails]:
         """Return true if the business with provided identifier is eligible for dissolution.
 
@@ -63,8 +63,8 @@ class InvoluntaryDissolutionService():
             eligibility_details (EligibilityDetails): Details regarding eligibility.
         """
         query = cls._get_businesses_eligible_query(
-            exclude_details.exclude_in_dissolution, exclude_details.exclude_future_effective_filing
-            ).filter(Business.identifier == identifier)
+            exclude_settings.exclude_in_dissolution, exclude_settings.exclude_future_effective_filing
+        ).filter(Business.identifier == identifier)
         result = query.one_or_none()
 
         if result is None:
--- a/legal-api/src/legal_api/services/involuntary_dissolution.py
+++ b/legal-api/src/legal_api/services/involuntary_dissolution.py
@@ -47,7 +47,7 @@ class InvoluntaryDissolutionService():
 
     @classmethod
     def check_business_eligibility(
-        cls, identifier: str, exclude_in_dissolution=True
+        cls, identifier: str, exclude_in_dissolution=True, exclude_future_effective_filing=False
     ) -> Tuple[bool, EligibilityDetails]:
         """Return true if the business with provided identifier is eligible for dissolution.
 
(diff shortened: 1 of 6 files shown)


legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -47,15 +47,15 @@ class EligibilityDetails:
 
     @classmethod
     def check_business_eligibility(
-        cls, identifier: str, exclude_in_dissolution=True
+        cls, identifier: str, exclude_in_dissolution=True, exclude_future_effective_filing=False
```
argush3, line 50: we had talked about things starting to get messy with the exclude parameters and to use an enum to encapsulate `exclude_in_dissolution` as well as `exclude_future_effective_filing`.  

Can we do that?

argush3:
Can we add relevant tests to verify that dissolution related warnings are able to return even when there is a future effective filing?  

We'll want to test change of address for sure as this was what broke the warnings code.

kzdev420:
> Can we add relevant tests to verify that dissolution related warnings are able to return even when there is a future effective filing?
> 
> We'll want to test change of address for sure as this was what broke the warnings code.

Yes, I will add the test too

legal-api/src/legal_api/services/involuntary_dissolution.py
kzdev420: I did

legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -45,18 +45,26 @@ class EligibilityDetails:
...
+    class ExclusionSettings:
+        """Details about the exclude of a business for dissolution."""
+
+        exclude_in_dissolution: bool = True
+        exclude_future_effective_filing: bool = False
+
     @classmethod
     def check_business_eligibility(
-        cls, identifier: str, exclude_in_dissolution=True
+        cls, identifier: str, exclude_settings: ExclusionSettings = ExclusionSettings()
```
chenhongjing, line 57: This change also breaks [furnishings job](https://github.com/bcgov/lear/blob/main/jobs/furnishings/src/furnishings/stage_processors/stage_one.py#L55). Can you also modify the job if it's convenient for you?

legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -94,7 +102,7 @@ def get_in_dissolution_batch_processing(business_id: int):
             one_or_none()
 
     @staticmethod
-    def _get_businesses_eligible_query(exclude_in_dissolution=True):
+    def _get_businesses_eligible_query(exclude_in_dissolution=True, exclude_future_effective_filing=False):
         """Return SQLAlchemy clause for fetching businesses eligible for involuntary dissolution.
 
         Args:
```
eason-pan-bc, line 108: Maybe consider updating the Args part of the docString for the newly added `exclude_future_effective_filing` argument?

legal-api/src/legal_api/services/involuntary_dissolution.py
kzdev420: Yes, It will break. If this change is good, then, I will update other services too

argush3: The updates for other BE services can be done in a separate PR so we can verify if tests pass.

legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -45,18 +45,26 @@ class EligibilityDetails:
         ar_overdue: bool
         transition_overdue: bool
 
+    @dataclass
+    class ExclusionSettings:
```
argush3, line 49: Can we rename this to `EligibilityFilters`?

legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -45,18 +45,26 @@ class EligibilityDetails:
...
 
         Returns:
             eligible (bool): True if the business is eligible for dissolution.
             eligibility_details (EligibilityDetails): Details regarding eligibility.
         """
-        query = cls._get_businesses_eligible_query(exclude_in_dissolution).\
-            filter(Business.identifier == identifier)
+        query = cls._get_businesses_eligible_query(
+            exclude_settings.exclude_in_dissolution, exclude_settings.exclude_future_effective_filing
+        ).filter(Business.identifier == identifier)
```
argush3, line 67: can we just pass the new dataclass(`EligibilityFilters`) for all functions?

legal-api/src/legal_api/services/involuntary_dissolution.py
```diff
@@ -94,7 +102,7 @@ def get_in_dissolution_batch_processing(business_id: int):
             one_or_none()
 
     @staticmethod
-    def _get_businesses_eligible_query(exclude_in_dissolution=True):
+    def _get_businesses_eligible_query(exclude_in_dissolution=True, exclude_future_effective_filing=False):
```
argush3, line 105: can we just use the new dataclass(`EligibilityFilters`)?

kzdev420: Updated

legal-api/src/legal_api/services/involuntary_dissolution.py
kzdev420: Updated

argush3 (approved):

chenhongjing (approved):

PR #2801, opened 2024-07-03
22033 - Fixed performance of get alternate names + misc. changes

JazzarKarim:
*Issue #:* /bcgov/entity#22033

*Description of changes:*
- Optimized the get_alternate_names function
- Moved the alternateNames array back to big JSON (removed it from slim json)
- Only get it in the slim json in search endpoint for firms
- fixed unit test

### Before these changes:
beofre performance fix

### After these changes:
after performance fix

### DIFF
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -490,8 +490,6 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
 
         None fields are not included.
         """
-        from legal_api.services import flags  # pylint: disable=import-outside-toplevel
-
         slim_json = self._slim_json()
         if slim:
             return slim_json
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -298,10 +298,7 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
 
         Legal Name Easy Fix
         """
-        from legal_api.services import flags  # pylint: disable=import-outside-toplevel
-
-        flag_on = flags.is_on('enable-legal-name-fix')
-        if self.is_firm and flag_on:
+        if self.is_firm:
             sort_name = func.trim(
                 func.coalesce(Party.organization_name, '') +
                 func.coalesce(Party.last_name + ' ', '') +
@@ -524,11 +521,9 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
             ).astimezone(timezone.utc).isoformat(),
             'noDissolution': self.no_dissolution,
             'associationType': self.association_type,
-            'allowedActions': self.allowable_actions
+            'allowedActions': self.allowable_actions,
+            'alternateNames': self.get_alternate_names()
         }
-
-        if flags.is_on('enable-legal-name-fix'):
-            d['alternateNames'] = self.get_alternate_names()
 
         self._extend_json(d)
 
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -522,7 +522,6 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
             'allowedActions': self.allowable_actions,
             'alternateNames': self.get_alternate_names()
         }
-
         self._extend_json(d)
 
         return d
--- a/legal-api/src/legal_api/models/business.py
+++ b/legal-api/src/legal_api/models/business.py
@@ -493,6 +493,8 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
 
         None fields are not included.
         """
+        from legal_api.services import flags  # pylint: disable=import-outside-toplevel
+
         slim_json = self._slim_json()
         if slim:
             return slim_json
@@ -524,14 +526,16 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
             'associationType': self.association_type,
             'allowedActions': self.allowable_actions
         }
+
+        if flags.is_on('enable-legal-name-fix'):
+            d['alternateNames'] = self.get_alternate_names()
+
         self._extend_json(d)
 
         return d
 
     def _slim_json(self):
         """Return a smaller/faster version of the business json."""
-        from legal_api.services import flags  # pylint: disable=import-outside-toplevel
-
         d = {
             'adminFreeze': self.admin_freeze or False,
             'goodStanding': self.good_standing,
@@ -541,9 +545,6 @@ class Business(db.Model):  # pylint: disable=too-many-instance-attributes,disabl
             'legalType': self.legal_type,
             'state': self.state.name if self.state else Business.State.ACTIVE.name
         }
-
-        if flags.is_on('enable-legal-name-fix'):
-            d['alternateNames'] = self.get_alternate_names()
 
         if self.tax_id:
             d['taxId'] = self.tax_id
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/models/business.py
```diff
@@ -524,14 +526,16 @@ def json(self, slim=False):
             'associationType': self.association_type,
             'allowedActions': self.allowable_actions
         }
+
+        if flags.is_on('enable-legal-name-fix'):
```
JazzarKarim, line 530: Moved this back to here in the big JSON. The only use I believe to have it in slim JSON is fore the search endpoint.

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -164,7 +164,13 @@ def search_businesses():
         draft_query = db.session.query(Filing).filter(
             and_(Filing.temp_reg.in_(identifiers), Filing.business_id.is_(None)))
 
-        bus_results = [x.json(slim=True) for x in bus_query.all()]
+        bus_results = []
+        for business in bus_query.all():
+            business_json = business.json(slim=True)
+            # add alternateNames array to slim json only to firms
+            if business.legal_type in (Business.LegalTypes.SOLE_PROP, Business.LegalTypes.PARTNERSHIP):
```
JazzarKarim, line 171: So, after investigating and looking at the code in AUTH WEB, I believe we don't need to have the alternateNames array for everything. We only show the alternate name in MBR for firms. Otherwise, we show legal name. This also helped boost performance by a lot.

legal-api/src/legal_api/models/business.py
```diff
@@ -694,65 +695,68 @@ def get_alternate_names(self) -> dict:
...
         alias_version = version_class(Alias)
-        aliases = db.session.query(alias_version). \
-            filter(alias_version.id.in_([a.id for a in self.aliases]),
-                   alias_version.end_transaction_id == None  # noqa: E711; pylint: disable=singleton-comparison
-                   ).all()
-        for alias in aliases:
-            filing = db.session.query(Filing). \
-                filter(Filing.transaction_id == alias.transaction_id
-                       ).one_or_none()
+        filing_alias = aliased(Filing)
```
JazzarKarim, line 700: This function was changed using a combination of my brain and AI's brain 😆 

The idea is, we're kind of doing a single query now and using joins. So, it's decreasing the amount of roundtrips.

legal-api/src/legal_api/resources/v2/business/business.py
argush3: You'll need to verify with @kialj876 whether he uses `alternateNames` at all for non-firms for the search endpoint.

If we remove this, it will be hard to maintain as it is a weird edge case and I believe the search api users are expecting alternate names for non-firms as well where it's applicable.

legal-api/src/legal_api/models/business.py
```diff
@@ -694,65 +695,68 @@ def get_alternate_names(self) -> dict:
...
-                        {
-                            'entityType': sole_prop.legal_type,
-                            'identifier': sole_prop.identifier,
-                            'name': sole_prop.legal_name,
-                            'registeredDate': sole_prop.founding_date.isoformat(),
-                            'startDate': start_date,
-                            'type': 'DBA'
-                        }
-                    )
+            proprietors_query = db.session.query(
```
seeker25, line 721: This is much better I think

legal-api/src/legal_api/models/business.py
```diff
@@ -694,65 +695,68 @@ def get_alternate_names(self) -> dict:
...
-        aliases = db.session.query(alias_version). \
-            filter(alias_version.id.in_([a.id for a in self.aliases]),
-                   alias_version.end_transaction_id == None  # noqa: E711; pylint: disable=singleton-comparison
-                   ).all()
-        for alias in aliases:
-            filing = db.session.query(Filing). \
-                filter(Filing.transaction_id == alias.transaction_id
-                       ).one_or_none()
+        filing_alias = aliased(Filing)
+        aliases_query = db.session.query(
```
seeker25, line 701: alot better if it works correctly

legal-api/src/legal_api/resources/v2/business/business.py
JazzarKarim: Very good point, makes a lot of sense.

My thinking was, when we first implemented the easy legal name fix, the alternateNames were in big JSON. It was only when we needed to fix the names in MBR that we decided to move it to slim so that it's fixed here in the search endpoint.

legal-api/src/legal_api/models/business.py
JazzarKarim: Yep, I've tested extensively. I'm getting the exact same results as before 🙏

legal-api/src/legal_api/models/business.py
kialj876: It is needed in the slim json for business search + person search updates (this was the original use case we added the slim json for as the call was timing out for several businesses)

kialj876: also I don't think it needs to check the flag anymore as you have released legal name fixes?

legal-api/src/legal_api/resources/v2/business/business.py
kialj876: I'm not using this endpoint for anything, but I think auth uses it in the proxy call for the 'My Business Registry' table so check with @seeker25

legal-api/src/legal_api/models/business.py
JazzarKarim: Got it, will revert this change back Kial.

Also, yeah I agree. We can probably get rid of the FF.

legal-api/src/legal_api/resources/v2/business/business.py
seeker25: Only usage here:
image

seeker25: Only SP and GP

legal-api/src/legal_api/models/business.py
kialj876: awesome thank you 👍

legal-api/src/legal_api/resources/v2/business/business.py
argush3: ok, for some reason I thought we might be using it outside of MBR.

think we can go with this change.  

Can you add a note the search endpoint to make note that we are not returning alternate_names for non-firms due to perf issues?

legal-api/src/legal_api/models/business.py
JazzarKarim: @kialj876 Kial, from my understanding, you only use on your side the search endpoint as well right? If that's the case, if the alternateNames is only there for firms (SPs and GPs) in slim JSON, does that cause issues on your side?

legal-api/src/legal_api/resources/v2/business/business.py
kialj876: FYI this is what I am talking about: https://github.com/bcgov/sbc-auth/blob/main/auth-api/src/auth_api/resources/v1/org.py#L355 

^ is the call effected by this change and you won't be able to determine the 'operating name' anymore in that table. 
Pretty sure this change would cause the table to display 'Hrvoje..' instead


Screenshot 2024-07-04 at 11 51 22 AM

JazzarKarim: Since this is an SP, the alternateNames array will be there (including the name). So, Auth Web will be able to get it and output it correctly I believe. I can make the call locally and show.

kialj876: oops nvm only sp / gp as travis said -- sorry github didn't update with all your comments for me till after haha

JazzarKarim: All good, no worries 🙏

legal-api/src/legal_api/models/business.py
kialj876: for the flows I am talking about here we use the business endpoint NOT the search endpoint

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -59,8 +59,11 @@ def get_businesses(identifier: str):
 
     # getting all business info is expensive so returning the slim version is desirable for some flows
     # - (i.e. business search update)
+    # need to add the alternateNames array here because it is not a part of slim JSON
     if str(request.args.get('slim', None)).lower() == 'true':
-        return jsonify(business=business.json(slim=True))
+        business_json = business.json(slim=True)
+        business_json['alternateNames'] = business.get_alternate_names()
```
JazzarKarim, line 65: @kialj876 @argush3 @seeker25 Are you guys OK with this being here? This'll keep the flow where we grab the alternateNames when calling the business endpoint. This'll keep the flow working on Kial's side as usual:
slim true call

legal-api/src/legal_api/models/business.py
```diff
@@ -298,10 +298,7 @@ def business_legal_name(self):
 
         Legal Name Easy Fix
         """
-        from legal_api.services import flags  # pylint: disable=import-outside-toplevel
-
-        flag_on = flags.is_on('enable-legal-name-fix')
```
JazzarKarim, line 303: Now that we released the easy legal name fix, it's a good idea to clean up the FF everywhere from the code as per Kial's comment.

legal-api/src/legal_api/resources/v2/business/business.py
seeker25: you moved the alternateNames out of the .json(slim=True)? yup as long as it's included in /search I think it is?

JazzarKarim: Yep, exactly. It's also still included in /search as long as it's a firm (this is what MBR uses).

argush3 (approved):

PR #2805, opened 2024-07-04
API - business allowable actions for comp auths

kialj876:
*Issue #:* /bcgov/entity#22030

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/services/authz.py
+++ b/legal-api/src/legal_api/services/authz.py
@@ -19,7 +19,7 @@ from typing import List
 from urllib.parse import urljoin
 
 import jwt as pyjwt
-from flask import current_app
+from flask import Response, current_app, request
 from flask_jwt_oidc import JwtManager
 from requests import Session, exceptions
 from requests.adapters import HTTPAdapter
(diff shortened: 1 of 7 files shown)


legal-api/src/legal_api/services/authz.py
```diff
@@ -82,33 +116,9 @@ def authorized(  # pylint: disable=too-many-return-statements
         if any(elem in action for elem in staff_only_actions):
             return False
 
-        template_url = current_app.config.get('AUTH_SVC_URL')
-        auth_url = f'{template_url}/entities/{identifier}/authorizations'
-
-        token = jwt.get_token_auth_header()
-        headers = {'Authorization': 'Bearer ' + token}
-        try:
```
kialj876, line 90: moved this into `_call_auth_api` so that it could be reused in `get_account_products`

kialj876:
tests are passing, but codecov is erring due to external issue I think

JazzarKarim (approved):
LGTM 👍

JazzarKarim:
> tests are passing, but codecov is erring due to external issue I think

The codecov issue has been there for a while. It's there for both FE and BE. We fixed it a while back but now the issue is back. No worries, we can ignore it for now.

legal-api/src/legal_api/services/authz.py
```diff
@@ -71,9 +104,10 @@ def authorized(  # pylint: disable=too-many-return-statements
             or jwt.validate_roles([COLIN_SVC_ROLE]):
         return True
 
-    # allow IDIM view access on everything
-    if len(action) == 1 and action[0] == 'view' and jwt.validate_roles([ACCOUNT_IDENTITY]):
-        return True
+    # allow IDIM and Competent Authorities view access on everything
+    if len(action) == 1 and action[0] == 'view':
```
hfekete, line 108: Just a note
You could have put 2nd if expression in the bracket and join it with `and`, if there will be no granularity needed for some other options.

hfekete (approved):
Looks good from my perspective, lets see what comments others have.

argush3 (approved):

chenhongjing (approved):

legal-api/src/legal_api/services/authz.py
kialj876: yeah I just separated them for readability

PR #2840, opened 2024-07-17
21953 Fix issue when auth returns 404 for GET <auth_url>/entities/{identifier} endpoint

argush3:
*Issue #:* /bcgov/entity#21953

*Description of changes:*

* Handles scenario where an auth call to retrieve email info may return a 404.  This shouldn't really happen but for imported BC companies data, there may be entries that haven't been created in auth which leads to auth returning a 404.  For now, we will log this as an error and just return `None` for the contact email so the furnishings job can continue running


kzdev420 (approved):
LGTM

chenhongjing (approved):

leodube-aot (approved):

leodube-aot:
Looks good, just the linting issues

PR #2843, opened 2024-07-18
22285 db update for storing xml payloads

kzdev420:
*Issue #:* /bcgov/entity#22285

*Description of changes:*


legal-api/migrations/versions/b14f77400786_xml_payloads.py
```diff
@@ -0,0 +1,43 @@
...
+    op.create_table('xml_payloads',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('payload', sa.TEXT, nullable=False),    
+    sa.Column('created_date', sa.TIMESTAMP(timezone=True), nullable=False),
+    sa.PrimaryKeyConstraint('id')
+    )
+
+    op.create_table('furnishing_groups',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('xml_payload_id', sa.Integer(), nullable=False),
```
chenhongjing, line 29: Can it be nullable because we might add more links other than `xml_payload_id`? 🤔 💭

argush3: And it's also possible that we only have the `id` value populated

kzdev420: Updated

leodube-aot (approved):

chenhongjing (approved):

legal-api/migrations/versions/b14f77400786_xml_payloads.py
```diff
@@ -0,0 +1,43 @@
...
+revision = 'b14f77400786'
+down_revision = 'feb206b2ce65'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table('xml_payloads',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('payload', sa.TEXT, nullable=False),
```
argush3, line 22: I was envisioning us using the postgres xml column type as it may have better performance and features specific to xml.  

Can you check if the following works?  

`sa.Column('payload', sa.XML(), nullable=False),`

kzdev420: But, `sa` doesn't support the XML
Screenshot 2024-07-19 at 1 30 44 PM

argush3: give this a try - `sa.Column('payload', postgresql.XML(), nullable=False),`

I think you'll need to do any import `from sqlalchemy.dialects import postgresql`

legal-api/src/legal_api/models/furnishing.py
```diff
@@ -77,9 +77,12 @@ class FurnishingStatus(BaseEnum):
     # parent keys
     batch_id = db.Column('batch_id', db.Integer, db.ForeignKey('batches.id'), index=True, nullable=False)
     business_id = db.Column('business_id', db.Integer, db.ForeignKey('businesses.id'), index=True, nullable=False)
+    furnishing_group_id = db.Column('furnishing_group_id', db.Integer, db.ForeignKey('furnishing_groups.id'),
+                                    index=True, nullable=False)
```
argush3, line 81: so we don't actually have an index on this column yet but I think it makes sense to.  

Can you add an index for this column in the alembic script?

argush3:
Looks good.  Will approve once my comments are addressed.

legal-api/migrations/versions/b14f77400786_xml_payloads.py
kzdev420: I tried, but got this error
`AttributeError: module 'sqlalchemy.dialects.postgresql' has no attribute 'XML'`

legal-api/src/legal_api/models/furnishing.py
kzdev420: Added

legal-api/migrations/versions/b14f77400786_xml_payloads.py
argush3: I'm pretty sure XML should be supported for postgres.

I may not have given an example that works but can you try figuring how to get XML column type working?

kzdev420: I tried to figure it out. But, didn't find the solution. ☹️

chenhongjing: Postgres supports XML while SQLAlchemy doesn't support it dirrectly. Maybe you can try to create a custom SQLAlchemy type to handle XML. 🤔

argush3: yes, please try Hongjing's suggestion, @kzdev420

kzdev420: I created the XML data type.

Screenshot 2024-07-24 at 8 27 15 AM

legal-api/src/legal_api/utils/custom_db_types.py
argush3, line 1: can we move this file directly under `src/legal_api/models`?

kzdev420: Moved

legal-api/src/legal_api/models/custom_db_types.py
```diff
@@ -0,0 +1,67 @@
+# Copyright © 2020 Province of British Columbia
```
argush3, line 1: 2024

kzdev420: fixed

argush3 (approved):

legal-api/src/legal_api/models/custom_db_types.py
```diff
@@ -0,0 +1,67 @@
...
+        Args:
+            dialect: The database dialect being used.
+            coltype: The column type.
+        Returns:
+            A function that processes the value retrieved from the database. In this case, the function
+            returns the value unchanged, as no special processing is needed.
+        """
+        def process(value):
+            return value
+        return process
```
vysakh-menon-aot, line 67: Just curious, what is the purpose of a user defined column type if there is no action/validation here?

vysakh-menon-aot (approved):

legal-api/src/legal_api/models/custom_db_types.py
argush3: I haven't looked into this but for now I just wanted to XML type created for the column so we can use xml specific functionality in postgres if we ever need it.

But I have a feeling that postgres will still block saving if it is not well formed xml by default.

PR #2846, opened 2024-07-18
22301-update change agm location

eason-pan-bc:
*Issue #:* /bcgov/entity#22301

*Description of changes:*
- added Agm_location_change schema to schemas section under components
- updated the Filing and filing_head schema to match the ones in Business Schemas
- updated the business schema under the components to match what returns in a get business api request
- added request and response example for agm location change filing (corporations only for the sake of scope)
- added 3 pairs of failed examples with error messages
image
image

**Note:**
- You will need the API-Key, account-id and select dev (API SERVER) in order to try
- We are currently just file draft, but also we can file agm location change completed (since it's free), refer to the screenshot below
image
- If we set the draft to true, then we would need to manually delete the draft on staff dashboard UI to file a new one.


docs/business.yaml
```diff
@@ -89,8 +89,8 @@ paths:
       summary: submit new filing
       operationId: post-businesses-identifier-filings
       responses:
-        '200':
-          description: OK
+        '201':
```
AimeeGao, line 92: Besides the 201 response, I am considering whether we should also add other responses like 401, 402, etc. 
My code only covers these two cases right now. Perhaps we could discuss other possible scenarios together and consider how to include corresponding schemas for these potential responses.

AimeeGao (approved):

docs/business.yaml
eason-pan-bc: You mean the other scenarios provided in the Postman collection? I personally don't think it's necessary, otherwise, there is no point to have the Postman collection. But we can double-check with @argush3 and Dave

eason-pan-bc: But you may be right. PPR did put the failed ones in it.

eason-pan-bc: added 3 pairs of failed examples with error messages 😄

chenhongjing (approved):

docs/business.yaml
```diff
@@ -113,6 +164,72 @@ paths:
...
               $ref: '#/components/schemas/Filing'
+            examples:
+              File-agm-location-change-request:
+                summary: '[Success] Submit a Notice of AGM Location Change Filing'
+                value:
+                  filing:
+                    header:
+                      date: '2024-07-18'
+                      name: agmLocationChange
+                      certifiedBy: Sample Cerfitied Person
```
argush3, line 175: typo

docs/business.yaml
```diff
@@ -113,6 +164,72 @@ paths:
...
+                      year: '2024'
+                      reason: API Specs Tests
+              Failed-agm-location-change-missing-reason-request:
+                summary: '[Failed] AGM Location Change - Missing Reason'
+                value:
+                  filing:
+                    header:
+                      date: '2024-07-18'
+                      name: agmLocationChange
+                      certifiedBy: Sample Cerfitied Person
```
argush3, line 208: typo

docs/business.yaml
```diff
@@ -113,6 +164,72 @@ paths:
...
+                      year: '2024'
+                      agmLocation: Victoria, BC, Canada
+              Failed-agm-location-change-missing-year-request:
+                summary: '[Failed] AGM Location Change - Missing Year'
+                value:
+                  filing:
+                    header:
+                      date: '2024-07-18'
+                      name: agmLocationChange
+                      certifiedBy: Sample Cerfitied Person
```
argush3, line 224: typo

argush3 (commented):
looks mostly good. 

just the one main comment about the request examples

eason-pan-bc:
> looks mostly good.
> 
> just the one main comment about the request examples

Thanks for the detailed feedback. The 3 invalid request examples have been removed, and all the typos are fixed now.

argush3 (approved):

PR #2847, opened 2024-07-19
22295-update-consent-to-continue-out-spec 

AimeeGao:
*Issue #:* /bcgov/entity#22295

*Description of changes:*

- Added Consent_continuation_out schema to the schema section under components 
- Added Foreign_jurisdiction schema to the schema section under components
- Added errorResponse schema to the schema section under components
- Added examples of request and response for Consent continuation out filing
- Added API response definitions for HTTP error status codes
- Added new error responses to POST /businesses/{identifier}/filings endpoint

**Request example**
image

**Response example**
image


docs/business.yaml
```diff
@@ -4236,6 +4439,62 @@ components:
       description: The name of the legalName of the document.
       schema:
         type: string
+  responses:
```
AimeeGao, line 4442: The responses code follows the PPR's pattern.

The examples provided are general scenarios. Not sure if we should expand the response examples to reflect specific types of filings. If so, this would involve detailing the possible error responses for different filing scenarios.

docs/business.yaml
```diff
@@ -105,6 +105,49 @@ paths:
...
+                        inColinOnly: false
+                        isCorrected: false
+                        isCorrectionPending: false
+                        isPaymentActionRequired: true
+                        name: consentContinuationOut
+                        paymentAccount: '3445'
+                        paymentStatusCode: 'CREATED'
+                        paymentToken: '38416'
+                        status: 'PENDING'
+                        submitter: 'api-key-account-3445-sandbox'
```
AimeeGao, line 146: Only cover 2 error responses now.
Plan to update with additional error responses for other potential scenarios.

docs/business.yaml
```diff
@@ -105,6 +105,49 @@ paths:
...
+        '400':
```
AimeeGao, line 147: Only cover 2 error responses now.
Plan to update with additional error responses for other potential scenarios.

eason-pan-bc: as discussed with Argus, maybe we want to add those scenarios covered in the Postman collections.

eason-pan-bc: In that case, maybe we can consider putting precise error messages for the bad calls?

AimeeGao: For sure, I'm thinking of expanding the corresponding example case under the corresponding error responses schema, something like this.
image
What do you think of this?

eason-pan-bc: I'd say, put them directly under each paths, like different request examples. Since we may have different response messages

docs/business.yaml
```diff
@@ -144,6 +144,56 @@ paths:
                         paymentToken: '38416'
                         status: 'PENDING'
                         submitter: 'api-key-account-3445-sandbox'
+        '400':
+          description: The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing).
```
AimeeGao, line 148: Add additional error responses for other potential scenarios.
400:
image

image

401:
image

422:
image

chenhongjing (approved):

eason-pan-bc (approved):
LGTM 😄

docs/business.yaml
```diff
@@ -136,6 +136,84 @@ paths:
...
+                        inColinOnly: false
+                        isCorrected: false
+                        isCorrectionPending: false
+                        isPaymentActionRequired: true
+                        name: consentContinuationOut
+                        paymentAccount: '3445'
+                        paymentStatusCode: 'CREATED'
+                        paymentToken: '38416'
+                        status: 'PENDING'
+                        submitter: 'api-key-account-3445-sandbox'
```
argush3, line 176: Please update to use "mocked submitter". 

As per our conversation, Eason will update any other references to `submitter: 'api-key-account-3445-sandbox'` in his upcoming work.

AimeeGao: Updated

argush3 (approved):

PR #2857, opened 2024-07-24
22294 update agm extension spec

AimeeGao:
*Issue #:* /bcgov/entity#22294

*Description of changes:*
- Added Agm_extension schema to the schema section under components
- Added examples of request and response for AGM extension filing
- Deleted unused schema party - copy and Suggestion_item - copy to ensure successful import into the Swagger Editor for QA

**Request**
image

**Response**
image

image

image

### DIFF
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -1888,77 +1888,15 @@ components:
               format: date
               description: Expiration date for approved extension.
               example: '2024-07-10'
-            currentDate:
-              type: string
-              format: date
-              description: The current date at the time of the AGM extension request.
-              example: '2023-07-21'
-            alreadyExtended:
-              type: boolean
-              description: Indicates if the AGM has already been extended.
-              example: false
-            requestExpired:
-              type: boolean
-              description: Indicates if the request for extension has expired.
-              example: false
-            isGoodStanding:
-              type: boolean
-              description: Indicates if the company is in good standing.
-              example: true
-            incorporationDate:
-              type: string
-              format: date-time
-              description: The date and time when the company was incorporated.
-              example: '2024-06-20T17:31:58.000Z'
-            agmYear:
-              type: string
-              description: The year of the AGM for which the extension is being requested. Usually same as 'year'.
-              example: '2024'
-            prevAgmDate:
-              type: string
-              format: date
-              nullable: true
-              description: The date of the previous AGM, if applicable.
-              example: null
-            isPrevExtension:
-              type: boolean
-              description: Indicates if there was a previous extension for this AGM.
-              example: false
-            prevExpiryDate:
-              type: string
-              format: date
-              nullable: true
-              description: The expiry date of the previous extension, if applicable.
-              example: null
-            agmDueDate:
-              type: string
-              format: date
-              description: The due date for the AGM after the extension is applied.
-              example: '2026-06-20'
-            isEligible:
-              type: boolean
-              description: Indicates if the company is eligible for an AGM extension.
-              example: true
           x-examples:
             Example 1:
               agmExtension:
                 year: '2024'
                 isFirstAgm: true
                 extReqForAgmYear: false
-                currentDate: '2024-07-10'
                 expireDateApprovedExt: '2026-06-20'
-                incorporationDate: '2024-01-15T00:00:00.000Z'
                 totalApprovedExt: 6
                 extensionDuration: 6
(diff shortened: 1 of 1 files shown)


docs/business.yaml
```diff
@@ -1417,6 +1514,124 @@ components:
...
+                extensionDuration: 6
+                agmYear: '2024'
+                agmDueDate: '2026-06-20'
+                alreadyExtended: false
+                isGoodStanding: true
+                isPrevExtension: false
+                prevAgmDate: null
+                prevExpiryDate: null
+                requestExpired: false
+                isEligible: true
```
chenhongjing, line 1634: Are `currentDate`, `incorporationDate`, `isGoodStanding`... necessary in schema?

chenhongjing (commented):
Looks good, can you help to rename `Agm` to `AGM` in summary, title and description sections?

docs/business.yaml
```diff
@@ -1417,6 +1514,124 @@ components:
...
+            agmDueDate:
+              type: string
+              format: date
+              description: The due date for the AGM after the extension is applied.
+              example: '2026-06-20'
+            isEligible:
+              type: boolean
+              description: Indicates if the company is eligible for an AGM extension.
+              example: true
+          x-examples:
```
eason-pan-bc, line 1615: Since we are providing comprehensive request and response examples under the path. Is the x-example necessary?

docs/business.yaml
AimeeGao: Thanks for pointing this out.

The fields `currentDate`, `incorporationDate`, `isGoodStanding`, etc., were added based on the information returned by the API when performing a POST filing call. 
image

We encountered a similar issue previously when updating the business schema. At that time, we found some fields in the GET business API response that were not present in the business schema. The feedback we received was to first add the fields from the API response into the schema. Therefore, for the AGM extension, I followed the same approach and added the fields based on the API return content.

Not sure, for the AGM extension should we also follow the same approach and add the fields based on the API return content?

AimeeGao:
> Looks good, can you help to rename `Agm` to `AGM` in summary, title and description sections?

Thank you for the suggestion. Make sense to me.
@eason-pan-bc  should we also update the naming of AGM in other sections like the agm location change to maintain consistency?

docs/business.yaml
AimeeGao: Yes, we have provided comprehensive request and response examples under the path. The example under the schema was added following the existing writing style of other schemas in the codebase. If we think it is unnecessary, I can remove it.

eason-pan-bc:
> > Looks good, can you help to rename `Agm` to `AGM` in summary, title and description sections?
> 
> Thank you for the suggestion. Make sense to me. @eason-pan-bc should we also update the naming of AGM in other sections like the agm location change to maintain consistency?

Yes, I changed it before the Voluntary Dissolution PR got merged.

docs/business.yaml
eason-pan-bc: I didn't put x-examples under AGM location change schema. I think we had this in other schemas, since there weren't examples under each path.

docs/business.yaml
argush3: As per our conversation, let's just return whatever is returned by the api right now as it doesn't map to an actual json schema definition.

chenhongjing: OK, thanks! And resolved!

docs/business.yaml
```diff
@@ -1590,6 +1687,124 @@ components:
           addressCountry: Canada
           postalCode: V8R 2P1
           deliveryInstructions: 'Our apartment is located at the back of the building.'
+    Agm_extension:
+      type: object
+      title: AGM Extension Information Schema
+      description: Notice of AGM(Annual General Meeting) date extension allows companies to request an extension to the deadline for holding an Annual General Meeting (AGM). The longest extension granted at one time is six months.
```
argush3, line 1693: As per our conversation, the response examples can be used as is but the schema definition should exclude fields not included in the corresponding schema in the business schema repo.

The schema definitions are mostly used for API POST/PUT/PATCH operations.  The only schema that is used for response is the business schema.

AimeeGao: Thanks for the detailed explanation, Argus. 
I got it. Updated the schema, only use the fields from the business schema for AGM extension.

argush3 (approved):

PR #2861, opened 2024-07-25
initial api spec update for incorporation application

PaulGarewal:
*Issue #:* /bcgov/entity#22115

*Description of changes:*

- initial commit for incorporation application
- missing some failed responses, pushing now as I will be away Friday and Monday, to get comments ready for when I come back.

### DIFF
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -106,6 +106,109 @@ paths:
                 type: object
                 properties: {}
               examples:
+                Incorporation-application-response:
+                  summary: Incorporation application response  
+                  value:
+                    errors:
+                      - null
+                    filing:
+                      business:
+                        identifier: TqcTmFjD9X  
+                      header:
+                        accountId: 3445
+                        affectedFilings: []
+                        availableOnPaperOnly: false
+                        certifiedBy: API Specs
+                        colinIds: []
+                        comments: []
+                        date: '2024-07-25T16:17:57.191108+00:00'
+                        deletionLocked: false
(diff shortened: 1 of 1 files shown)


docs/business.yaml
```diff
@@ -106,6 +106,109 @@ paths:
...
+                        deletionLocked: false
+                        effectiveDate: '2024-07-25T16:17:57.191129+00:00'
+                        email: no_one@never.get
+                        filingId: 150333
+                        inColinOnly: false
+                        isCorrected: false
+                        isCorrectionPending: false
+                        name: incorporationApplication
+                        status: DRAFT
+                        submitter: api-key-account-3445-sandbox
```
AimeeGao, line 134: Great work. I have a small suggestion regarding the submitter field. To prevent any potential misuse, perhaps we could use a more generic placeholder. 
This is how we handle this field: `submitter: mocked submitter`

eason-pan-bc (commented):
Overall, nice work!! Other than the ones in the comments, 
- please double-check whether the relevant schemas (e.g. `cooperative`, `name_translations`, `share_sctructure`, etc.) used in incorporation application match the ones in Business Schemas. https://github.com/bcgov/business-schemas
- please double-check the endpoint used for IA

ketaki-deodhar (changes requested):
Paul, this not the correct place for Incorporation application. This path is for submitting a filing for an existing business. IA is to create a new business. it should be under path `\businesses`

PR #2864, opened 2024-07-29
 22293 update change of directors spec

AimeeGao:
*Issue #:* /bcgov/entity#22293

*Description of changes:*

- Updated change of directors schema to the schema section under components
- Added examples of request and response for change of directors filing
- Added error response for change of directors filing

**Request**
image

**Response**
image

image
image
image
image

### DIFF
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -503,6 +503,196 @@ paths:
                         status: PENDING
                         submitter: mocked submitter
                         waiveFees: false
+                change-of-directors-success-response:
+                  summary: Change Of Directors Response
+                  value:
+                    filing:
+                      business:
+                        foundingDate: '2024-07-08T07:00:00.000+00:00'
+                        identifier: BC0871442
+                        legalName: JIA TEST BC TO ULC
+                        legalType: ULC
+                      changeOfDirectors:
+                        directors:
+                          - actions:
+                            - appointed
+                            appointmentDate: '2024-07-09'
+                            cessationDate: null
+                            deliveryAddress:
+                              addressCity: Richmond
+                              addressCountry: CA
+                              addressRegion: BC
+                              deliveryInstructions: ''
+                              postalCode: V7E 1V2
+                              streetAddress: 10300 4th Ave
+                              streetAddressAdditional: ''
+                            mailingAddress:
+                              addressCity: Richmond
+                              addressCountry: CA
+                              addressRegion: BC
+                              deliveryInstructions: ''
+                              postalCode: V7E 1V2
+                              streetAddress: 10300 4th Ave
+                              streetAddressAdditional: ''
+                            officer:
+                              firstName: Doe
+                              lastName: Smith
+                              middleInitial: ''
+                              prevFirstName: ''
+                              prevLastName: ''
+                              prevMiddleInitial: ''
+                          - actions:
+                            - appointed
+                            appointmentDate: '2024-07-09'
+                            cessationDate: null
+                            deliveryAddress:
+                              addressCity: Victoria
+                              addressCountry: CA
+                              addressRegion: BC
+                              deliveryInstructions: ''
+                              postalCode: V8W 1W4
+                              streetAddress: 1-940 Blanshard
+                              streetAddressAdditional: ''
+                            mailingAddress:
+                              addressCity: Victoria
+                              addressCountry: CA
+                              addressRegion: BC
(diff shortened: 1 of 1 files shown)


docs/business.yaml
```diff
@@ -503,6 +503,196 @@ paths:
...
+                              streetAddressAdditional: ''
+                            mailingAddress:
+                              addressCity: Victoria
+                              addressCountry: CA
+                              addressRegion: BC
+                              deliveryInstructions: ''
+                              postalCode: V8W 1W4
+                              streetAddress: 1-940 Blanshard
+                              streetAddressAdditional: ''
+                            officer:
```
eason-pan-bc, line 564: I saw `partyType` in some of the `officer` object in the Postman collection, is it just optional?

docs/business.yaml
```diff
@@ -503,6 +503,196 @@ paths:
...
```
eason-pan-bc, line 668: Found an `email`  under some of the `officer` in the Postman collection. Not sure about it, but maybe you'd like to add them to here?
image
image

docs/business.yaml
```diff
@@ -2712,41 +3097,122 @@ components:
...
-            $ref: '#/components/schemas/party'
+          type: object
+          description: 'This section contains all the changes you want to make to your company''s directors.'
+          required:
+            - directors
+          properties:
+            directors:
+              type: array
+              items:
+                allOf:
```
eason-pan-bc, line 3108: Maybe consider having `director` and `actions` as 2 separate schemas and reference them in the CoD? Since, there is a chance that other places need to use them in the future.
image

docs/business.yaml
AimeeGao: Thank you for pointing that out. 

Not sure about the `partyType`. According to the [business schema](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/directors.json), I did not see this field referenced in the person structure under change of directors. Currently, when sending a POST request without this `partyType`, the change of directors filing works correctly.

Should we discuss this situation like finding the reference where partyType is used? And think about how to deal with the case like if it needs to be included or if it's optional.

eason-pan-bc: as per our discussion, just update the Postman collection to follow the updated ones in the specs when you have time.

docs/business.yaml
AimeeGao: Yes, I noticed that the `directors` schema is also used in the annual report. Thank you for pointing this out. I think creating a separate directors schema based on the structure in the [business schema](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/directors.json) would be a good idea.

Regarding `actions`, I am not sure if we should create a separate schema since right now this definition is only used in the change of directors.

@chenhongjing  @argush3  Hope we can get your insights or suggestions on this matter when you have a moment.

eason-pan-bc: Yeah, I think you are right, the `actions` is only used in CoD, since it's from the definitions, not from a separate schema. 
For the `directors`, probably creating a separate one and update the places using $ref is a good idea?

chenhongjing: I think it's good to create a separate schema for `director` while keeping `actions` within COD.

docs/business.yaml
```diff
@@ -3740,6 +3680,78 @@ components:
                 title: Path of the property in json
       description: '***Item not used*** What will be changed in the correction filing.'
       type: object
+    Directors:
+      title: Directors Schema
+      type: object
```
argush3, line 3685: think the schema should just represent on director.  and anything using it can define an array of directors referencing this schema if required.

docs/business.yaml
```diff
@@ -2712,41 +3097,62 @@ components:
       description: 'Add or remove a director, or update information regarding a current director.'
```
chenhongjing, line 3097: Can you help to add `title` for this schema (and for change of address) like that is done for AGM Extension (line 2463)?

docs/business.yaml
AimeeGao: Thank you, Argus. 
Yes, you're right. Will update it.

docs/business.yaml
AimeeGao: Sure

chenhongjing (approved):

eason-pan-bc (approved):
LGTM 😃

argush3 (approved):

PR #2902, opened 2024-08-08
22204 - Created alembic script to update batch_processing enum

leodube-aot:
*Issue #:* /bcgov/entity#22204

*Description of changes:*
- Created alembic script to update batch_processing enum
- Should be merged before #2903 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #2903, opened 2024-08-08
22204 Filer + Dissolutions Job - Updated to support queued status

leodube-aot:
*Issue #:* /bcgov/entity#22204

*Description of changes:*
- Updated to support queued status
- Waiting for #2902 to merge before I can fully confirm this works and can mark as ready for review.

### DIFF
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -284,7 +284,7 @@ async def stage_3_process(app: Flask, qsm: QueueService):
             batch_processing.filing_id = filing.id
             batch_processing.status = BatchProcessing.BatchProcessingStatus.QUEUED
             app.logger.debug(
-                f'Batch Processing with identifier: {batch_processing.business_identifier} has been marked as queued.'  # noqa: E501
+                f'Batch Processing with identifier: {batch_processing.business_identifier} has been marked as queued.'
             )
         else:
             batch_processing.status = BatchProcessing.BatchProcessingStatus.WITHDRAWN
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -283,9 +283,9 @@ async def stage_3_process(app: Flask, qsm: QueueService):
 
             batch_processing.filing_id = filing.id
             batch_processing.step = BatchProcessing.BatchProcessingStep.DISSOLUTION
-            batch_processing.status = BatchProcessing.BatchProcessingStatus.COMPLETED
+            batch_processing.status = BatchProcessing.BatchProcessingStatus.QUEUED
             app.logger.debug(
-                f'Batch Processing with identifier: {batch_processing.business_identifier} has been marked as complete.'
+                f'Batch Processing with identifier: {batch_processing.business_identifier} has been moved to dissolution step.'
             )
         else:
             batch_processing.status = BatchProcessing.BatchProcessingStatus.WITHDRAWN
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -282,10 +282,9 @@ async def stage_3_process(app: Flask, qsm: QueueService):
             await put_filing_on_queue(filing.id, app, qsm)
 
             batch_processing.filing_id = filing.id
-            batch_processing.step = BatchProcessing.BatchProcessingStep.DISSOLUTION
             batch_processing.status = BatchProcessing.BatchProcessingStatus.QUEUED
             app.logger.debug(
-                f'Batch Processing with identifier: {batch_processing.business_identifier} has been moved to the dissolution step.'  # noqa: E501
+                f'Batch Processing with identifier: {batch_processing.business_identifier} has been marked as queued.'  # noqa: E501
             )
         else:
             batch_processing.status = BatchProcessing.BatchProcessingStatus.WITHDRAWN
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -279,19 +279,22 @@ async def stage_3_process(app: Flask, qsm: QueueService):
         if eligible:
             filing = create_invountary_dissolution_filing(batch_processing.business_id)
             app.logger.debug(f'Created Involuntary Dissolution Filing with ID: {filing.id}')
-            await put_filing_on_queue(filing.id, app, qsm)
-
             batch_processing.filing_id = filing.id
             batch_processing.step = BatchProcessing.BatchProcessingStep.DISSOLUTION
             batch_processing.status = BatchProcessing.BatchProcessingStatus.QUEUED
+            batch_processing.last_modified = datetime.utcnow()
+            batch_processing.save()
+
+            await put_filing_on_queue(filing.id, app, qsm)
+
             app.logger.debug(
                 f'Batch Processing with identifier: {batch_processing.business_identifier} has been marked as queued.'
             )
         else:
             batch_processing.status = BatchProcessing.BatchProcessingStatus.WITHDRAWN
             batch_processing.notes = 'Moved back into good standing'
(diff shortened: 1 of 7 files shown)


jobs/involuntary-dissolutions/involuntary_dissolutions.py
```diff
@@ -281,10 +281,9 @@ async def stage_3_process(app: Flask, qsm: QueueService):
             app.logger.debug(f'Created Involuntary Dissolution Filing with ID: {filing.id}')
             await put_filing_on_queue(filing.id, app, qsm)
 
-            batch_processing.step = BatchProcessing.BatchProcessingStep.DISSOLUTION
-            batch_processing.status = BatchProcessing.BatchProcessingStatus.COMPLETED
+            batch_processing.status = BatchProcessing.BatchProcessingStatus.QUEUED
```
leodube-aot, line 284: Do we want the filer to handle changing the step to `DISSOLUTION`?

chenhongjing: hmmmm...think we can keep it in the job. 😕 💭

argush3: I think the filer should change the step from `QUEUED` to `DISSOLUTION`.  Otherwise, not sure how we would be able to tell the dissolution filing went through in a timely way for a business being dissolved.

My original vision was to have another listener that would process any QUEUED entries in the dissolution process but decided not to do it for the MVP implementation.

Let me know if you have more thoughts about this @chenhongjing and @leodube-aot

chenhongjing: Was Leo talking about changing `BatchProcessingStep` from `WARNING_LEVEL_2` to `DISSOLUTION` instead of changing `BatchProcessingStatus` from `COMPLETED` to `QUEUED`?

leodube-aot: I was wondering about moving these 2 lines (284 and 285) [ into the filer](https://github.com/bcgov/lear/pull/2903/files#diff-54617f2b54e8519023b9faaa7fbaf8083aaada150add761198d8972384613f40R105-R106), so the processing step will remain at `WARNING_LEVEL_2` in this job until the filing can be processed

chenhongjing: I was talking about keeping step change in the job but move status change to filer...not sure if that is what you want @argush3 💭

argush3: > I was talking about keeping step change in the job but move status change to filer...not sure if that is what you want @argush3 💭

ok, sorry, i was getting confused a bit between step and status I think.

Yes, I think it makes sense to keep `batch_processing.step = BatchProcessing.BatchProcessingStep.DISSOLUTION` in the dissolutions job and to do the status change(`QUEUED` to `COMPLETED`) in filer.

@leodube-aot hopefully that makes sense.  I think doing it this way we'll be able to easily tell if a business got to stage 3 and if filer has processed the dissolution filing successfully yet.

leodube-aot:
### Notes on the unit tests:

_Involuntary Dissolutions Job tests:_ All tests pass.

_Legal API tests:_ No new tests fail.

_Entity Filer tests:_ This one is a bit trickier. The CI shows several dissolutions related unit tests are failing, however no additional information can be gathered from it since it just shows the `No such file or directory: 'docker-compose'` error. When I run the unit tests locally the tests pass:

Screenshot 2024-08-14 at 9 46 23 AM

jobs/involuntary-dissolutions/involuntary_dissolutions.py
leodube-aot: I moved the step process change back into the job, and made a note about the unit tests below. This PR is ready to have a second look at!

queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
```diff
@@ -95,6 +97,15 @@ def process(business: Business, filing: Dict, filing_rec: Filing, filing_meta: F
             'dissolutionDate': LegislationDatetime.format_as_legislation_date(business.dissolution_date)
         }
 
+    # update batch processing entry, if any is present
+    if flag_on:
+        batch_processings = BatchProcessing.find_by(business_id=business.id)
```
argush3, line 102: i think you should search the batch processing entry by filing id.

leodube-aot: Updated and confirmed it works locally + unit tests still pass locally

chenhongjing (approved):
LGTM 😸

queue_services/entity-filer/tests/unit/filing_processors/test_dissolution.py
```diff
@@ -92,14 +93,33 @@ def test_dissolution(app, session, minio_server, legal_type, identifier, dissolu
...
+            batch_id=batch.id,
+            business_id=business.id,
+            business_identifier=business.identifier,
+            step=BatchProcessing.BatchProcessingStep.DISSOLUTION,
+            status=BatchProcessing.BatchProcessingStatus.QUEUED,
+            created_date=datetime.utcnow()-datedelta(days=42),
+            trigger_date=datetime.utcnow(),
+            last_modified=datetime.utcnow()
+        )
+        batch_processing.save()
```
argush3, line 113: you said your test works?  don't you need the filing id to be populated here?

leodube-aot: Just reran the tests for this file and yeah they still do, but good point I feel like that does need to be populated. Having a look into what's going on now.

Screenshot 2024-08-14 at 1 50 01 PM

leodube-aot: Figured out what's going on. If `BatchProcessing.find_by` is called with `None` for any of the parameters, then it will return a list of **all** batch_processings. 

Screenshot 2024-08-14 at 2 02 30 PM

Since only one batch_processing exists for a given test case, the tests were still passing properly. Anyway, I've updated the unit tests to set up the batch_processing correctly.

argush3: that makes sense but i still expected the test to fail in that scenario.  i think the issue might be how we are grabbing the dissolution filing id.

queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
```diff
@@ -95,6 +97,15 @@ def process(business: Business, filing: Dict, filing_rec: Filing, filing_meta: F
             'dissolutionDate': LegislationDatetime.format_as_legislation_date(business.dissolution_date)
         }
 
+    # update batch processing entry, if any is present
+    if flag_on:
+        batch_processings = BatchProcessing.find_by(filing_id=dissolution_filing.get("id"))
```
argush3, line 102: I don't think the filing json will have the filing id?

maybe try using `filing_rec.id`?

queue_services/entity-filer/tests/unit/filing_processors/test_dissolution.py
leodube-aot: Yeah that could be it, I updated to grab it from `filing_rec` based on your comment below

queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py
chenhongjing: Yep, try to use `filing_rec.id`. And I have one more question, do we know if filing_id must be saved into batch_processing before filer processes it? I notice [it](https://github.com/bcgov/lear/blob/911b7ef4efa5168e97fa7108df2d97f7689e665d/jobs/involuntary-dissolutions/involuntary_dissolutions.py#L294) is saved after the message is pushed into queue (though I know filer will process it async)

leodube-aot: I think it's probably a good call to save the filing_id to the batch_processing before putting it on the queue. In theory the filer won't process it right away, which will give the job a chance to save the batch_processing entry (this was my experience testing locally). Still better to be safe imo so I've updated the job.

argush3 (approved):

PR #2911, opened 2024-08-13
21954 - Sync involuntary dissolutions back to COLIN

chenhongjing:
*Issue #:* /bcgov/entity#21954

*Description of changes:*
- No updates for update-colin-filings job
- Updated colin-api to support adding events for involuntary dissolutions
- Updated legal-api to populate batch_processing.meta_data into filing json returned for COLIN related services

Note that similar with administrative dissolution, no filing will be created for involuntary dissolution. It just creates events and changes the state of the companies.

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/colin_sync.py
+++ b/legal-api/src/legal_api/resources/v2/business/colin_sync.py
@@ -87,7 +87,7 @@ def get_completed_filings_for_colin():
                     continue  # do not break this function because of one filing
             elif (filing.filing_type == 'dissolution' and
                   filing.filing_sub_type == 'involuntary'):
-                batch_processings = BatchProcessing.find_by(business_id=filing.business_id)
+                batch_processings = BatchProcessing.find_by(filing_id=filing.id)
                 if not batch_processings:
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/resources/v2/business/colin_sync.py
```diff
@@ -76,6 +85,12 @@ def get_completed_filings_for_colin():
                 except Exception as ex:  # noqa: B902
                     current_app.logger.info(ex)
                     continue  # do not break this function because of one filing
+            elif (filing.filing_type == 'dissolution' and
+                  filing.filing_sub_type == 'involuntary'):
+                batch_processings = BatchProcessing.find_by(business_id=filing.business_id)
```
argush3, line 90: I think there is a potential scenario where a business that has been involuntarily dissolved that gets restore and involuntarily dissolved again.  Your code would still take the first `batch_processing` entry which is ok as long as the sort order is correct.  But searching for the matching `batch_processing` entry using the `batch_processing.filing_id` is probably the safest.

vysakh-menon-aot (approved):

argush3 (approved):

PR #2931, opened 2024-08-19
22673 - Hide involuntary dissolution ledger item

JazzarKarim:
*Issue #:* /bcgov/entity#22673

*Description of changes:*
- Hid the involuntary dissolution ledger entry/item
- Added unit test

Example of postman calls:
invol display item
vol display item

### DIFF
--- a/legal-api/src/legal_api/core/filing.py
+++ b/legal-api/src/legal_api/core/filing.py
@@ -460,10 +460,14 @@ class Filing:
     @staticmethod
     def _is_display_ledger(filing: FilingStorage) -> bool:
         """Return boolean that display the ledger."""
-        # If filing is an admin freeze or involuntary dissolution, we will NOT display it on ledger
-        if (filing.filing_type == Filing.FilingTypes.ADMIN_FREEZE or
-            (filing.filing_type == Filing.FilingTypes.DISSOLUTION and filing.filing_sub_type == 'involuntary')): return False
-        return True
+        # If filing is NOT an admin freeze or involuntary dissolution, we will display it on ledger
+        return not (
+            filing.filing_type == Filing.FilingTypes.ADMIN_FREEZE or
+            (
+                filing.filing_type == Filing.FilingTypes.DISSOLUTION and
+                filing.filing_sub_type == 'involuntary'
+            )
+        )
 
     @staticmethod
     def get_document_list(business,  # pylint: disable=too-many-locals disable=too-many-branches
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/core/filing.py
```diff
@@ -460,7 +460,10 @@ def _add_ledger_order(filing: FilingStorage, ledger_filing: dict) -> dict:
     @staticmethod
     def _is_display_ledger(filing: FilingStorage) -> bool:
         """Return boolean that display the ledger."""
-        return filing.filing_type != Filing.FilingTypes.ADMIN_FREEZE
+        # If filing is an admin freeze or involuntary dissolution, we will NOT display it on ledger
+        if (filing.filing_type == Filing.FilingTypes.ADMIN_FREEZE or
+            (filing.filing_type == Filing.FilingTypes.DISSOLUTION and filing.filing_sub_type == 'involuntary')): return False
```
JazzarKarim, line 465: I changed the logic here so that it's easier to understand and read. I got rid of the negation.

Also, I was hoping to use an ENUM for the `involuntary` but saw that the string was being used everywhere as well. I was thinking of changing it everywhere to use an enum.

legal-api/src/legal_api/core/filing.py
```diff
@@ -460,7 +460,14 @@ def _add_ledger_order(filing: FilingStorage, ledger_filing: dict) -> dict:
     @staticmethod
     def _is_display_ledger(filing: FilingStorage) -> bool:
         """Return boolean that display the ledger."""
-        return filing.filing_type != Filing.FilingTypes.ADMIN_FREEZE
+        # If filing is NOT an admin freeze or involuntary dissolution, we will display it on ledger
+        return not (
+            filing.filing_type == Filing.FilingTypes.ADMIN_FREEZE or
+            (
+                filing.filing_type == Filing.FilingTypes.DISSOLUTION and
+                filing.filing_sub_type == 'involuntary'
```
JazzarKarim, line 468: I was hoping to use an ENUM for the involuntary but saw that the string was being used everywhere as well. I was thinking of changing it everywhere to use an enum.

kzdev420 (approved):

legal-api/src/legal_api/core/filing.py
leodube-aot: I think it makes sense to use an enum, if it isn't too much effort to make the same changes everywhere else

JazzarKarim: I'm just going to leave it for now. If this pops up again, might be a good idea to create an enum then.

chenhongjing (approved):

PR #2943, opened 2024-08-26
22353 correction of C, CBEN, CCC, CUL

vysakh-menon-aot:
*Issue #:* /bcgov/entity#22353

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/incorporation_application.py
+++ b/legal-api/src/legal_api/services/filings/validations/incorporation_application.py
@@ -133,7 +133,11 @@ def validate_roles(filing_dict: dict, legal_type: str, filing_type: str = 'incor
         Business.LegalTypes.BCOMP.value: 1,
         Business.LegalTypes.COMP.value: 1,
         Business.LegalTypes.BC_ULC_COMPANY.value: 1,
-        Business.LegalTypes.BC_CCC.value: 3
+        Business.LegalTypes.BC_CCC.value: 3,
+        Business.LegalTypes.BCOMP_CONTINUE_IN.value: 1,
+        Business.LegalTypes.CONTINUE_IN.value: 1,
+        Business.LegalTypes.ULC_CONTINUE_IN.value: 1,
+        Business.LegalTypes.CCC_CONTINUE_IN.value: 3
     }
     parties_array = filing_dict['filing'][filing_type]['parties']
     msg = []
--- a/queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
+++ b/queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
@@ -47,9 +47,7 @@ def process(correction_filing: Filing, filing: Dict, filing_meta: FilingMeta, bu
     )
 
     corrected_filing_type = filing['correction']['correctedFilingType']
-    # added CP, change of directors / change of address for CP is allowed
-    if business.legal_type in ['SP', 'GP', 'BC', 'BEN', 'CC', 'ULC', 'CP'] and \
-            corrected_filing_type != 'conversion':
+    if corrected_filing_type != 'conversion':
         correct_business_data(business, correction_filing, filing, filing_meta)
     else:
         # set correction filing to PENDING_CORRECTION, for manual intervention
--- a/queue_services/entity-filer/tests/unit/test_worker/test_worker.py
+++ b/queue_services/entity-filer/tests/unit/test_worker/test_worker.py
@@ -422,66 +422,6 @@ async def test_process_filing_completed(app, session, mocker):
     assert business.last_ar_date
 
 
-async def test_correction_filing(app, session, mocker):
-    """Assert we can process a correction filing."""
-    # vars
-    payment_id = str(random.SystemRandom().getrandbits(0x58))
-    identifier = 'CP1111111'
-    correction_filing_comment = 'We need to fix directors'
-
-    # get a fixed datetime to use in comparisons, in "local" (Pacific) timezone
-    local_timezone = pytz.timezone('US/Pacific')
-    correction_filing_date = \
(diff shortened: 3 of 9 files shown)


legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -133,7 +133,11 @@ def validate_roles(filing_dict: dict, legal_type: str, filing_type: str = 'incor
         Business.LegalTypes.BCOMP.value: 1,
         Business.LegalTypes.COMP.value: 1,
         Business.LegalTypes.BC_ULC_COMPANY.value: 1,
-        Business.LegalTypes.BC_CCC.value: 3
+        Business.LegalTypes.BC_CCC.value: 3,
```
vysakh-menon-aot, line 136: This func is used in correction

queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
```diff
@@ -47,9 +47,7 @@ def process(correction_filing: Filing, filing: Dict, filing_meta: FilingMeta, bu
     )
 
     corrected_filing_type = filing['correction']['correctedFilingType']
-    # added CP, change of directors / change of address for CP is allowed
-    if business.legal_type in ['SP', 'GP', 'BC', 'BEN', 'CC', 'ULC', 'CP'] and \
```
vysakh-menon-aot, line 51: unnecessary if condition, this contains all legal types

queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
```diff
@@ -47,9 +47,7 @@ def process(correction_filing: Filing, filing: Dict, filing_meta: FilingMeta, bu
...
-            corrected_filing_type != 'conversion':
+    if corrected_filing_type != 'conversion':
```
vysakh-menon-aot, line 50: not sure if we need this if as well. any thoughts @argush3?

argush3:
> @argush3 I looked into the `entity-emailer` test failure, It seems to me that this issue is because of the legal name change in business.json() for SP's. Not sure if it's a real issue while emailing or just a unit test issue. If you don't have a ticket for this please create one

yeah, it's here https://app.zenhub.com/workspaces/entities---david-65af11a89e89f5043c2911f6/issues/gh/bcgov/entity/22367

queue_services/entity-filer/tests/unit/test_worker/test_worker.py
```diff
@@ -422,66 +422,6 @@ async def test_process_filing_completed(app, session, mocker):
     assert business.last_ar_date
 
 
-async def test_correction_filing(app, session, mocker):
```
vysakh-menon-aot, line 425: not valid anymore. CP will not go to PENDING_CORRECTION status

queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
severinbeauvais: What if we add new legal types?

queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
argush3: I think we can remove L50 now.  We previously didn't allow corrections of corrections but we allowed it via https://github.com/bcgov/entity/issues/16858.  think we forgot to update the filter for this.

severinbeauvais: FWIW, the UI doesn't allow correction of a record conversion. That's a function of a FF and allowable actions. This check here might be good for API safety...?

queue_services/entity-filer/tests/unit/test_worker/test_worker.py
severinbeauvais: That status was for "local corrections", right?

Yes, local corrections are removed from UI code. The only possibility is something like a conversion correction (as above), but even that's not allowed, I think.

queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
vysakh-menon-aot: it should be handled in allowable actions

queue_services/entity-filer/src/entity_filer/filing_processors/correction.py
vysakh-menon-aot: Since we are correcting latest data checking corrected_filing_type make no sense

severinbeauvais: FWIW2, there are other filings that we don't allow correcting.

Ref: https://github.com/bcgov/business-filings-ui/blob/29cc2f6b204f3755ad13eafd45aa4e524337b071/src/components/Dashboard/FilingHistoryList/HeaderActions.vue#L136

argush3: actually, after looking at https://app.zenhub.com/workspaces/entities---david-65af11a89e89f5043c2911f6/issues/gh/bcgov/entity/16858 more closely, it seems like the change was related to allow local correction of conversion filing. 

The L50 check is used to make sure the correction stays in `PENDING_CORRECTION` status instead of actually getting processed by the filer like other filing types.

So I think we should keep L50 as is.

severinbeauvais: > Since we are correcting latest data checking corrected_filing_type make no sense

So nothing is done with corrected_filing_type? It's obsolete? (Makes sense to me.)

argush3 (approved):

severinbeauvais (approved):

PR #2949, opened 2024-08-28
22011 - Reviews endpoint: apply sorting and filtering

ketaki-deodhar:
*Issue #:* /bcgov/entity#22011

*Description of changes:*

- filter by date range, NR number, identifying number, completing party, status
- sorting in table header (all fields)
- initially sort oldest to newest

example url with params: `{{url}}/api/v2/admin/reviews?startDate=2024-07-05&endDate=2024-08-23&sortBy=submissionDate&sortDesc=true&page=1&limit=20`


legal-api/src/legal_api/models/dataclass.py
```diff
@@ -0,0 +1,33 @@
...
+    status: List[str] = field()
+    start_date: str = ''
+    end_date: str = ''
+    nr_number: str = ''
+    identifier: str = ''
+    completing_party: str = ''
+    submitted_sort_by: str = ''
+    submitted_sort_order: bool = ''
+    page: int = 1
+    limit: int = 10
```
vysakh-menon-aot, line 33: Can you move this into `review.py` instead of separate file. For now we don't have any thing other than model in this folder. I would have suggested to move this into `utils` or create a new folder for dataclass, since we are planning to strip out model into a separate lib while migrating to GCP its better to keep it inside

legal-api/src/legal_api/models/review.py
```diff
@@ -79,15 +80,48 @@ def get_review(cls, filing_id) -> Review:
...
-
-        pagination = query.paginate(per_page=limit, page=page)
+            join(Filing, Filing.id == Review.filing_id)
+
+        if review_filter.start_date:
+            start_date_utc = LegislationDatetime.as_utc_timezone_from_legislation_date_str(review_filter.start_date)
+            query = query.filter(Review.submission_date >= start_date_utc)
+        if review_filter.end_date:
+            end_date_utc = LegislationDatetime.as_utc_timezone_from_legislation_date_str(review_filter.end_date)
+            query = query.filter(Review.submission_date <= end_date_utc)
```
vysakh-menon-aot, line 93: I am not sure about the date conversion. (https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/review.py#L120) in json we just provide a date (without conversion), does the UI provide a date in legislative timezone? @severinbeauvais

vysakh-menon-aot:
@ketaki-deodhar Looks good to me. Can you add test case for these

legal-api/src/legal_api/models/review.py
severinbeauvais: @ArwenQin ^^^ What date-time format do you expect in the UI?

ArwenQin: I expect '2024-07-01T19:00:00.000+00:00'

ArwenQin (approved):

legal-api/src/legal_api/models/review.py
vysakh-menon-aot: In json response its in the isoformat (UTC). I believe you convert that before showing in the table. When user select a date from the filter do you convert that to UTC before calling API?

ketaki-deodhar: This what I see in network tab when I select date range:
https://legal-api-dev.apps.silver.devops.gov.bc.ca/api/v2/admin/reviews?startDate=2024-08-01&endDate=2024-08-28&nrNumber=&identifier=&completingParty=&status=CHANGE_REQUESTED&status=RESUBMITTED&status=AWAITING_REVIEW&sortBy=submissionDate&sortDesc=false&page=1&limit=5

startDate=2024-08-01
endDate=2024-08-28

ArwenQin: I convert it to 'yyyy-mm-dd' when show it in the table; and when user selects a date from the filter, I pass in  'yyyy-mm-dd' format, date only and without time

severinbeauvais: > In json response its in the isoformat (UTC). I believe you convert that before showing in the table.

Yes.

> When user select a date from the filter do you convert that to UTC before calling API?

No. It's just a YYYY-MM-DD date (assumed to be Pacific timezone).

severinbeauvais: Ketaki, if I understand this correctly, you are converting from a "legislation date string" (date only in Pacific tz) to a UTC date-time string for the db query. This sounds right to me.

legal-api/src/legal_api/models/review.py
```diff
@@ -79,15 +80,48 @@ def get_review(cls, filing_id) -> Review:
...
+        if review_filter.nr_number:
+            query = query.filter(Review.nr_number.ilike(f'%{review_filter.nr_number}%'))
+        if review_filter.identifier:
+            query = query.filter(Review.identifier.ilike(f'%{review_filter.identifier}%'))
+        if review_filter.completing_party:
+            query = query.filter(Review.identifier.ilike(f'%{review_filter.completing_party}%'))
+        if review_filter.status:
+            query = query.filter(Review.status.in_(review_filter.status))
```
severinbeauvais, line 101: Does this support multiple statuses?

severinbeauvais: Doh! That's what `in` means, isn't it? Resolving.

legal-api/src/legal_api/models/review.py
ketaki-deodhar: Yes that is correct

severinbeauvais (approved):

vysakh-menon-aot (approved):

PR #2956, opened 2024-08-29
22011 - implement filter and sorting for future effective date

ketaki-deodhar:
*Issue #:* /bcgov/entity#22011

*Description of changes:*

implemented filter and sorting for future effective date (missed it out earlier)

### DIFF
--- a/legal-api/src/legal_api/models/review.py
+++ b/legal-api/src/legal_api/models/review.py
@@ -89,10 +89,12 @@ class Review(db.Model):  # pylint: disable=too-many-instance-attributes
 
         if review_filter.start_date:
             start_date_utc = LegislationDatetime.as_utc_timezone_from_legislation_date_str(review_filter.start_date)
-            query = query.filter(Review.submission_date >= start_date_utc)
+            start_of_day = datetime.combine(start_date_utc, datetime.min.time())
+            query = query.filter(Review.submission_date >= start_of_day)
         if review_filter.end_date:
             end_date_utc = LegislationDatetime.as_utc_timezone_from_legislation_date_str(review_filter.end_date)
-            query = query.filter(Review.submission_date <= end_date_utc)
+            end_of_day = datetime.combine(end_date_utc, datetime.max.time())
+            query = query.filter(Review.submission_date <= end_of_day)
         if review_filter.nr_number:
             query = query.filter(Review.nr_number.ilike(f'%{review_filter.nr_number}%'))
         if review_filter.identifier:
--- a/legal-api/src/legal_api/models/review.py
+++ b/legal-api/src/legal_api/models/review.py
@@ -100,7 +100,7 @@ class Review(db.Model):  # pylint: disable=too-many-instance-attributes
         if review_filter.identifier:
             query = query.filter(Review.identifier.ilike(f'%{review_filter.identifier}%'))
         if review_filter.completing_party:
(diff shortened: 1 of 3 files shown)


ArwenQin (approved):
Looks good!

legal-api/src/legal_api/models/review.py
```diff
@@ -88,13 +89,19 @@ def get_paginated_reviews(cls, review_filter, mapped_sort_by_column):
             join(Filing, Filing.id == Review.filing_id)
 
         if review_filter.start_date:
-            start_date_utc = LegislationDatetime.as_utc_timezone_from_legislation_date_str(review_filter.start_date)
-            start_of_day = datetime.combine(start_date_utc, datetime.min.time())
-            query = query.filter(Review.submission_date >= start_of_day)
+            start_date = LegislationDatetime.get_start_of_date(review_filter.start_date)
+            query = query.filter(Review.submission_date >= start_date)
```
vysakh-menon-aot, line 93: Can you try this? We may not need `get_start_of_date`
```suggestion
            start_date = LegislationDatetime.as_utc_timezone_from_legislation_date_str(review_filter.start_date)
            query = query.filter(Review.submission_date >= start_date.date())
```

legal-api/src/legal_api/utils/legislation_datetime.py
```diff
@@ -137,3 +137,17 @@ def is_future(date_string: str) -> bool:
...
+
+    @staticmethod
+    def get_start_of_date(date_string: str) -> datetime:
+        """Convert date string to date as UTC and return start of the day for a DateTime."""
+        date_utc = LegislationDatetime.as_utc_timezone_from_legislation_date_str(date_string)
+        start_of_day = datetime.combine(date_utc, datetime.min.time())
+        return start_of_day
+
+    @staticmethod
+    def get_end_of_date(date_string: str) -> datetime:
```
vysakh-menon-aot, line 149: can you move this funtion to https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/utils/datetime.py#L21 and rename to end_of_day

then you can use `datetime.end_of_day()`

legal-api/src/legal_api/models/review.py
severinbeauvais: I'm OK with either, though converting a date-time to a date seems a bit like cheating to me because we assume the time becomes 00:00. Also I have a slight preference for keeping comparisons of like objects (datetime to datetime , not datetime to date), but maybe that doesn't matter here. FWIW.

legal-api/src/legal_api/utils/legislation_datetime.py
severinbeauvais: Should get_start_of_date() also be renamed and moved to datetime.py?

ketaki-deodhar: I removed this function altogether and keep the code in query itself

legal-api/src/legal_api/models/review.py
ketaki-deodhar: @vysakh-menon-aot unfortunately your suggestion did not work but now the function is consolidated but still needs the start and end of the day

ketaki-deodhar:
examples for both date ranges:

image

image

vysakh-menon-aot (approved):

severinbeauvais (approved):

PR #2963, opened 2024-09-04
22499 Configurations - Alembic script to add furnishings SFTP configurations

leodube-aot:

*Description of changes:*
- Created alembic script to add furnishings SFTP configurations (to address [Argus' comment](https://app.zenhub.com/workspaces/entities---david-65af11a89e89f5043c2911f6/issues/gh/bcgov/entity/22498) in the SFTP spike ticket)

### DIFF
--- a/legal-api/migrations/versions/a7d3a3aa76de_add_sftp_configuration.py
+++ b/legal-api/migrations/versions/a7d3a3aa76de_add_sftp_configuration.py
@@ -1,0 +1,28 @@
+"""Add SFTP configuration.
+
+Revision ID: a7d3a3aa76de
+Revises: 7d486343384b
+Create Date: 2024-08-30 13:07:08.377059
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'a7d3a3aa76de'
+down_revision = '7d486343384b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("""INSERT INTO configurations (name, val, short_description, full_description) VALUES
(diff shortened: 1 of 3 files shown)


chenhongjing:
The other [PR](#2964) looks good, but I'm still thinking about this one... 💭 

Although as Argus mentioned, we need to turn SFTP on and off, flag values in db are more easily changed by accident, as many people have access. Servers of BCLaws and BCMail+ are primitive. I'm worried about what will happen if we transfer unnecessary files to them. 

What do you think of it? What about just skipping the SFTP process if there's no value set in `.env`?

leodube-aot:
> What do you think of it? What about just skipping the SFTP process if there's no value set in `.env`?

That's a good point. I'm okay with just skipping the SFTP process if there is no `.env` connection values set. @argush3 what are your thoughts on doing this vs. having db config values?

argush3:
> > What do you think of it? What about just skipping the SFTP process if there's no value set in `.env`?
> 
> That's a good point. I'm okay with just skipping the SFTP process if there is no `.env` connection values set. @argush3 what are your thoughts on doing this vs. having db config values?

After thinking and talking with Hongjing about this a bit, I think we can use 2 feature flags
* `disable-dissolution-sftp-bcmail`
* `disable-dissolution-sftp-bclaws` feature flags.  

Hongjing has created the two flags in LD already.

The flags will be turned on in dev/test/prod for now to disable SFTP.   When we go live, we can turn off the 2 flags for prod so SFTP works.

As for dev & test, assuming things are configured correctly in 1pass config for the SFTP staging servers for BCLaws & BCMail+, we can toggle on the corresponding flag when we need to test SFTP functionality.

leodube-aot:
Closing this PR, and will update the other PR to handle the feature flags

PR #2964, opened 2024-09-04
22499 Furnishings job - Updates for SFTP XML to BC Laws

leodube-aot:

*Description of changes:*
- Updated config for BCLaws SFTP (also added config options for BCMail+).
- Added a `SftpConnection` class.
- Updated post processor.
- Added unit tests.

_Local Testing:_
**Setup**
1. I created a [git repo with a basic SFTP server with ssh key authentication](https://github.com/leodube-aot/sftp-server) to help with local testing. Clone the repo and follow the `README.md` to get it set up.
2. Update the `jobs/furnishings/.env` file to include the new config variables. See my [comment below](https://github.com/bcgov/lear/pull/2964/files#r1742860422) for more info.
3. I used the same test batch_processings from #2948.

---

**Test `ENABLE_BCLAWS_SFTP` flag off**
With `Configurations.ENABLE_BCLAWS_SFTP=False`, run the furnishings job. Observe:
1. The created `GAZETTE` furnishings are still marked as `QUEUED`. No furnishing group is created and no xml payload is saved.
Screenshot 2024-09-04 at 9 21 38 AM

2. No file is uploaded to the SFTP server
Screenshot 2024-09-04 at 9 20 41 AM  
  
---

**Test SFTP**
Delete the furnishings created in the previous run. With `Configurations.ENABLE_BCLAWS_SFTP=True`, run the furnishings job. Observe:
1. The created `GAZETTE` furnishings are marked as `PROCESSED`, with a new furnishing group.
Screenshot 2024-09-04 at 9 24 52 AM

2. An xml file is uploaded to the SFTP server.
Screenshot 2024-09-04 at 9 26 43 AM

---


jobs/furnishings/src/furnishings/sftp.py
```diff
@@ -0,0 +1,111 @@
...
+
+class PublicKeyAlgorithms(BaseEnum):
+    """Cipher types."""
+
+    RSA = auto()
+    ED25519 = auto()
+
+
+class SftpConnection:
+    """SFTP Connection object."""
```
leodube-aot, line 34: I made some tweaks to the PPR `SftpConnection` object to make it a context manager, so it can be used as:

```
connection = SftpConnection(kwargs)
with connection as client:
    client.put(...)
```

This helps make sure the socket is closed after use. 

I also removed the previous `put` functions since we can just use the [`paramiko.SFTPClient` API](https://docs.paramiko.org/en/latest/api/sftp.html) directly.

jobs/furnishings/src/furnishings/stage_processors/post_processor.py
```diff
@@ -12,87 +12,22 @@
...
 from legal_api.utils.legislation_datetime import LegislationDatetime
+from paramiko.ssh_exception import AuthenticationException, NoValidConnectionsError
 
+from furnishings.sftp import SftpConnection
 
-XML_DATE_FORMAT: Final = '%B %-d, %Y'
-
-
-class PostProcessor:
-    """Processor after stage run of furnishings job."""
```
leodube-aot, line 29: Move the `PostProcessor` class below the `XMLMeta` class so it can be closer to the `process` function.

jobs/furnishings/src/furnishings/stage_processors/post_processor.py
```diff
@@ -152,10 +87,119 @@ def get_info_by_name(name: Furnishing.FurnishingName) -> dict:
...
+        xml_data = self._format_furnishings(self._furnishings_dict, self._processed_date.strftime('%I:%M %p'))
+        self._app.logger.debug('Formatted furnishing details presented in XML file')
+
+        # SFTP to BC Laws
+        payload = self._build_xml_data(xml_data)
+        filename = f'QP_CORP_{LegislationDatetime.format_as_legislation_date(self._processed_date)}.xml'
+
+        with self._bclaws_sftp_connection as client:
+            resp = client.putfo(
+                    fl=StringIO(payload),
```
leodube-aot, line 171: We can use `StringIO` as an in-memory file-like object so that we can upload to the sftp without having to save a file locally first.

jobs/furnishings/src/furnishings/stage_processors/post_processor.py
```diff
@@ -152,10 +87,119 @@ def get_info_by_name(name: Furnishing.FurnishingName) -> dict:
...
         processor = PostProcessor(app, furnishings_dict)
         processor.process()
+    except AuthenticationException as err:
+        notes = 'SFTP Error: Unable to authenticate.'
+        processor.update_furnishings_status(Furnishing.FurnishingStatus.FAILED, notes=notes)
+        app.logger.error(err)
+    except (NoValidConnectionsError, IOError) as err:
+        notes = f'SFTP Error: {os.strerror(err.errno)}.'
+        processor.update_furnishings_status(Furnishing.FurnishingStatus.FAILED, notes=notes)
+        app.logger.error(err)
```
leodube-aot, line 201: Added some error handling to catch SFTP errors. When this happens:
- Furnishings are marked as failed.
- The error message is added to the furnishings note.
- The furnishing group and xml payload are not saved.

jobs/furnishings/tests/unit/test_sftp.py
leodube-aot, line 1: These sftp unit tests use [`pytest-sftpserver`](https://pypi.org/project/pytest-sftpserver/), a local SFTP-server fixture, which serves content from python-like objects. In general, when setting it up for a test, it looks like:

```python
with sftpserver.serve_content({"directory": {"filename.txt": "file content"}}):
   pass
```

One thing to consider is that since it is a python object, multiple directories in the same string do not work.
```python
with sftpserver.serve_content({"dir1/dir2": {"filename.txt": "file content"}}):
        with sftpconnection as sftpclient:
            with sftpclient.open(path.join("dir1", "dir2", "filename.txt")) as file:
                 # will fail

with sftpserver.serve_content({"dir1": {"dir2": {"filename.txt": "file content"}}}):
        with sftpconnection as sftpclient:
            with sftpclient.open(path.join("dir1", "dir2", "filename.txt")) as file:
                 # will work
```

jobs/furnishings/src/furnishings/config.py
```diff
@@ -82,6 +82,24 @@ class _Config:  # pylint: disable=too-few-public-methods
...
     )
 
+    # BCLaws SFTP
+    BCLAWS_SFTP_STORAGE_DIRECTORY = os.getenv('BCLAWS_SFTP_STORAGE_DIRECTORY', None)
+    BCLAWS_SFTP_HOST = os.getenv('BCLAWS_SFTP_HOST', None)
+    BCLAWS_SFTP_PORT = os.getenv('BCLAWS_SFTP_PORT', None)
+    BCLAWS_SFTP_USERNAME = os.getenv('BCLAWS_SFTP_USERNAME', None)
+    BCLAWS_SFTP_PRIVATE_KEY_ALGORITHM = os.getenv('BCLAWS_SFTP_PRIVATE_KEY_ALGORITHM', 'ED25519')
+    BCLAWS_SFTP_PRIVATE_KEY_PASSPHRASE = os.getenv('BCLAWS_SFTP_PRIVATE_KEY_PASSPHRASE', None)
+    BCLAWS_SFTP_PRIVATE_KEY = os.getenv('BCLAWS_SFTP_PRIVATE_KEY', None)
```
leodube-aot, line 92: These will also need to be added to the dev's `.env` file. For local testing with the [sftp-server repo](https://github.com/leodube-aot/sftp-server), these are:

- `BCLAWS_SFTP_STORAGE_DIRECTORY`= The `SFTP_STORAGE_DIRECTORY` from `sftp-server.env`. Make sure the directory exists on the SFTP server.
- `BCLAWS_SFTP_HOST=localhost`
- `BCLAWS_SFTP_PORT=22`
- `BCLAWS_SFTP_USERNAME`= Any username
- `BCLAWS_SFTP_PRIVATE_KEY_ALGORITHM=ED25519`
- `BCLAWS_SFTP_PRIVATE_KEY_PASSPHRASE`= If a passphrase was used when generating the ssh keys
- `BCLAWS_SFTP_PRIVATE_KEY` = The private id key found in `keys/id_ed25519`

leodube-aot:
Outstanding questions:
1. Do we have any clarification on the exact filename format? As far as I know they want it to start with QP_CORP. I added a date to the end of the filename for now.
2. What SFTP folder path do BCLaws want us to upload to? If the directory does not exist, should we create it? What about the possibility of duplicate filenames, do we overwrite or throw an error?

chenhongjing:
> Outstanding questions:
> 
> 1. Do we have any clarification on the exact filename format? As far as I know they want it to start with QP_CORP. I added a date to the end of the filename for now.
> 2. What SFTP folder path do BCLaws want us to upload to? If the directory does not exist, should we create it? What about the possibility of duplicate filenames, do we overwrite or throw an error?

Answer some of them based on my conversations with BCLaws

1. They just need the file starting with QP_CORP, and there's no furthur requirements. I think it's good to add a date to the end of the filename.
2. They have the folder path for us to upload to. We don't need to create it. Besides, they give limited access to their server, just r/w access to specific folder. We are not able to delete anything.
3. I'm wondering about the possibility of duplicate filenames. In your question, is it that we happen to upload the same file twice or that just the files have the same name?

chenhongjing:
Is your SFTP server repo a private repo? 😄

leodube-aot:
@chenhongjing Thanks for the clarification on 1 and 2, I'll keep the filename as it is for now.

For 3 I don't think it will be an issue if we include the date in the filename. I was mainly thinking if for some reason a new/different xml payload was uploaded with a duplicate filename it could cause some issues. Realistically the job will only run once a day however.

> Is your SFTP server repo a private repo? 😄 

Oops that would be useful haha! I've made it public.

jobs/furnishings/src/furnishings/stage_processors/post_processor.py
```diff
@@ -152,10 +87,124 @@ def get_info_by_name(name: Furnishing.FurnishingName) -> dict:
...
+        self._app = app
+        self._furnishings_dict = furnishings_dict
+        self._processed_date = LegislationDatetime.now()
+
+        # setup the sftp connection objects
+        self._bclaws_sftp_connection = SftpConnection(
+            username=app.config.get('BCLAWS_SFTP_USERNAME'),
+            host=app.config.get('BCLAWS_SFTP_HOST'),
+            port=app.config.get('BCLAWS_SFTP_PORT'),
+            private_key=app.config.get('BCLAWS_SFTP_PRIVATE_KEY'),
```
chenhongjing, line 104: A note on passing in the `private_key` value. [PPR's](https://github.com/bcgov/ppr/blob/main/services/document-delivery-service/src/document_delivery_service/service.py#L54) value of private key is encoded. I guess we will do the similar step, but it depends on how we store the values in 1password.

leodube-aot: Yeah I'm not sure how PPR was storing the private key value, but when I included the decoding code it was giving me `binascii.Error: Incorrect padding`. Probably just a matter of how I've been saving the private key in the `.env` file. Something to keep in mind though for when we add it to 1pass.

chenhongjing: I got the same error when I was testing it. I had to base64-encode the private key (though the string is already encrypted) and load it directly from the encoded file. Then the error was gone.

leodube-aot: Not sure I follow. Are you saying you set the `.env.BCLAWS_SFTP_PRIVATE_KEY ` variable to the private key file path? And also base64-encode the contents of the private key file?

chenhongjing: No, `BCLAWS_SFTP_PRIVATE_KEY` is the a base64-encoded content of the private key.

I did this, and it worked.
```python
# before it, run `base64 private_key > private_key_base64`
with open (PRIVATE_KEY_PATH, 'rb') as private_key_file:
    private_key = private_key_file.read()

sftp = SftpConnection(
    username=USERNAME,
    host=HOST,
    port=PORT,
    private_key=base64.b64decode(private_key).decode('utf-8'),
    private_key_algorithm=PRIVATE_KEY_ALGORITHM
)
```

chenhongjing: hmmm...I think so, but it depends. We can leave it for now. Once we get the key and other configs setup, we can go back to make a change.

jobs/furnishings/src/furnishings/config.py
```diff
@@ -125,6 +143,12 @@ class TestConfig(_Config):  # pylint: disable=too-few-public-methods
         name=DB_NAME,
     )
 
+    # BCLaws SFTP
+    BCLAWS_SFTP_STORAGE_DIRECTORY = 'bclaws'
+
+    # BCMail+ SFTP
+    BCMAIL_SFTP_STORAGE_DIRECTORY = 'bcmail'
```
chenhongjing, line 150: It's better to get these 2 directories from `.env` because we need to set up them later.

jobs/furnishings/tests/unit/test_sftp.py
leodube-aot, line 1: These sftp unit tests use [`pytest-sftpserver`](https://pypi.org/project/pytest-sftpserver/), a local SFTP-server fixture, which serves content from python-like objects. In general, when setting it up for a test, it looks like:

```python
with sftpserver.serve_content({"directory": {"filename.txt": "file content"}}):
   pass
```

One thing to consider is that since it is a python object, multiple directories in the same string do not work.
```python
with sftpserver.serve_content({"dir1/dir2": {"filename.txt": "file content"}}):
        with sftpconnection as sftpclient:
            with sftpclient.open(path.join("dir1", "dir2", "filename.txt")) as file:
                 # will fail

with sftpserver.serve_content({"dir1": {"dir2": {"filename.txt": "file content"}}}):
        with sftpconnection as sftpclient:
            with sftpclient.open(path.join("dir1", "dir2", "filename.txt")) as file:
                 # will work
```

jobs/furnishings/src/furnishings/config.py
leodube-aot: These directories in `TestConfig` are only used in the unit tests, and never actually connect to an SFTP server, so we don't need to set these directories up at all. The unit tests use `pytest-sftpserver`, which is an in-memory sftp "server" for pytests. Since it requires no setup and is not used outside of the unit tests I still think it might be better to keep these values hard-coded (like how `legal-api` has some hard-coded values in `TestConfig`).

There are also a couple quirks with using `pytest-sftpserver` that I go into more detail in [this comment](https://github.com/bcgov/lear/pull/2964#discussion_r1750624263).

chenhongjing: You're correct and thanks for explanation! 😄 I mistook these in test config for the ones in base config.

chenhongjing (approved):
LGTM!

JazzarKarim (approved):
👍 👍 👍

PR #2973, opened 2024-09-09
23054 resolution date in amlagamationApplication (Horizontal & Vertical) document

vysakh-menon-aot:
*Issue #:* /bcgov/entity#23054

*Description of changes:*


legal-api/src/legal_api/services/authz.py
```diff
@@ -333,14 +333,14 @@ def get_allowable_filings_dict():
                     'fullRestoration': {
                         'legalTypes': ['BC', 'BEN', 'CC', 'ULC', 'C', 'CBEN', 'CUL', 'CCC'],
                         'blockerChecks': {
-                            'invalidStateFilings': ['continuationIn', 'continuationOut'],
```
vysakh-menon-aot, line 336: `continuationIn` shouldn't be here

severinbeauvais: Just so I'm clear, this is to prevent a full restoration if the last filing in the ledger is a continuation in application, which can't happen if this business needs a full restoration (ie, is historical), right?

Does a continuation in application count as a state filing anyway?

legal-api/tests/unit/services/test_authorization.py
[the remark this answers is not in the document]
vysakh-menon-aot: nope. we don't test templates

legal-api/src/legal_api/services/authz.py
vysakh-menon-aot: Right. We don't count `continuationIn` as a state filing

ketaki-deodhar (approved):
Looks good

legal-api/tests/unit/services/test_authorization.py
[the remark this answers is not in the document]
severinbeauvais: OK.

The screenshot looks great :)

severinbeauvais (approved):

PR #2989, opened 2024-09-18
22499 Furnishings Job - Update private key decoding

leodube-aot:

*Description of changes:*
- Updated to decode the sftp private key from base 64
- Updated vaults.json
- Confirmed it uploaded to local sftp correctly.

### DIFF
--- a/jobs/furnishings/src/furnishings/config.py
+++ b/jobs/furnishings/src/furnishings/config.py
@@ -145,6 +145,7 @@ class TestConfig(_Config):  # pylint: disable=too-few-public-methods
 
     # BCLaws SFTP
     BCLAWS_SFTP_STORAGE_DIRECTORY = 'bclaws'
+    BCLAWS_SFTP_PRIVATE_KEY = ''
 
     # BCMail+ SFTP
     BCMAIL_SFTP_STORAGE_DIRECTORY = 'bcmail'
--- a/jobs/furnishings/src/furnishings/stage_processors/post_processor.py
+++ b/jobs/furnishings/src/furnishings/stage_processors/post_processor.py
@@ -98,6 +98,7 @@ class PostProcessor:
         self._app = app
         self._furnishings_dict = furnishings_dict
         self._processed_date = LegislationDatetime.now()
+        self._disable_bclaws_sftp = flags.is_on('disable-dissolution-sftp-bclaws')
 
         # setup the sftp connection objects
         self._bclaws_sftp_connection = SftpConnection(
@@ -166,8 +167,8 @@ class PostProcessor:
         self._app.logger.debug('Formatted furnishing details presented in XML file')
 
         # Skip rest of processing if sftp is disabled
-        if flag_on := flags.is_on('disable-dissolution-sftp-bclaws'):
-            self._app.logger.debug(f'disable-dissolution-sftp-bclaws flag on: {flag_on}')
+        if self._disable_bclaws_sftp:
+            self._app.logger.debug(f'disable-dissolution-sftp-bclaws flag on: {self._disable_bclaws_sftp}')
             return
 
         # SFTP to BC Laws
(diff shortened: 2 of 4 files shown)


jobs/furnishings/src/furnishings/stage_processors/post_processor.py
```diff
@@ -103,7 +104,7 @@ def __init__(self, app, furnishings_dict):
             username=app.config.get('BCLAWS_SFTP_USERNAME'),
             host=app.config.get('BCLAWS_SFTP_HOST'),
             port=app.config.get('BCLAWS_SFTP_PORT'),
-            private_key=app.config.get('BCLAWS_SFTP_PRIVATE_KEY'),
+            private_key=base64.b64decode(app.config.get('BCLAWS_SFTP_PRIVATE_KEY')).decode('utf-8'),
```
leodube-aot, line 107: To encode the private key for the `.env` file, I ran the following command on the generated private key file.
```
openssl base64 -A -in keys/id_ed25519 -out keys/id_ed25519_encoded
```

This resulted in a single line encoded string, which I copied into the `.env`. I also updated the `Makefile` in my sftp-server repo to [include these commands](https://github.com/leodube-aot/sftp-server/blob/main/Makefile#L10-L12).

leodube-aot:
Just noticed the failing tests, fixing rn

JazzarKarim (approved):
@chenhongjing This should fix your issue hopefully 🤞

jobs/furnishings/src/furnishings/config.py
```diff
@@ -145,6 +145,7 @@ class TestConfig(_Config):  # pylint: disable=too-few-public-methods
 
     # BCLaws SFTP
     BCLAWS_SFTP_STORAGE_DIRECTORY = 'bclaws'
+    BCLAWS_SFTP_PRIVATE_KEY = ''
```
leodube-aot, line 148: In `post_processor` the `base64.b64decode` expects a string value. If it is not set by the `.env` it will receive a NoneType, throwing an error. I set an empty string as for the `TestConfig` to prevent this.

jobs/furnishings/tests/unit/stage_processors/test_post_processor.py
```diff
@@ -79,10 +79,11 @@ def test_processor(app, session, sftpserver, sftpconnection, test_name, furnishi
     processor = PostProcessor(app, furnishing_dict)
     with sftpserver.serve_content({app.config.get('BCLAWS_SFTP_STORAGE_DIRECTORY'): {}}):
         with patch.object(processor, '_bclaws_sftp_connection', new=sftpconnection):
-            processor.process()
-            # assert xml file is uploaded
-            with sftpconnection as sftpclient:
-                assert len(sftpclient.listdir(app.config.get('BCLAWS_SFTP_STORAGE_DIRECTORY'))) == 1
+            with patch.object(flags, 'is_on', return_value=False):
```
leodube-aot, line 82: Probably not the best way to do this. I'm basically trying to make sure the `post_processor` runs for this test even if the feature flag is preventing it. 

I could potentially refactor the `process` function so that it takes `flags` as a parameter (like `processor.process(flags=...)`) to make it easier to test.

Any thoughts?

JazzarKarim: If it's not too much work, I think refactoring the function would be awesome.

Otherwise, I'm OK with keeping it as is.

chenhongjing: It's ok to use patch method. And another way to make the `post_processor` run in debugging/test is to add the flag value you want into `flags.json` file

leodube-aot: I'll give the flags.json a try! The only reason I'd like to avoid using patch here is that it'll return False for any flag value. So in the future if we add other flags this patch might not work.

chenhongjing (approved):

PR #2992, opened 2024-09-19
23413 put the email msg back on the queue if notify-api fails

vysakh-menon-aot:
*Issue #:* /bcgov/entity#23413

*Description of changes:*

### DIFF
--- a/queue_services/entity-emailer/src/entity_emailer/worker.py
+++ b/queue_services/entity-emailer/src/entity_emailer/worker.py
@@ -114,7 +114,7 @@ def send_email(email: dict, token: str):
         )
         if resp.status_code != HTTPStatus.OK:
             raise EmailException
-    except Exception as err:  # noqa B902; pylint: disable=W0703; we don't want to fail out the email, so ignore all.
+    except Exception:  # noqa B902; pylint: disable=W0703; we don't want to fail out the email, so ignore all.
         # this should log the error and put the email msg back on the queue
         raise EmailException('Unsuccessful response when sending email.')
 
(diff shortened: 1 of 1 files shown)


queue_services/entity-emailer/src/entity_emailer/worker.py
```diff
@@ -103,15 +103,18 @@ def send_email(email: dict, token: str):
...
+            f'{APP_CONFIG.NOTIFY_API_URL}',
+            json=email,
+            headers={
+                'Content-Type': 'application/json',
+                'Authorization': f'Bearer {token}'
+            }
+        )
+        if resp.status_code != HTTPStatus.OK:
+            raise EmailException
+    except Exception:  # noqa B902; pylint: disable=W0703; we don't want to fail out the email, so ignore all.
```
severinbeauvais, line 117: Any idea what exception occurs in the ticket scenario? Maybe we should check for that specifically?

Or, maybe, continue only if the response status code is "OK"?

vysakh-menon-aot: Its max retry error https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/23413#issuecomment-2362276393
But we should any unexpected error

severinbeauvais: I agree that we should handle any error. I guess this is already the cleanest code to do that.

severinbeauvais (approved):

PR #2970, opened 2024-09-06
Experimental code testing new db versioning + toggling between old/new db versioning via feature flag

argush3:
*Issue #:* /bcgov/entity###

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #2997, opened 2024-09-24
23171 Experimental code for updating db versioning

chenhongjing:
**This continues the work in #2970 as part 2. The rough experiemental code is intended to be used as reference _ONLY_ for db versioning spike work. The detailed implementation in the next step should be considered more carefully.**

*Issue #:* /bcgov/entity#23171

*Description of changes:*
- Tweak the new db versioning in part 1, changing the timing to trigger versioning stuff
- Tweak the code for toggling between old and new versioning in part 1, using another structure
- Misc updates to make sure initialization works correctly
- Misc updates to allow toggling at component level
- Misc updates to fix issues occurred during testing
- Test more scenarios via endpoint and queue listener. These include:
    - insert single record
    - insert multiple records
    - delete single record
    - delete multiple records
    - update single record
    - update multiple records
    - multiple add + flush operations
    - retrive versioned data
    - retrive transaction id manually
- Confirmed the scenarios tested work for multiple operations and switchings. These include:
    - start with old -> do operations -> switch to new versioning -> do operations -> switch back -> do operations
    - start with new -> do operations -> switch to old versioning -> do operations -> switch back -> do operations
    - switch back and forth between old and new versioning randomly
- Confirmed the scenarios tested work at component level. These include:
    - turn new versioning FF on in legal-api + turn FF off in filer
    - turn new versioning FF off in legal-api + turn FF on in filer
    - switch flag back and forth in legal-api while keeping filer unchanged
    - switch flag back and forth in filer while keeping legal-api unchanged

The main changes are in `legal-api/src/legal_api/models/db.py` and `python/common/sql-versioning/sql_versioning/versioned_history.py`.

Please check my following comments for detailed explanations.


python/common/sql-versioning/sql_versioning/versioned_history.py
```diff
@@ -0,0 +1,256 @@
...
+def _after_flush(session, flush_context):
+    print("Entering _after_flush")
+    try:
+        for obj in session.new.union(session.dirty).union(session.deleted):
+            if isinstance(obj, Versioned):
+                operation_type = 'I' if obj in session.new else 'U' if obj in session.dirty else 'D'
+                create_version(session, obj, operation_type)
+    except Exception as e:
+        print(f"Error in new.after_flush: {str(e)}")
+    print("Exiting _after_flush")
```
chenhongjing, line 194: I keep most of the experimental code in part1 for new versioning module. The big change is the timing to write versioned record into db - moving it from `before_flush` to `after_blush` to support more scenarios than just updating single record.

legal-api/src/legal_api/models/db.py
```diff
@@ -11,17 +11,206 @@
...
+    @event.listens_for(SignallingSession, 'after_transaction_end')
+    @debug
+    def clear_transaction(session, transaction):
+        VersioningProxy.unlock_versioning(session, transaction)
+
+    _enable_new_versioning()
+    make_versioned(user_cls=None)
+
+
+setup_versioning()  # it should be called before data model initialzed, otherwise, old versioning doesn't work properly
```
chenhongjing, line 216: I did an experiement to enable/disable `sqlalchemy-continuum` by using its functions `make_versioned()/remove_versioning()` during the app run. And it failed.

It turns out that `continuum` should be initialized before code data models are initialized, otherwise the versioning won't work.

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -212,3 +213,171 @@ def search_businesses():
         current_app.logger.info(err)
         current_app.logger.error('Error searching over business information for: %s', identifiers)
         return {'error': 'Unable to retrieve businesses.'}, HTTPStatus.INTERNAL_SERVER_ERROR
+
+
+# experimental debugging endpoints
```
chenhongjing, line 218: The following endpoints are for scenario testings. Singular form (e.g. `insert`) indicates that only one record is modified and plural form (e.g. `inserts`) indicate that multiple FK related records are updated.

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -212,3 +213,171 @@ def search_businesses():
...
+    """
+    business = Business.find_by_internal_id(business_id)
+    db.session.delete(business)
+    db.session.commit()
+    return jsonify({'success': True, 'id': business_id})
+
+
+@bp.route('/db_versioning_test/flush', methods=['GET'])
+@cross_origin(origins='*')
+def db_versioning_test_flush():
```
chenhongjing, line 321: This is an extra scenario to test several `add` and `flush` operations before final `commit`. The multiple `add` and `flush` operations are all in one single transaction and should not result in multiple transaction records.

legal-api/src/legal_api/models/db.py
```diff
@@ -11,17 +11,206 @@
...
+    uow = versioning_manager.unit_of_work(session)
+    transaction = uow.create_transaction(session)
+    return transaction.id
+
+def _new_get_transaction_id(session):
+    new_transaction_manager = TransactionManager(session)
+    return new_transaction_manager.create_transaction()
+
+
+class VersioningProxy:
```
chenhongjing, line 78: I change to use proxy instead of custom db session to dispatch two versionings. Hope it will be clearer.

python/common/sql-versioning/sql_versioning/versioned_history.py
```diff
@@ -0,0 +1,256 @@
...
+        if not hasattr(cls, '_version_cls'):
+            class_name = f'{cls.__name__}Version'
+            table_name = f'{cls.__tablename__}_version'
+
+            attrs = {
+                '__tablename__': table_name,
+                'id': Column(Integer, primary_key=True),
+                'transaction_id': Column(BigInteger, primary_key=True, nullable=False),
+                'end_transaction_id': Column(BigInteger, nullable=True),
+                'operation_type': Column(SmallInteger, nullable=False),
```
chenhongjing, line 75: Should use a composite primary key (`id` + `transaction_id`). The types used are also consistent with those in `continuum `.

legal-api/src/legal_api/models/db.py
```diff
@@ -11,17 +11,200 @@
...
+
+        current_versioning = session.info['_versioning_locked']
+        print(f'\033[31mCurrent versioning type={current_versioning}\033[0m')
+
+        return cls._versioning_control[current_versioning]['version_class'](obj)
+
+
+@debug
+def setup_versioning():
+    @event.listens_for(SignallingSession, 'after_transaction_create')  # skip events for continuum's internal session/txn operations
```
chenhongjing, line 196: `SignallingSession` is the default session that our flask app uses and it extends the core session. I use it to distinguish the session that `continuum` uses and want to make it easier to track transaction lifecycle. `continuum` uses core session for its internal operations. It creates transaction in its own session and detaches it from its session and add it to the flask session.

queue_services/entity-filer/requirements.txt
```diff
@@ -26,4 +26,4 @@ reportlab==3.6.12
 git+https://github.com/bcgov/sbc-connect-common.git#egg=gcp-queue&subdirectory=python/gcp-queue
 git+https://github.com/bcgov/business-schemas.git@2.18.27#egg=registry_schemas
 git+https://github.com/bcgov/lear.git#egg=entity_queue_common&subdirectory=queue_services/common
-git+https://github.com/bcgov/lear.git#egg=legal_api&subdirectory=legal-api
+# git+https://github.com/bcgov/lear.git#egg=legal_api&subdirectory=legal-api
```
chenhongjing, line 29: If you want to play around this experimental code, you need to manually install legal-api and sql-versioning by `pip install -e <path>` (or build the wheels and install it)

chenhongjing:
@argush3 @vysakh-menon-aot @leodube-aot @JazzarKarim @kzdev420 @AimeeGao @eason-pan-bc 👀 
Same as Argus' previous pr, it's not really for review but to keep everyone in the loop. But I did change lots of stuff based off his work. Let me know if you have any questions or comments. Please feel free to share any opinions and have discussions here. 😺

legal-api/src/legal_api/resources/v2/business/business.py
argush3: thanks for the thorough testing

vysakh-menon-aot:
Note: A version should be created only from entity-filer when we retrieve transaction_id `VersioningProxy.get_transaction_id(db.session)`. This is to map a version row (transaction id) to a filing.

chenhongjing:
> Note: A version should be created only from entity-filer when we retrieve transaction_id `VersioningProxy.get_transaction_id(db.session)`. This is to map a version row (transaction id) to a filing.

hmmm....true but not completely true if I understand what you mean well. Some other services will use model's `save()` method (e.g. use `business.save()` after updating some values) and this will cut a new versioning row even using existing versioning method 💭

argush3:
> > Note: A version should be created only from entity-filer when we retrieve transaction_id `VersioningProxy.get_transaction_id(db.session)`. This is to map a version row (transaction id) to a filing.
> 
> hmmm....true but not completely true if I understand what you mean well. Some other services will use model's `save()` method (e.g. use `business.save()` after updating some values) and this will cut a new versioning row even using existing versioning method 💭

I think this almost never happens outside of the filer. 

And if it happens, it's saves done on tables that don't have a versioning table.

Are you able to find an example?

chenhongjing:
> > > Note: A version should be created only from entity-filer when we retrieve transaction_id `VersioningProxy.get_transaction_id(db.session)`. This is to map a version row (transaction id) to a filing.
> > 
> > 
> > hmmm....true but not completely true if I understand what you mean well. Some other services will use model's `save()` method (e.g. use `business.save()` after updating some values) and this will cut a new versioning row even using existing versioning method 💭
> 
> I think this almost never happens outside of the filer.
> 
> And if it happens, it's saves done on tables that don't have a versioning table.
> 
> Are you able to find an example?

yes. just take `business.save()` as an example.

https://github.com/bcgov/lear/blob/main/jobs/email-reminder/email_reminder.py#L171
https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/resources/v2/internal_services.py#L53
https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/resources/v2/business/colin_sync.py#L274

chenhongjing:
> > > Note: A version should be created only from entity-filer when we retrieve transaction_id `VersioningProxy.get_transaction_id(db.session)`. This is to map a version row (transaction id) to a filing.
> > 
> > 
> > hmmm....true but not completely true if I understand what you mean well. Some other services will use model's `save()` method (e.g. use `business.save()` after updating some values) and this will cut a new versioning row even using existing versioning method 💭
> 
> I think this almost never happens outside of the filer.
> 
> And if it happens, it's for saves done on tables that don't have a versioning table.
> 
> Are you able to find an example?

Anyway, we need the auto-creation logic like that the current one is doing in the background right now, don't we?

argush3:
> Anyway, we need the auto-creation logic like that the current one is doing in the background right now, don't we?

Yes, I think this is fine then as we are already doing this.  I'm assuming the continuum versioning just auto creates transactions used for the versioning records in the `business.save()` scenarios you linked.

vysakh-menon-aot:
Yes,. I verified, it does create in the background. We may need to verify the get version data has no impact due to this

PR #2998, opened 2024-09-24
23034 add the config

kzdev420:
*Issue #:* /bcgov/entity#23034

*Description of changes:*

### DIFF
--- a/queue_services/entity-emailer/src/entity_emailer/config.py
+++ b/queue_services/entity-emailer/src/entity_emailer/config.py
@@ -74,7 +74,6 @@ class _Config():  # pylint: disable=too-few-public-methods
     PAY_API_URL = os.getenv('PAY_API_URL', None)
     AUTH_URL = os.getenv('AUTH_URL', None)
     ACCOUNT_SVC_AUTH_URL = os.getenv('ACCOUNT_SVC_AUTH_URL', None)
-    AUTH_SVC_URL = os.getenv('AUTH_URL', 'https://')
     # secrets
     ACCOUNT_SVC_CLIENT_ID = os.getenv('ACCOUNT_SVC_CLIENT_ID', None)
     ACCOUNT_SVC_CLIENT_SECRET = os.getenv('ACCOUNT_SVC_CLIENT_SECRET', None)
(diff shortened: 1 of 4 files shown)


chenhongjing (commented):
Does emailer need it? and the `entity-service-account` should be added to some vaults.json too.

queue_services/entity-emailer/src/entity_emailer/config.py
```diff
@@ -74,6 +74,7 @@ class _Config():  # pylint: disable=too-few-public-methods
     PAY_API_URL = os.getenv('PAY_API_URL', None)
     AUTH_URL = os.getenv('AUTH_URL', None)
     ACCOUNT_SVC_AUTH_URL = os.getenv('ACCOUNT_SVC_AUTH_URL', None)
+    AUTH_SVC_URL = os.getenv('AUTH_URL', 'https://')
```
vysakh-menon-aot, line 77: we have it in line 75

kzdev420: Yes, we have it. But, the emailer import the legal-api.service that use as AUTH_SVC_URL not AUTH_URL
https://github.com/bcgov/lear/blob/881ad4d19940bd0f7e119d4fe9e19a55d08a2582/legal-api/src/legal_api/services/bootstrap.py#L282

kzdev420:
> Does emailer need it? and the `entity-service-account` should be added to some vaults.json too.

Currently, the emailer doesn't use it yet. but I thought it can be used in the future.

chenhongjing:
> > Does emailer need it? and the `entity-service-account` should be added to some vaults.json too.
> 
> Currently, the emailer doesn't use it yet. but I thought it might be used in the future.

You should remove it since it's duplicate with previous line.

queue_services/entity-emailer/src/entity_emailer/config.py
kzdev420: removed

kzdev420:
> > > Does emailer need it? and the `entity-service-account` should be added to some vaults.json too.
> > 
> > 
> > Currently, the emailer doesn't use it yet. but I thought it might be used in the future.
> 
> You should remove it since it's duplicate with previous line.

Removed

chenhongjing (approved):
Make sure you have added value to secrets & cronjob yaml files in OCP before it's merged.

PR #2999, opened 2024-09-25
22585 - update allowable actions for amalgamations

ketaki-deodhar:
*Issue #:* /bcgov/entity#22585

*Description of changes:*

- Updated allowable actions for Amalgamations (staff, general user). Amalgamation will not allowed when there is pending draft for the business


severinbeauvais (approved):
If I understand this correctly, lgtm :)

I'm kind of curious why the tests don't directly match the code changes. Can you confirm that the new BusinessBlocker changes are being tested?

severinbeauvais:
Also, can you please have a look at the linting issues? I don't know if you caused any of them but it would be good to fix them all.

ketaki-deodhar:
> If I understand this correctly, lgtm :)
> 
> I'm kind of curious why the tests don't directly match the code changes. Can you confirm that the new BusinessBlocker changes are being tested?

Amalagamtions were removed from allowed filings from expected parameter. 
image

Actual function is called in the test is below. 
image

ketaki-deodhar:
> Also, can you please have a look at the linting issues? I don't know if you caused any of them but it would be good to fix them all.

I don't think these changes caused it. I looked into it but somehow cannot see these when run pylint locally

ArwenQin (approved):

legal-api/src/legal_api/services/authz.py
```diff
@@ -170,29 +170,32 @@ def get_allowable_filings_dict():
                     'regular': {
                         'legalTypes': ['BEN', 'BC', 'ULC', 'CC', 'C', 'CBEN', 'CUL', 'CCC'],
                         'blockerChecks': {
-                            'business': [BusinessBlocker.BUSINESS_FROZEN,
+                            'business': [BusinessBlocker.DEFAULT,
                                          BusinessBlocker.NOT_IN_GOOD_STANDING,
-                                         BusinessBlocker.IN_DISSOLUTION],
+                                         BusinessBlocker.IN_DISSOLUTION,
+                                         BusinessBlocker.DRAFT_PENDING],
```
vysakh-menon-aot, line 176: No need to add `DRAFT_PENDING`. DEFAULT covers both DRAFT_PENDING and BUSINESS_FROZEN.

Ref. 
https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/authz.py#L659
https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/authz.py#L663

vysakh-menon-aot (approved):

PR #3008, opened 2024-09-25
23255 - Refactor dissolution business summary display name to support C companies

ketaki-deodhar:
*Issue #:* /bcgov/entity#23255

*Description of changes:*

- Updated `FILING_SUMMARY_DISPLAY_NAME` to fallback to a common name unless there is a specific name

_**C9900778:**_
image

_**FM1019711 (SP) Voluntary Dissolution:**_
image

### DIFF
--- a/legal-api/src/legal_api/reports/business_document.py
+++ b/legal-api/src/legal_api/reports/business_document.py
@@ -567,11 +567,13 @@ class BusinessDocument:
     def _get_summary_display_name(filing_type: str,
                                   filing_sub_type: Optional[str],
                                   legal_type: Optional[str]) -> str:
+        display_name = None
         if filing_type == 'dissolution':
-            if filing_sub_type == 'voluntary' and legal_type in ['SP', 'GP']:
-                return 'Dissolution Application'
+            if legal_type in BusinessDocument.FILING_SUMMARY_DISPLAY_NAME[filing_type][filing_sub_type]:
+                display_name = BusinessDocument.FILING_SUMMARY_DISPLAY_NAME[filing_type][filing_sub_type][legal_type]
             else:
-                return BusinessDocument.FILING_SUMMARY_DISPLAY_NAME[filing_type][filing_sub_type]
+                display_name = BusinessDocument.FILING_SUMMARY_DISPLAY_NAME[filing_type][filing_sub_type]['defaultName']
+            return display_name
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/reports/business_document.py
```diff
@@ -568,7 +568,10 @@ def _get_summary_display_name(filing_type: str,
                                   filing_sub_type: Optional[str],
                                   legal_type: Optional[str]) -> str:
         if filing_type == 'dissolution':
-            return BusinessDocument.FILING_SUMMARY_DISPLAY_NAME[filing_type][filing_sub_type][legal_type]
+            if filing_sub_type == 'voluntary' and legal_type in ['SP', 'GP']:
+                return 'Dissolution Application'
+            else:
+                return BusinessDocument.FILING_SUMMARY_DISPLAY_NAME[filing_type][filing_sub_type]
```
vysakh-menon-aot, line 574: This will work but it not extensible. Can you try a fallback method.
Ref:
https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/reports/report.py#L236
https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/reports/report.py#L1420

ketaki-deodhar: Okay. Will do

legal-api/src/legal_api/reports/business_document.py
```diff
@@ -588,25 +593,40 @@ def _get_legal_type_description(legal_type: str) -> str:
                 'BEN': 'Voluntary Dissolution',
                 'ULC': 'Voluntary Dissolution',
                 'CC': 'Voluntary Dissolution',
+                'C': 'Voluntary Dissolution',
+                'CBEN': 'Voluntary Dissolution',
+                'CUL': 'Voluntary Dissolution',
+                'CCC': 'Voluntary Dissolution',
```
ketaki-deodhar, line 599: I added all C types as was not sure if to let it fall back on default name. I can remove it if you think that is the way to go

legal-api/src/legal_api/reports/business_document.py
vysakh-menon-aot: Will refactor to an extensible code when we work on this later

vysakh-menon-aot (approved):

PR #3010, opened 2024-09-26
23354 - Misc updates for Notice of Withdrawal

chenhongjing:
*Issue #:* /bcgov/entity#23354

*Description of changes:*
- add filing type
- add fee codes
- add filing status
- add tests

Let me know if I'm missing something

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #3013, opened 2024-09-26
Legal API Linting Fix

chenhongjing:
*Issue #:* /bcgov/entity#23354

*Description of changes:*

This issue is mentioned in #3010. Trying to provide a temporary solution for it since `too-many-positional-arguments` is only available after version 3.3.0 and directly disabling it globally or locally will make pylint 3.2.x throw errors.

- make legal-api linting work for both pylint 3.2.x and pylint 3.3.x

### DIFF
--- a/legal-api/Makefile
+++ b/legal-api/Makefile
@@ -64,7 +64,15 @@ install-dev: ## Install local application
 ci: lint flake8 test ## CI flow
 
 pylint: ## Linting with pylint
-	. venv/bin/activate && pylint --rcfile=setup.cfg  --load-plugins=pylint_flask src/$(PROJECT_NAME)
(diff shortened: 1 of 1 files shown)


vysakh-menon-aot:
Did you try updating in setup.cfg? Is that what you mean you get error?

chenhongjing:
> Did you try updating in setup.cfg instead of this? Is that what you mean you get error?

Yes, it showed errors for pylint 3.2.x. and python 3.3.x supports python 3.9 and above.

kzdev420 (approved):

eason-pan-bc (approved):

AimeeGao (approved):
LGTM (^-^)

PR #3014, opened 2024-09-27
23348 - Add allowable actions for notice of withdrawal

AimeeGao:
*Issue #:* /bcgov/entity#23348

*Description of changes:*

- Added Allowable action item for Notice of Withdrawal, restricted to staff only
- Implemented logic for applicable company types (BC, BEN, CC & ULC + their Continued In versions, C, CCC, CBEN & CUL)
- Added checks to prevent filing when:
   - Business doesn't have FED filings
   - Business is frozen
   - Business has a pending or draft filing
- Updated unit tests

image

**Note**
This implementation only covers the Notice of Withdrawal logic for existing businesses. The temporary business scenario is currently excluded.


legal-api/src/legal_api/services/authz.py
```diff
@@ -648,7 +656,8 @@ def business_blocker_check(business: Business, is_ignore_draft_blockers: bool =
         BusinessBlocker.DRAFT_PENDING: False,
         BusinessBlocker.NOT_IN_GOOD_STANDING: False,
         BusinessBlocker.AMALGAMATING_BUSINESS: False,
-        BusinessBlocker.IN_DISSOLUTION: False
+        BusinessBlocker.IN_DISSOLUTION: False,
+        BusinessBlocker.NO_FUTURE_EFFECTIVE_FILING: True
```
vysakh-menon-aot, line 660:
```suggestion
        BusinessBlocker.NO_FUTURE_EFFECTIVE_FILING: False
```

legal-api/src/legal_api/services/authz.py
```diff
@@ -671,6 +680,9 @@ def business_blocker_check(business: Business, is_ignore_draft_blockers: bool =
     if business.in_dissolution:
         business_blocker_checks[BusinessBlocker.IN_DISSOLUTION] = True
 
+    if has_any_future_effective_filing(business):
+        business_blocker_checks[BusinessBlocker.NO_FUTURE_EFFECTIVE_FILING] = False
```
vysakh-menon-aot, line 684:
```suggestion
        business_blocker_checks[BusinessBlocker.NO_FUTURE_EFFECTIVE_FILING] = True
```

legal-api/src/legal_api/services/authz.py
```diff
@@ -753,6 +765,13 @@ def has_blocker_future_effective_filing(business: Business, blocker_checks: dict
     return is_fed
 
 
+def has_any_future_effective_filing(business: Business):
+    """Return True if the business has any future effective filing."""
+    now = datetime.utcnow().replace(tzinfo=timezone.utc)
+    pending_filings = Filing.get_filings_by_status(business.id, [Filing.Status.PENDING.value, Filing.Status.PAID.value])
```
vysakh-menon-aot, line 771:
```suggestion
    now = datetime.now(timezone.utc)
    pending_filings = Filing.get_filings_by_status(business.id, [Filing.Status.PAID.value])
```

vysakh-menon-aot: Before removing PENDING from the query, confirm non-paid filing has to be considered or not

AimeeGao: Many thanks, Vysakh!

legal-api/src/legal_api/services/authz.py
```diff
@@ -314,6 +315,13 @@ def get_allowable_filings_dict():
                             'business': [BusinessBlocker.DEFAULT]
                         }
                     }
+                },
+                'noticeOfWithdrawal': {
+                    'legalTypes': ['BC', 'BEN', 'CC', 'ULC', 'C', 'CBEN', 'CUL', 'CCC'],
+                    'blockerChecks': {
+                        'business': [BusinessBlocker.BUSINESS_FROZEN,
+                                     BusinessBlocker.NO_FUTURE_EFFECTIVE_FILING]
```
chenhongjing, line 323: There's a scenario that I think you should think of...A business has a FED filing and not in frozen status but it also has a draft NoW filing, in this case, it should be not allowed to start another NoW filing again.

chenhongjing (commented):
You should make this filing support for both temporary business(new) and existing business. 

References: 
[allowable actions for amalgamataion application](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/authz.py#L169)
[business requirement check](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/authz.py#L560-L564)
[flow to save filing](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py#L473)

legal-api/tests/unit/services/test_authorization.py
```diff
@@ -2674,6 +2693,172 @@ def mock_auth(one, two):  # pylint: disable=unused-argument; mocks of library me
...
+
+    def mock_auth(one, two):  # pylint: disable=unused-argument; mocks of library methods
+        return headers[one]
+
+    with app.test_request_context():
+        monkeypatch.setattr('flask.request.headers.get', mock_auth)
+
+        for legal_type in legal_types:
+            identifier = (f'BC{random.SystemRandom().getrandbits(0x58)}')[:9]
+            business = factory_business(identifier=identifier,
```
chenhongjing, line 2840: It will be better if you can add some tests for temporary business scenarios. 💭

chenhongjing (approved):
As per our conversations with Argus, there should be extra design and investigation on the scenarios of temporary businesses. As the priorities are shifting, we may hold off further NoW work. So now we only handle allowable actions for existing businesses, and it will be good if you add some comments on the spreadsheet, and leave notes for QA. Overall, good work!

eason-pan-bc (approved):
LGTM 😄

AimeeGao:
> As per our conversations with Argus, there should be extra design and investigation on the scenarios of temporary businesses. As the priorities are shifting, we may hold off further NoW work. So now we only handle allowable actions for existing businesses, and it will be good if you add some comments on the spreadsheet, and leave notes for QA. Overall, good work!

Thank you, Hongjing, for your detailed explanation and reminder. I’ve added comments for the Notice of Withdrawal in the spreadsheet, and I will also leave relevant notes for QA.

PR #3021, opened 2024-10-07
23539 add batch_processing to colin_event_ids table

kzdev420:
*Issue #:* /bcgov/entity#23539

*Description of changes:*

Screenshot 2024-10-07 at 9 20 21 AM

### DIFF
--- a/legal-api/migrations/versions/83a110e10979_add_batch_processing_id_to_colin_event_.py
+++ b/legal-api/migrations/versions/83a110e10979_add_batch_processing_id_to_colin_event_.py
@@ -18,7 +18,7 @@ depends_on = None
 
 def upgrade():
     op.add_column('colin_event_ids', sa.Column('batch_processing_id', sa.Integer(), nullable=True))
-    op.add_column('colin_event_ids', sa.Column('batch_processing_step', sa.String(length=1000), nullable=True))
+    op.add_column('colin_event_ids', sa.Column('batch_processing_step', sa.Enum('WARNING_LEVEL_1', 'WARNING_LEVEL_2', 'DISSOLUTION', name='batch_processing_step'), nullable=True))
     op.create_foreign_key('colin_event_ids_batch_processing_id_fkey', 'colin_event_ids', 'batch_processing', ['batch_processing_id'], ['id'])
 
 
@@ -26,3 +26,4 @@ def downgrade():
     op.drop_constraint('colin_event_ids_batch_processing_id_fkey', 'colin_event_ids', type_='foreignkey')
     op.drop_column('colin_event_ids', 'batch_processing_id')
     op.drop_column('colin_event_ids', 'batch_processing_step')
+    op.execute("DROP TYPE batch_processing_step;")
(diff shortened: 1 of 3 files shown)


argush3:
don't believe you've addressed all todos defined in the ticket description.

image

legal-api/migrations/versions/83a110e10979_add_batch_processing_id_to_colin_event_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '83a110e10979'
+down_revision = 'd55dfc5c1358'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('colin_event_ids', sa.Column('batch_processing_id', sa.Integer(), nullable=True))
+    op.add_column('colin_event_ids', sa.Column('batch_processing_step', sa.String(length=1000), nullable=True))
```
argush3, line 21: i think this should use the existing enum to ensure that `batch_processing` uses so we lock down the values that can be used for this column

kzdev420: Fixed. please kindly review.

legal-api/migrations/versions/83a110e10979_add_batch_processing_id_to_colin_event_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '83a110e10979'
+down_revision = 'd55dfc5c1358'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('colin_event_ids', sa.Column('batch_processing_id', sa.Integer(), nullable=True))
+    op.add_column('colin_event_ids', sa.Column('batch_processing_step', sa.Enum('WARNING_LEVEL_1', 'WARNING_LEVEL_2', 'DISSOLUTION', name='batch_processing_step'), nullable=True))
```
leodube-aot, line 21: Something I found interesting and didn't know about how alembic and sqlalchemy handles re-using enums:

To me it looked like you were re-creating the enum, so I was wondering what would happen if I changed the enum values here, and if it would have any effect on the existing `batch_processing_step` enum.

Screenshot 2024-10-09 at 10 15 58 AM

Turns out this has no effect on the values `batch_processing.batch_processing_step` or `colin_event_id.batch_processing_step`. So alembic must be pointing the `colin_event_id.batch_processing_step` enum to the existing one.

Turns out, you can remove all the enum values here and it still works as long as the `name` parameter points to the correct enum.

Screenshot 2024-10-09 at 10 19 21 AM

Can you update this line to remove the enum values? Just to avoid any confusion as to what this line is doing.

kzdev420: Good Catch.
I removed it and tested it. It works well

argush3 (approved):

PR #3027, opened 2024-10-11
23543 COLIN BE Services - Sync involuntary dissolution stages back to COLIN

leodube-aot:
*Issue #:* /bcgov/entity#23543

*Description of changes:*
- Updated `BatchProcessing` model:
  - Added a static method to return eligible batch processings for COLIN
- Updated Legal API:
  - Created a new GET endpoint to return the eligible batch_processings.
  - Created a new PATCH endpoint to create a `colin_event_id` entry for synced batch processings.
- Updated COLIN API:
  - Created a new POST endpoint to create a CORP_STATE entry and an EVENT entry.
- Updated the COLIN sync job to work with batch processings.
- Created unit tests for legal api related changes.

_Local Testing_
In depth local testing can be found in the [test notes of the ticket](https://app.zenhub.com/workspaces/entities---david-65af11a89e89f5043c2911f6/issues/gh/bcgov/entity/23543).


colin-api/src/colin_api/resources/business.py
```diff
@@ -154,6 +154,7 @@ def get(legal_type, identifier, name_type):
 @cors_preflight('GET')
 @API.route('/internal/<string:info_type>', methods=['GET'])
 @API.route('/internal/<string:legal_type>/<string:identifier>/<string:info_type>', methods=['GET'])
+@API.route('/internal/<string:legal_type>/<string:identifier>', methods=['PATCH'])
```
leodube-aot, line 157: I wasn't entirely sure where the best place was to add this endpoint. I figured since we are updating the corp state and not actually creating a filing, then using an internal PATCH endpoint on the business made the most sense.

jobs/update-colin-filings/update_colin_filings.py
```diff
@@ -144,45 +194,88 @@ def get_bearer_token(app):
...
+            identifier = filing['filing']['business']['identifier']
+            if identifier in corps_with_failed_filing:
+                # pylint: disable=no-member; false positive
+                app.logger.debug(f'Skipping filing {filing_id} for'
+                                 f' {filing["filing"]["business"]["identifier"]}.')
+            else:
+                colin_ids = send_filing(app, token, filing, filing_id)
+                update = None
+                if colin_ids and any(colin_ids):
+                    update = update_colin_id(app, token, filing_id, colin_ids)
```
leodube-aot, line 222: I found that `colin_ids` can return as `[None]`, which actually evaluates to True. We need to make sure to not attempt updating colin ids if they are None, as it will still insert into the `colin_event_ids` table using a default sequence value.

legal-api/src/legal_api/models/batch_processing.py
```diff
@@ -103,3 +128,35 @@ def find_by(cls,  # pylint: disable=too-many-arguments
...
+            )
+        )
+        batch_processings = db.session.query(BatchProcessing).\
+            filter(Batch.id == BatchProcessing.batch_id).\
+            filter(Batch.batch_type == Batch.BatchType.INVOLUNTARY_DISSOLUTION).\
+            filter(BatchProcessing.step.in_([
+                BatchProcessing.BatchProcessingStep.WARNING_LEVEL_1,
+                BatchProcessing.BatchProcessingStep.WARNING_LEVEL_2,
+            ])).\
+            filter(BatchProcessing.status == BatchProcessing.BatchProcessingStatus.PROCESSING).\
```
chenhongjing, line 151: Thinking if it's proper to only select batch processing in `PROCESSING` status...?

jobs/update-colin-filings/update_colin_filings.py
vysakh-menon-aot: Its not expected to receive `[None]`. When did you receive it? any specific filing?

vysakh-menon-aot: For a successful sync we must receive an event id

argush3:
As per our conversation, this work is on hold until we take another look at what really needs to be synced back to COLIN if anything.

jobs/update-colin-filings/update_colin_filings.py
leodube-aot: Yup so I tracked down the source of the issue.

It's caused when an involuntary dissolution filing is passed to colin api without any `meta_data`, or with incorrectly formed `meta_data`. You can see that in [`Filing.add_involuntary_dissolution_event` can return None if this is the case](https://github.com/bcgov/lear/blob/main/colin-api/src/colin_api/models/filing.py#L1178-L1199). This gets [returned as `event_id = None` and saved into the `colinIds`](https://github.com/bcgov/lear/blob/main/colin-api/src/colin_api/resources/filing.py#L166-L174) for the filing (with HTTP status CREATED, even though nothing was created). 

It's probably pretty low risk, as I only encountered it because I made a mistake manually creating local test data.

Thinking more about this, it should probably be fixed in colin api rather than the updater job like I did here. This PR may not get merged (see Argus' comment), but this bug might still be worth fixing anyway.

vysakh-menon-aot: Good catch

legal-api/src/legal_api/models/batch_processing.py
leodube-aot: This ticket is on hold for now, but to still answer your question: yes I think it should only select PROCESSING status. My analysis on the statuses is that:
- `ERROR`, `HOLD`, and `QUEUED` statuses we don't want to to sync since there is some issue or wait associated with the batch processing.
- `COMPLETED` status will always have a dissolution filing. This dissolution filing is already synced correctly (creates and EVENT and sets the CORP_STATE).
- `WITHDRAWN` status should always have an annual report filing. This AR filing sync needs to be verified in https://github.com/bcgov/entity/issues/23545

If we were to sync `COMPLETED` and `WITHDRAWN` status batch processings, we would get duplicates in COLIN since we would be syncing both the batch processing and their associated filings.

leodube-aot:
Closing as no longer a requirement

PR #3028, opened 2024-10-11
23697 - fix stage1 to stage2 dissolution transition bug

eason-pan-bc:
*Issue #:* /bcgov/entity#23697

*Description of changes:*
- Added the `disable-dissolution-sftp-bcmail` feature flag option in `flags.json`
- Updated the logic for furnishing entries check when moving from stage 1 to stage 2 of involuntary dissolution
    - check furnishing entries - 2 scenarios
        1. if email processed, and mail processed (after 5 business days still not in good standing)
        2. only has a mail entry, and it's processed
- Updated the check to support SFTP function FF

#### Local Test Results 
**Action**: Trying to move a business from stage 1 to stage 2 of involuntary dissolution

**Scenario 1** - ERROR and not moving to stage 2
_when this business has an email entry and haven't sent out a mail _
Pasted image 20241011153110
Pasted image 20241011153047

**Scenario 2** - SUCCESS moving to stage 2
_when this business has an email entry and after 5 business days still not in good standing, a mail sent out_
Pasted image 20241011153215
Pasted image 20241011153238

### DIFF
--- a/jobs/furnishings/flags.json
+++ b/jobs/furnishings/flags.json
@@ -1,5 +1,6 @@
 {
     "flagValues": {
-        "enable-involuntary-dissolution": true
+        "enable-involuntary-dissolution": true,
+        "disable-dissolution-sftp-bcmail": true
     }
 }
--- a/jobs/furnishings/flags.json
+++ b/jobs/furnishings/flags.json
@@ -1,6 +1,7 @@
 {
     "flagValues": {
         "enable-involuntary-dissolution": true,
-        "disable-dissolution-sftp-bcmail": true
+        "disable-dissolution-sftp-bcmail": true,
+        "disable-dissolution-sftp-bclaws": false
     }
 }
--- a/jobs/involuntary-dissolutions/involuntary_dissolutions.py
+++ b/jobs/involuntary-dissolutions/involuntary_dissolutions.py
@@ -253,13 +253,13 @@ def stage_2_process(app: Flask):
             for furnishing in furnishings
         )
         # if SFTP function is off, we expect the mail status will be QUEUED
-        expected_mail_status = Furnishing.FurnishingStatus.PROCESSED
+        expected_mail_status = [Furnishing.FurnishingStatus.PROCESSED]
         if flags.is_on('disable-dissolution-sftp-bcmail'):
-            expected_mail_status = Furnishing.FurnishingStatus.QUEUED
+            expected_mail_status.append(Furnishing.FurnishingStatus.QUEUED)
 
         mail_processed = any(
-            furnishing.furnishing_type == Furnishing.FurnishingType.MAIL
-            and furnishing.status == expected_mail_status
+            (furnishing.furnishing_type == Furnishing.FurnishingType.MAIL)
+            and (furnishing.status in expected_mail_status)
             for furnishing in furnishings
         )
         email_exists = any(furnishing.furnishing_type == Furnishing.FurnishingType.EMAIL for furnishing in furnishings)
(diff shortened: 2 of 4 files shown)


jobs/involuntary-dissolutions/involuntary_dissolutions.py
```diff
@@ -239,16 +239,34 @@ def stage_2_process(app: Flask):
...
-            furnishing.furnishing_type in (Furnishing.FurnishingType.EMAIL, Furnishing.FurnishingType.MAIL)
+        email_processed = any(
+            furnishing.furnishing_type == Furnishing.FurnishingType.EMAIL
             and furnishing.status == Furnishing.FurnishingStatus.PROCESSED
             for furnishing in furnishings
         )
+        # if SFTP function is off, we expect the mail status will be QUEUED
+        expected_mail_status = Furnishing.FurnishingStatus.PROCESSED
+        if flags.is_on('disable-dissolution-sftp-bcmail'):
+            expected_mail_status = Furnishing.FurnishingStatus.QUEUED
```
chenhongjing, line 258: We will turn FF on and off for testing purpose, so you should make sure that both `PROCESSED` and `QUEUED` pass the check. For example, the mail furnishing is set processed in stage 1 with disable-FF off, and then we turn this flag on and move it into stage 2 and this is also expected to pass the check.

eason-pan-bc: good call! thanks for pointing it out 👍

jobs/furnishings/flags.json
```diff
@@ -1,5 +1,6 @@
 {
     "flagValues": {
-        "enable-involuntary-dissolution": true
+        "enable-involuntary-dissolution": true,
+        "disable-dissolution-sftp-bcmail": true
```
JazzarKarim, line 4: Since you're here, I think you can also add the FF for BCLaws 🙏

JazzarKarim (commented):
Eason, there's a small SonarCloud issue: https://sonarcloud.io/project/issues?id=bcgov_lear&pullRequest=3028&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true
image

If you can fix that, that would be awesome. You just might need to do a very small refactor for the function.

jobs/involuntary-dissolutions/involuntary_dissolutions.py
eason-pan-bc: adjusted the logic, now it appends the QUEUED status for checking when FF is on

jobs/furnishings/flags.json
eason-pan-bc: added 😄

eason-pan-bc:
> Eason, there's a small SonarCloud issue: https://sonarcloud.io/project/issues?id=bcgov_lear&pullRequest=3028&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true image
> 
> If you can fix that, that would be awesome. You just might need to do a very small refactor for the function.

Refactored 😄 
image

jobs/involuntary-dissolutions/involuntary_dissolutions.py
chenhongjing: Think you can also update tests a bit 👀

eason-pan-bc: haha, you read my mind, I'm adding tests, like now 😄

chenhongjing (approved):
LGTM

AimeeGao (approved):
LGTM(^_^)

PR #3029, opened 2024-10-17
23563 - update review endpoint

ketaki-deodhar:
*Issue #:* /bcgov/entity#23563

*Description of changes:*

Removed unneeded properties like FED from filter and sort

### DIFF
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -4639,7 +4639,7 @@ components:
     Continuation_in:
       type: object
       title: Continuation In Filing
-      description: Represents the filing for a foreign corporation applying to “continue in” to BC. To do so, authorization must have been received from the foreign jurisdiction. Authorization is typically in the form of a letter. The authorization must be sent to the registry as a precondition to this transaction.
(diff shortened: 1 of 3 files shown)


docs/business.yaml
```diff
@@ -4649,9 +4649,14 @@ components:
...
           - authorization
           - nameRequest
+          - contactPoint
           - offices
           - parties
+          - shareStructure
           properties:
+            isApproved:
+              type: boolean
+              description: 'Continuation In Authorization has been approved.'
```
vysakh-menon-aot, line 4659: We don't need to publish `isApproved`. Its only used for internal purpose (for validation)

ketaki-deodhar: Removed. The only change I see is authorization date is not required anymore and added some required properties

ArwenQin (approved):

vysakh-menon-aot (approved):

vysakh-menon-aot (approved):

argush3:
looks good.  should be good for publishing

PR #3042, opened 2024-10-28
23744 Script to delete migrated data

chenhongjing:
*Issue #:* /bcgov/entity#23744

*Description of changes:*
- Create a script to delete migrated data
- Refer to doc delete_script in shared folder for tables included in this script

### DIFF
--- a/data-tool/.corps.env.sample
+++ b/data-tool/.corps.env.sample
@@ -68,4 +68,4 @@ BATCH_SIZE=300
 
 ## delete corps record in auth db, corp_processing of colin extract
 DELETE_AUTH_RECORDS=False
-DELETE_CORP_PREOCESSING_RECORDS=True
+DELETE_CORP_PROCESSING_RECORDS=True
--- a/data-tool/.corps.env.sample
+++ b/data-tool/.corps.env.sample
@@ -54,15 +54,18 @@ ACCOUNT_SVC_CLIENT_SECRET=
 ACCOUNT_SVC_TIMEOUT=
 
 
-# Auth database (Used for delete script)
+# delete script
+## Auth database
 DATABASE_USERNAME_AUTH=
 DATABASE_PASSWORD_AUTH=
 DATABASE_NAME_AUTH=
 DATABASE_HOST_AUTH=
 DATABASE_PORT_AUTH=
 
-# delete batch configuration
-BATCHES=1  # optional, no such variable means dynamically calculating the value during runtime
+## batch configs, must be positive integer
+BATCHES=1
 BATCH_SIZE=300
+
+## delete corps record in auth db, corp_processing of colin extract
 DELETE_AUTH_RECORDS=False
-DELETE_COLIN_RECORDS=True
+DELETE_CORP_PREOCESSING_RECORDS=True
--- a/data-tool/.corps.env.sample
+++ b/data-tool/.corps.env.sample
@@ -53,3 +53,16 @@ ACCOUNT_SVC_CLIENT_ID=
 ACCOUNT_SVC_CLIENT_SECRET=
 ACCOUNT_SVC_TIMEOUT=
 
+
+# Auth database (Used for delete script)
+DATABASE_USERNAME_AUTH=
+DATABASE_PASSWORD_AUTH=
+DATABASE_NAME_AUTH=
+DATABASE_HOST_AUTH=
+DATABASE_PORT_AUTH=
+
+# delete batch configuration
+BATCHES=1  # optional, no such variable means dynamically calculating the value during runtime
+BATCH_SIZE=300
+DELETE_AUTH_RECORDS=False
+DELETE_COLIN_RECORDS=True
--- a/data-tool/.corps.env.sample
+++ b/data-tool/.corps.env.sample
@@ -36,8 +36,8 @@ DATABASE_POOL_SIZE=1000
 DATABASE_MAX_OVERFLOW=15
 DATABASE_POOL_RECYCLE=3600
 
-#corp name prefix to append for test purposes
-CORP_NAME_PREFIX=" - IMPORT_TEST"
+#corp name suffix to append for test purposes
+CORP_NAME_SUFFIX=" - IMPORT_TEST"
 
 #AUTH - NO Auth values as no auth creation should take place
 UPDATE_ENTITY=False
--- a/data-tool/flows/batch_delete_flow.py
+++ b/data-tool/flows/batch_delete_flow.py
@@ -290,12 +290,13 @@ def auth_delete(db_engine: Engine, identifiers: list):
 @task
 def colin_delete(config, db_engine: Engine, identifiers: list):
     with db_engine.connect() as conn:
-        plan = execute_query(conn, {
-            'source': 'corp_processing',
-            'params': {'corp_num': identifiers, 'environment': [config.DATA_LOAD_ENV]}
-        })
+        with replica_role(conn):
+            plan = execute_query(conn, {
+                'source': 'corp_processing',
+                'params': {'corp_num': identifiers, 'environment': [config.DATA_LOAD_ENV]}
+            })
 
-        delete_by_ids(conn, 'corp_processing', plan['corp_processing'])
+            delete_by_ids(conn, 'corp_processing', plan['corp_processing'])
 
 
 @task(persist_result=False)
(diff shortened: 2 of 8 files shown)


data-tool/flows/batch_delete_flow.py
```diff
@@ -0,0 +1,439 @@
...
+
+businesses_cnt_query = """
+SELECT COUNT(*) FROM businesses
+"""
+
+identifiers_query = """
+SELECT id, identifier FROM businesses
+-- WHERE legal_type IN ('BC', 'ULC', 'CC')
+LIMIT :batch_size
+"""
```
chenhongjing, line 19: Should change this query to select a subset of businesses to delete.

argush3: i think the default should probably be pretty locked down and whoever uses it will need to manually update the query if they need something else.

So may lock down to legal_type, `BC', 'ULC', 'CC', `businesses.legal name` contains `IMPORT_TEST` and a few `identifiers` by default.

We can nail down exactly what we want as we start using it to wipe corps in dev/test.  

For firms we could lock down identifiers to start with FM0 but it seems like some corps have BC1 prefix so don't think we can use that.

data-tool/flows/config.py
```diff
@@ -109,6 +123,12 @@ class _Config():  # pylint: disable=too-few-public-methods
     ACCOUNT_SVC_CLIENT_SECRET = os.getenv('ACCOUNT_SVC_CLIENT_SECRET')
     ACCOUNT_SVC_TIMEOUT = os.getenv('ACCOUNT_SVC_TIMEOUT')
 
+    # batch delete flow
+    BATCHES = int(os.getenv('BATCHES', sys.maxsize))
+    BATCH_SIZE = int(os.getenv('BATCH_SIZE', 300))
```
argush3, line 128: I think if there are no values given, we should default these two to 0. 

Anyone who runs this flow needs to set the values explicitly.

Think you'll need to also update the logic to maybe exit out of the flow immediately when one of these values are 0 as you'll run into division by zero errors.

data-tool/flows/batch_delete_flow.py
```diff
@@ -0,0 +1,444 @@
...
+
+@task
+def colin_delete(config, db_engine: Engine, identifiers: list):
+    with db_engine.connect() as conn:
+        plan = execute_query(conn, {
+            'source': 'corp_processing',
+            'params': {'corp_num': identifiers, 'environment': [config.DATA_LOAD_ENV]}
+        })
+
+        delete_by_ids(conn, 'corp_processing', plan['corp_processing'])
```
argush3, line 298: `corp_processing` entries did not get deleted when i tried this locally.

The entries still exist after the flow runs.  I could see that the right environment and business identifiers were being picked up via the delete flow logging.

**corp_processing table after delete flow ran**
image

**delete flow log snippet**
```
11:26:51.319 | INFO    | Flow run 'heavy-mastodon' - 🚀 Running round 0 to delete 5 busiesses...
11:26:51.362 | INFO    | Task run 'colin_delete-7db' - colin_delete - identifiers: ['BC0813248', 'BC0814225', 'BC0814580', 'BC0812630', 'BC0814579']
11:26:51.364 | INFO    | Task run 'colin_delete-7db' - colin delete environment - local_migr_test
11:26:51.406 | INFO    | Task run 'execute_query-d6e' - Finished in state Completed()
11:26:51.410 | INFO    | Task run 'execute_query-736' - Finished in state Completed()
11:26:51.632 | INFO    | Task run 'delete_by_ids-d46' - Delete 5 rows from corp_processing
11:26:51.664 | INFO    | Task run 'delete_by_ids-d46' - Finished in state Completed()
11:26:51.687 | INFO    | Task run 'execute_query-532' - Finished in state Completed()
11:26:51.688 | INFO    | Task run 'colin_delete-7db' - Finished in state Completed()
11:26:51.694 | INFO    | Task run 'execute_query-b64' - Finished in state Completed()
```

argush3:
Can we add a make command for this flow to the makefile?

Maybe something like what we have for `run-corps-migration` command.

data-tool/.corps.env.sample
```diff
@@ -36,8 +36,8 @@ DATABASE_POOL_SIZE=1000
 DATABASE_MAX_OVERFLOW=15
 DATABASE_POOL_RECYCLE=3600
 
-#corp name prefix to append for test purposes
-CORP_NAME_PREFIX=" - IMPORT_TEST"
+#corp name suffix to append for test purposes
+CORP_NAME_SUFFIX=" - IMPORT_TEST"
```
argush3, line 40: thanks for catching this!

data-tool/flows/batch_delete_flow.py
```diff
@@ -0,0 +1,444 @@
...
+                # there're some Comment records saved by legal-api directly instead of filer
+                # should ignore them (assume that it won't be populated with migrated data)
+                {
+                    'source': 'comments',
+                    'params': {'filing_id': filing_ids},
+                },
+                {
+                    'source': 'consent_continuation_outs',
+                    'params': {'filing_id': filing_ids},
+                },
```
argush3, line 190: I think we need to delete the entries where the comment data is directly linked via the business_id too

argush3:
Great work!

I like how you've made things quite generic and easy to add additional tables to remove if we wanted to.

I ran this locally in conjunction with the corp migration pipeline and was able to test the delete flow a bit.  Mainly verifying the  expected record counts in each table after removing a specific corp.  Things seem to check out from some initial testing.  

As I had mentioned, `corp_processing` entries were not deleted from the COLIN extract which you'll need to sort out.

Auth is something we'll have to maybe verify later when we get access to the db in GCP.

data-tool/.corps.env.sample
argush3, line 1: can we add your new params for the delete flow into this file?

data-tool/flows/config.py
```diff
@@ -109,6 +123,12 @@ class _Config():  # pylint: disable=too-few-public-methods
...
+    DELETE_AUTH_RECORDS = os.getenv('DELETE_AUTH_RECORDS').lower() == 'true'
+    DELETE_COLIN_RECORDS = os.getenv('DELETE_COLIN_RECORDS').lower() == 'true'
```
argush3, line 130: I think maybe we should name this something else.  

For now it's strictly for deleting corp_processing records from the extract.

`DELETE_CORP_PROCESSING_RECORDS` or something...

data-tool/flows/config.py
chenhongjing: I introduced `BATHES` because sometimes we may want to run specific number of runs (1 or 2 used for debugging etc.)...most of time, the value will be set dynamically via calculation. And it is used as 
```python
batches = min(math.ceil(total/batch_size), config.BATCHES)
```
in the [code](https://github.com/bcgov/lear/blob/4184b4dc53b5fdece1572dff535ce256c7165da7/data-tool/flows/batch_delete_flow.py#L405). Do you still want me to set it to 0 by default?

But I'll handle the divison by zero error, which I didn't consider well before.

argush3: I'm thinking we should still set `BATCHES` to 0 by default just to be careful.

We can maybe add some validation to the start of `batch_delete_flow` and exit right away if there are zero values.

```python
if config.BATCH_SIZE <= 0:
    raise ValueError("BATCH_SIZE must be explicitly set to a positive number")
if config.BATCHES <= 0:
    raise ValueError("BATCHES must be explicitly set to a positive number")
```

data-tool/flows/config.py
```diff
@@ -109,6 +123,15 @@ class _Config():  # pylint: disable=too-few-public-methods
...
     ACCOUNT_SVC_TIMEOUT = os.getenv('ACCOUNT_SVC_TIMEOUT')
 
+    # batch delete flow
+    BATCHES = os.getenv('BATCHES')
+    BATCHES = int(BATCHES) if BATCHES.isnumeric() else 0
+    BATCH_SIZE = os.getenv('BATCH_SIZE')
+    BATCH_SIZE = int(BATCH_SIZE) if BATCH_SIZE.isnumeric() else 0
+
+    DELETE_AUTH_RECORDS = os.getenv('DELETE_AUTH_RECORDS').lower() == 'true'
+    DELETE_CORP_PREOCESSING_RECORDS = os.getenv('DELETE_CORP_PREOCESSING_RECORDS').lower() == 'true'
```
argush3, line 133: typo on processing

argush3 (approved):

PR #3033, opened 2024-10-22
22942 update model to use new sql_versioning library

kzdev420:
*Issue #:* /bcgov/entity#22942

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #3044, opened 2024-10-29
Suggested changes for 22942, 22943

chenhongjing:
*Issue #:* /bcgov/entity#22942 /bcgov/entity#22943 /bcgov/entity#22944

*Description of changes:*

- ❌**JUST SUGGESTED CHANGES! DO NOT MERGE**❌
- Tweaks to support nested transaction (Didn't consider it since it didn't appear in BE services but now it shows it will appear in unit tests)
- Tweaks to support differences between column name and property name (e.g. resolution data model) 
- Suggested fix for #3033 

Notes:
❗❗ Think we may need some work to support relationships mapping in versioned record, or just be careful about query of every versioned record

Differences of relationship mapping between old and new method
- old (has relationship)
image

- new (relationship doesn't exist)
image

### DIFF
--- a/legal-api/flags.json
+++ b/legal-api/flags.json
@@ -11,7 +11,7 @@
             "exclude-accounts": []
         },
         "db-versioning": {
-            "legal-api": false,
+            "legal-api": true,
             "emailer": false,
             "filer": false,
             "entity-bn": false,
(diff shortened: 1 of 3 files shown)


kzdev420 (approved):
LGTM.
Thanks @chenhongjing

PR #3047, opened 2024-10-29
Fix 24073 - fix founding date validation

ArwenQin:
*Issue #:* /bcgov/entity#24073

*Description of changes:*
- Fix the bug due to different date format

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/continuation_in.py
+++ b/legal-api/src/legal_api/services/filings/validations/continuation_in.py
@@ -192,7 +192,21 @@ def validate_business_in_colin(filing_json: dict, filing_type: str) -> list:
         elif legal_name != response_json['business']['legalName']:
             msg.append({'error': 'Legal name does not match with company legal name from Colin.',
                         'path': business_legal_name_path})
-        elif founding_date != response_json['business']['foundingDate']:
-            msg.append({'error': 'Founding date does not match with founding date from Colin.',
-                        'path': business_founding_date_path})
+        else:
+            # Check if founding date is None
+            if founding_date is None:
+                msg.append({
+                    'error': 'Founding date is missing in the filing data.',
+                    'path': business_founding_date_path
+                })
+            else:
+                # Truncate both dates to "YYYY-MM-DDTHH:MM:SS" for comparison
+                founding_date_truncated = founding_date[:19]
+                colin_founding_date_truncated = response_json['business']['foundingDate'][:19]
+
+                if founding_date_truncated != colin_founding_date_truncated:
+                    msg.append({
+                        'error': 'Founding date does not match with founding date from Colin.',
+                        'path': business_founding_date_path
+                    })
     return msg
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/services/filings/validations/continuation_in.py
```diff
@@ -192,7 +192,21 @@ def validate_business_in_colin(filing_json: dict, filing_type: str) -> list:
...
-        elif founding_date != response_json['business']['foundingDate']:
-            msg.append({'error': 'Founding date does not match with founding date from Colin.',
-                        'path': business_founding_date_path})
+        else:
+            # Check if founding date is None
+            if founding_date is None:
+                msg.append({
+                    'error': 'Founding date is missing in the filing data.',
+                    'path': business_founding_date_path
+                })
```
ArwenQin, line 201: Do we need this?

legal-api/src/legal_api/services/filings/validations/continuation_in.py
```diff
@@ -192,7 +192,21 @@ def validate_business_in_colin(filing_json: dict, filing_type: str) -> list:
...
+            else:
+                # Truncate both dates to "YYYY-MM-DDTHH:MM:SS" for comparison
+                founding_date_truncated = founding_date[:19]
+                colin_founding_date_truncated = response_json['business']['foundingDate'][:19]
+
```
ArwenQin, line 206: Colin is  '2009-07-23T07:00:00-00:00', the filing json is '2009-07-23T07:00:00.000+00:00'
truncated after 19th

legal-api/src/legal_api/services/filings/validations/continuation_in.py
severinbeauvais: If the schema already makes this required then no.

severinbeauvais (approved):
LGTM.

Vysakh, please merge if OK with you.

legal-api/src/legal_api/services/filings/validations/continuation_in.py
vysakh-menon-aot: its not a required filed as per schema

ArwenQin: Just to confirm how to handle this: if per filing json is none, do we check Colin returns none too (does this happen?); or return error message?

legal-api/src/legal_api/services/filings/validations/continuation_in.py
vysakh-menon-aot: This is not a good approach. Can you try this one
```suggestion
datetime.fromisoformat(founding_date) !=  datetime.fromisoformat(response_json['business']['foundingDate'])
```

ArwenQin: Updated, thanks. Only concern is whether Colin can have different time zone? Other than -00:00, like +03:00. If not I think this approach is safe

vysakh-menon-aot (approved):

PR #3063, opened 2024-11-06
Fix validation rules for filing, add a could file detection for if th…

BrandonSharratt:
…e user is allowed to do something to a legal type, fix flake violations in continuation_in

*Issue #:* /bcgov/entity#22051

*Description of changes:*
Fixed differences between api and spreadsheet (see below). Also added a could file which just returns if a user could do a filing on a business of the type to the get endpoint.

Differences
Staff
Amalgamation - Regular/Horizontal/Vertical — were additionally blocked on pending filings
Put back on was allowed when it had 'amalgamationApplication' or ’administrative' filings or a future effective dissolution filing
Full Restoration was allowed when there was a continuation in filing
Limited Restoration was allowed when there was a continuation in filing
Change of Registration - was blocked for staff when required info was missing and shouldn’t be
Continuation In didn’t support any of the types in the spreadsheet
Correction required no missing info always not just on SP/GP
Voluntary Dissolution required no missing info always not just on SP/GP
Administrative Dissolution required no missing info always not just on SP/GP
Notice of withdrawal was allowed when business was frozen or had a pending filing

General
Amalgamation - Regular/Horizontal/Vertical — were additionally blocked on pending filings
Continuation In didn’t support any of the types in the spreadsheet
Voluntary Dissolution required no missing info always not just on SP/GP

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -225,6 +225,11 @@ def get_allowable_for_business_type(business_type: str, business_state: str):
     business_type = business_type.upper()
     if business_type.startswith('T'):
         return {'message': babel('No information on temp registrations.')}, 200
+    
+    bs_state = getattr(Business.State, business_state, False)
+    bs_type = getattr(Business.LegalTypes, business_type, False)
+    if not bs_state or not bs_type:
+        return {'message': babel('Invalid business type or state.')}, 400
 
     could_file = get_could_files(jwt, business_type, business_state)
 
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -223,7 +223,7 @@ def get_allowable_for_business_type(business_type: str, business_state: str):
     """Return a JSON object with information about what a user could theoretically file for a business type."""
     business_state = business_state.upper()
     business_type = business_type.upper()
-    
+
     bs_state = getattr(Business.State, business_state, False)
     bs_type = getattr(Business.LegalTypes, business_type, False)
     if not bs_state or not bs_type:
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -223,8 +223,6 @@ def get_allowable_for_business_type(business_type: str, business_state: str):
     """Return a JSON object with information about what a user could theoretically file for a business type."""
     business_state = business_state.upper()
     business_type = business_type.upper()
-    if business_type.startswith('T'):
-        return {'message': babel('No information on temp registrations.')}, 200
     
     bs_state = getattr(Business.State, business_state, False)
     bs_type = getattr(Business.LegalTypes, business_type, False)
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -34,7 +34,7 @@ from legal_api.services import (  # noqa: I001;
     RegistrationBootstrapService,
     check_warnings,
 )  # noqa: I001;
-from legal_api.services.authz import get_allowable_actions, get_allowed
+from legal_api.services.authz import get_allowable_actions, get_allowed, get_could_files
 from legal_api.utils.auth import jwt
 
 from .bp import bp
@@ -64,6 +64,7 @@ def get_businesses(identifier: str):
         business_json = business.json(slim=True)
         # need to add the alternateNames array here because it is not a part of slim JSON
         business_json['alternateNames'] = business.get_alternate_names()
+
         return jsonify(business=business_json)
 
     warnings = check_warnings(business)
@@ -75,6 +76,10 @@ def get_businesses(identifier: str):
     business.allowable_actions = allowable_actions
 
     business_json = business.json()
+
+    could_file = get_could_files(jwt, business)
+    business_json['couldFile'] = could_file
+
     recent_filing_json = CoreFiling.get_most_recent_filing_json(business.id, None, jwt)
     if recent_filing_json:
         business_json['submitter'] = recent_filing_json['filing']['header']['submitter']
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -225,9 +225,13 @@ def get_allowable_for_business_type(business_type: str, business_state: str):
     business_type = business_type.upper()
 
     bs_state = getattr(Business.State, business_state, False)
-    bs_type = getattr(Business.LegalTypes, business_type, False)
-    if not bs_state or not bs_type:
-        return {'message': babel('Invalid business type or state.')}, 400
+    if not bs_state:
+        return {'message': babel('Invalid business state.')}, 400
+
+    try:
+        _ = Business.LegalTypes(business_type)
+    except ValueError:
+        return {'message': babel('Invalid business type.')}, 400
 
     could_file = get_could_files(jwt, business_type, business_state)
 
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -226,12 +226,12 @@ def get_allowable_for_business_type(business_type: str, business_state: str):
 
     bs_state = getattr(Business.State, business_state, False)
     if not bs_state:
-        return {'message': babel('Invalid business state.')}, 400
+        return {'message': babel('Invalid business state.')}, HTTPStatus.BAD_REQUEST
 
     try:
         _ = Business.LegalTypes(business_type)
     except ValueError:
-        return {'message': babel('Invalid business type.')}, 400
+        return {'message': babel('Invalid business type.')}, HTTPStatus.BAD_REQUEST
 
     could_file = get_could_files(jwt, business_type, business_state)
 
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -75,6 +76,10 @@ def get_businesses(identifier: str):
     business.allowable_actions = allowable_actions
 
     business_json = business.json()
+
+    could_file = get_could_files(jwt, business)
+    business_json['couldFile'] = could_file
```
vysakh-menon-aot, line 81: Since we are changing this now, its better to move to a separate endpoint (`'/<string:identifier>/allowable-actions'`), as this is required only in filings-ui (and api users).

legal-api/src/legal_api/services/authz.py
```diff
@@ -537,6 +555,65 @@ def get_allowable_actions(jwt: JwtManager, business: Business):
...
+                   x[1].get('legalTypes', []), allowable_filing_value.items())
+
+        for filing_sub_type_item_key, _ in filing_sub_type_items:
+
+            allowable_filing_sub_type = {'name': allowable_filing_key,
+                                         'type': filing_sub_type_item_key,
+                                         'displayName': FilingMeta.get_display_name(legal_type,
+                                                                                    allowable_filing_key,
+                                                                                    filing_sub_type_item_key),
+                                         'feeCode': Filing.get_fee_code(legal_type,
```
vysakh-menon-aot, line 607: If you are planning to use `feeCode` values, there is one exception case (a dynamic fee code based on the user action in alteration. if alteration filing is to convert `BC to ULC` then fee code will be `NOALU` (https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/filing.py#L116) and this fee code will not be available in get_fee_code function

legal-api/src/legal_api/services/authz.py
```diff
@@ -537,6 +555,65 @@ def get_allowable_actions(jwt: JwtManager, business: Business):
     return result
 
 
+def get_could_file(business: Business,
```
vysakh-menon-aot, line 558: If I compare this function with `get_allowed_filings` the only change you have done here is to remove `business_blocker_check`. can you explain why?

legal-api/src/legal_api/resources/v2/business/business.py
severinbeauvais: AFAIK, only Filings UI uses allowable actions (though I'd like Edit UI to use it in the future).

legal-api/src/legal_api/services/authz.py
BrandonSharratt: So the intent behind could_file is to have an api answer on whether a user could ever do a type of filing. So while allowedActions says you can/can't do this for this business, couldFile responds that you could do it for a business of this legalType. We intend to use it to show/hide options, IE couldFile --> visibility, allowedActions --> disabled/enabled

legal-api/src/legal_api/services/authz.py
BrandonSharratt: We aren't intending to use feeCode, I just wanted to keep the structure the same, and while I would have normally added another flag to allowedActions to remove the blockers check I figured this would be an easier review because it can't possibly affect the current use cases as they all use allowedActions

legal-api/src/legal_api/resources/v2/business/business.py
BrandonSharratt: I'd prefer to not move it to another endpoint, as it increases the network verbosity by requiring 2 api calls when we can currently get it in 1 (and both can be ignored with the slim param)

legal-api/src/legal_api/services/authz.py
severinbeauvais: I think there "could" (pun intended) be other things that determiend whether or not to show/hide an action (eg, current user).

I'd prefer to defer this implementation until a complete list of show/hide conditions is known.

legal-api/src/legal_api/resources/v2/business/business.py
vysakh-menon-aot: The business endpoint is used in many places, which does not care about allowable actions. I agree it adds an additional api call but it reduces the latency for all other calls to business endpoint

legal-api/src/legal_api/services/authz.py
BrandonSharratt: The current user and the business type is considered

vysakh-menon-aot: If you get all the filings a user can file for a business, how do you check if that action is blocked or not?

BrandonSharratt: We still use allowedActions and disable/enable the button based on that flag which checks blocked. the Blocked flag is business specific and doesn't tell you whether or not the current user could file a certain type of filing for a different business of the same type. IE they could file this for a BC business (so it's in could file) but they aren't allowed to for this business (so it isn't in allowedActions)

argush3: well, if i'm understanding things correctly.  it doesn't sound like the functionality you are looking for fits under the GET /businesses/<identifier> endpoint.  

It sounds like you need another endpoint that retrieves allowed filings based off of current user and business type.  

So I don't think it makes sense that you are trying to retrieve this info from a specific business?

legal-api/src/legal_api/resources/v2/business/business.py
vysakh-menon-aot: I am not suggesting to remove `business.allowable_actions` from the business endpoint for now, instead introduce new endpoint for the new UI and based on the release of the new UI we can depreciate `business.allowable_actions` in business endpoint.

legal-api/src/legal_api/services/authz.py
severinbeauvais: > The current user and the business type is considered

The current implementation is simplistic (basically, on or off depending on product key). I know that some user types will have more or less permissions than others (ie, staff > Service BC > reg user).

Before implementing get_could_files, I'd like to see what conditions (in the old UI) are used for showing/hiding actions, and making sure they're considered here.

severinbeauvais: I guess, in a nutshell, I'd like to see this feature generalized.

BrandonSharratt:
This PR has been updated and no longer has the DO NOT MERGE tag necessary (that was only needed when the rules were in flight). A test has been added for the new endpoint. Api Documentation/Spec for this new endpoint is to be created after merge as a tech debt ticket

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -212,3 +214,18 @@ def search_businesses():
...
         return {'error': 'Unable to retrieve businesses.'}, HTTPStatus.INTERNAL_SERVER_ERROR
+
+
+@bp.route('/allowable/<string:business_type>/<string:business_state>', methods=['GET'])
+@cross_origin(origin='*')
+@jwt.requires_auth
+def get_allowable_for_business_type(business_type: str, business_state: str):
+    """Return a JSON object with information about what a user could theoretically file for a business type."""
+    business_state = business_state.upper()
+    business_type = business_type.upper()
```
argush3, line 225: I would validate to ensure valid business state and business types are being provided.

argush3 (approved):
Looks good.  

Pre-approving with assumption that business type and business state validation comment will be addressed for new endpoint.

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -212,3 +214,18 @@ def search_businesses():
...
+    if business_type.startswith('T'):
+        return {'message': babel('No information on temp registrations.')}, 200
```
vysakh-menon-aot, line 227: it is identifier which starts with T. you can remove this validation

BrandonSharratt: Removed

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -212,3 +214,21 @@ def search_businesses():
...
+@jwt.requires_auth
+def get_allowable_for_business_type(business_type: str, business_state: str):
+    """Return a JSON object with information about what a user could theoretically file for a business type."""
+    business_state = business_state.upper()
+    business_type = business_type.upper()
+
+    bs_state = getattr(Business.State, business_state, False)
+    bs_type = getattr(Business.LegalTypes, business_type, False)
+    if not bs_state or not bs_type:
+        return {'message': babel('Invalid business type or state.')}, 400
```
vysakh-menon-aot, line 230:
```suggestion
        return {'message': babel('Invalid business type or state.')}, HTTPStatus.BAD_REQUEST
```

vysakh-menon-aot:
Please fix the sonarcloud issue

severinbeauvais (approved):
LGTM but please update HTTP response code per Vysakh's comment.

legal-api/src/legal_api/services/authz.py
severinbeauvais: Obsolete comment. Resolving.

legal-api/src/legal_api/resources/v2/business/business.py
severinbeauvais: @vysakh-menon-aot Why should this validation be removed?

vysakh-menon-aot: this endpoint doesn't have identifier as parameter

vysakh-menon-aot (approved):

legal-api/src/legal_api/resources/v2/business/business.py
severinbeauvais: Oh, right 😁 Thanks.

PR #3077, opened 2024-11-14
24051 Tombstone pipeline (first cut)

chenhongjing:
*Issue #:* /bcgov/entity#24051

*Description of changes:*
- add tombstone pipeline (first cut)
    - able to load corp snapshot (exclude filings) into db
- move common utils
- update extract scripts
- update makefile

Note that there are lots of TODOs in the code and they will be done in future work.

*local testing*
_This is a testing corp without sensitive info_
image

### DIFF
--- a/data-tool/flows/corps_tombstone_flow.py
+++ b/data-tool/flows/corps_tombstone_flow.py
@@ -18,7 +18,7 @@ def get_unprocessed_corps(config, colin_engine: Engine) -> list:
     query = get_unprocessed_corps_query(
         'local',
         config.DATA_LOAD_ENV,
-        config.BATCH_SIZE
+        config.TOMBSTONE_BATCH_SIZE
     )
     sql_text = text(query)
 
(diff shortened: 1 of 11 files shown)


data-tool/flows/corps_tombstone_flow.py
```diff
@@ -0,0 +1,232 @@
...
+        config = get_config()
+        colin_engine = colin_init(config)
+        lear_engine = lear_init(config)
+
+        total = get_unprocessed_count(config, colin_engine)
+
+        if config.BATCHES <= 0:
+            raise ValueError('BATCHES must be explicitly set to a positive integer')
+        if config.BATCH_SIZE <= 0:
+            raise ValueError('BATCH_SIZE must be explicitly set to a positive integer')
```
argush3, line 189: if we have config for `BATCHES` and `BATCH_SIZE`, I think they should it should be specific to the tombstone flow.  i.e. the delete flow and tombstone flow should not share "batch" related config.

argush3 (approved):
Great work!  

The abstraction and making things follow a reusable pattern for building the queries and loading the snapshot will be nice.

There are still a lot of misc things we need to do but for a first cut this is great.

We can address all the other things in future tickets as we've talked about.

PR #3086, opened 2024-11-20
24387 Tombstone pipeline - Load placeholder filings & user info

chenhongjing:
*Issue #:* /bcgov/entity#24387

*Description of changes:*
- update tombstone pipeline to
    - load plcaholder filings
    - load user info (and prevent duplicate user loaded into db)
    - only support a few basic event_filing_type mapping now (will re-visit it in the future work)

Note that this is a rough update, is going on hold for a short time until bcgov/entity#24486 complete

*local testing*
_This is a testing corp without sensitive info_
image

### DIFF
--- a/data-tool/flows/tombstone/tombstone_mappings.py
+++ b/data-tool/flows/tombstone/tombstone_mappings.py
@@ -8,7 +8,7 @@ class EventFilings(str, Enum):
     # AGM Location Change
     FILE_AGMLC = 'FILE_AGMLC'
 
-    # TODO: Alteration
+    # TODO: Alteration (Some need further confirmation)
     FILE_NOALA = 'FILE_NOALA'
     FILE_NOALB = 'FILE_NOALB'
     FILE_NOALU = 'FILE_NOALU'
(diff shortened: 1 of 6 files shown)


data-tool/flows/tombstone/tombstone_mappings.py
```diff
@@ -0,0 +1,87 @@
...
+    # transition
+    FILE_TRANS = 'FILE_TRANS'
+
+
+    @classmethod
+    def has_value(cls, value):
+        return value in cls._value2member_map_
+    
+
+EVENT_FILING_LEAR_TARGET_MAPPING = {
```
chenhongjing, line 47: Planned to use `from legal_api.core import Filing as FilingCore` and leverage code in `corps/event_filing_service.py` at the first beginning, but dependency issues related to sqlalchemy kept comming up even I referred to the solution in `flows/migrate_corps_flow.py`...So eventually choose to get rid of any legal-api and sqlalchemy dependencies.

data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -210,22 +221,160 @@ def format_resolutions_data(data: dict) -> list[dict]:
...
+        effective_date = x['f_effective_dt_str']
+        if not effective_date:
+            effective_date = x['e_event_dt_str']
+
+        meta_data = {
+            'colinFilingInfo': {
+                'eventType': x['e_event_type_cd'],
+                'filingType': x['f_filing_type_cd']
+            }
+        }
```
argush3, line 249: when we add the placeholder filings, can we also add the following to the root of filngs.meta_data?

`isLedgerPlaceholder: true`

argush3:
Looks good so far.

Will take another look once you move to ready for review.

argush3 (approved):

PR #3089, opened 2024-11-21
24486 - Update delete script

chenhongjing:
*Issue #:* /bcgov/entity#24486

*Description of changes:*
Miscellaneous updates for delete script, including
- to support deleting entity & affiliation via auth api
- to delete lear_tombstone filing
- to delete no versioning records

p.s. we should delete `colin_event_ids`, so keep it unchanged (already support it in previous work)

### DIFF
--- a/data-tool/flows/batch_delete_flow.py
+++ b/data-tool/flows/batch_delete_flow.py
@@ -1,7 +1,10 @@
 import math
+from collections import defaultdict
 from contextlib import contextmanager
+from http import HTTPStatus
 
-from common.init_utils import auth_init, colin_init, get_config, lear_init
+import requests
+from common.init_utils import colin_init, get_config, lear_init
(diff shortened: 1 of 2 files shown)


data-tool/flows/batch_delete_flow.py
```diff
@@ -256,6 +352,83 @@ def colin_delete(config, db_engine: Engine, identifiers: list):
             delete_by_ids(conn, 'corp_processing', plan['corp_processing'])
 
 
+@task
+def auth_api_delete(config, identifiers: list):
```
chenhongjing, line 356: Similar situation as that in #3086, reuse the code in legal-api with some tweaks to get rid of the dependency on it.

argush3 (approved):
Looks good!

Thanks for the screenshots from your local testing

PR #3093, opened 2024-11-26
22942 fix attr name issue

kzdev420:
*Issue #:* /bcgov/entity#22942

*Description of changes:*

When saving the data into the Resolution table, there was an error.
Because the attribute name is different between sql_table(type) and code_model(resolution_type)

These updates fixed that issue when saving the data into the Resolution and Resolution_version tables.

### DIFF
--- a/python/common/sql-versioning/sql_versioning/versioning.py
+++ b/python/common/sql-versioning/sql_versioning/versioning.py
@@ -93,7 +93,7 @@ def _create_version(session, target, operation_type):
         return
 
     transaction_manager = TransactionManager(session)
-    transaction_id = transaction_manager.create_transaction()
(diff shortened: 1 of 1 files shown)


leodube-aot (approved):
Looks good.

For reference to other devs: it was decided that since this package already exists in the common components of `main` (and packages on the `feature-db-versioning` branch import it), we would sync these changes to the main branch. That way we don't have to worry about changing a bunch of imports later on.

eason-pan-bc (approved):
LGTM, thanks for updating the sql-versioning package.
This change has already been reviewed and approved in a PR in the feature branch https://github.com/bcgov/lear/pull/3084

AimeeGao (approved):
LGTM : )

PR #3099, opened 2024-11-28
20619

ketaki-deodhar:
*Issue #:* /bcgov/entity#20619

*Description of changes:*

fix allowable actions for:

-  when business is NIGS, allow Voluntary Dissolution for staff only
-  when business is NIGS, allow Alteration for staff only
-  when there is an existing draft filing, disallow Amalgamations (x3) (**Already working as expected. Mihai confirmed the requirement**)
-  when business is NIGS, disallow Special Resolution for reg users only


vysakh-menon-aot (approved):

ArwenQin (approved):
LGTM!

PR #3103, opened 2024-11-29
Add details for parValue and maxNumberOfShares fields

eason-pan-bc:
*Issue #:* /bcgov/entity###  No ticket number for this

*Description of changes:*
As per discussion with Leo and Dave, made a quick update to add information about set relevant fields to `null` in certain scenarios, so that API users would be able to put correct value for the fields.

#### Before:
image

#### After:
image

### DIFF
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -6733,11 +6733,11 @@ components:
         maxNumberOfShares:
           type: number
           title: 'Maximum number of shares in the class. '
-          description: The maximum number of shares that a corporation can issue to the investors called authorized shares.
+          description: 'The maximum number of shares that a corporation can issue to the investors called authorized shares. If hasMaximumShares is false, set this field to null.'
         parValue:
           type: number
           title: 'Initial value of each share. '
-          description: 'The value of a single common share as set by a corporation''s charter. '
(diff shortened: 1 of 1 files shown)


docs/business.yaml
```diff
@@ -6733,11 +6733,11 @@ components:
         maxNumberOfShares:
           type: number
           title: 'Maximum number of shares in the class. '
-          description: The maximum number of shares that a corporation can issue to the investors called authorized shares.
+          description: 'The maximum number of shares that a corporation can issue to the investors called authorized shares. If hasMaximumShares is false, set this field to null.'
         parValue:
           type: number
           title: 'Initial value of each share. '
-          description: 'The value of a single common share as set by a corporation''s charter. '
+          description: 'The value of a single common share as set by a corporation''s charter. If hasMaximumShares is false, set this field to null.'
```
AimeeGao, line 6740: Good job, but for the description of parValue, should we change it to ' If hasParValue is false, set this field to null.'

eason-pan-bc: nice catch, pasted a wrong sentence from my notebook

eason-pan-bc: fixed

AimeeGao (approved):
LGTM(^-^)

PR #3112, opened 2024-12-05
22955 - Verify update-colin-filings job to use new db versioning

AimeeGao:
*Issue #:* /bcgov/entity#22955

*Description of changes:*

1. Investigated the job's functionality:

    - Job only handles API calls from legal-api and colin-api
    - No direct DB operations
    - No legal-api and db versioning package usage

2. Conclusions:

    - No changes needed in db versioning setup
    - Remove the update-colin-filings flag in legal-api's flags.json

3. Findings of db versioning relationship issue:
   - Found versioned objects (BusinessVersion) missing relationship attributes (`Resolutions`)
   - Root cause:       
       - Our db versioning only copies table columns but misses the relationship mappings
      - When versioning is enabled, `BusinessVersion` lacks the `resolutions` relationship that exists in the original Business model
   - Solutions:
       - Added direct Resolution query to replace relationship access
       - Updated `_set_shares` method to handle both versioned and non-versioned cases
       - Code changes:
      ```python
      # Before
      business_dates = [item.resolution_date.isoformat() for item in primary_or_holding_business.resolutions]
      
      # After
      resolutions = Resolution.get_resolution_by_business_id(primary_or_holding_business.id)
      business_dates = [item.resolution_date.isoformat() for item in resolutions]
      ```
 
image
    
  - Debug logs showing the issue:
    
    When the FF is ON:
    image
    image

    When the FF is OFF:
    image
    image
  

4. Verification steps:

    - Tested legal-api with feature flags ON & OFF
    - Verified update-colin-filings job functionality
    - Confirmed queue message processing works correctly in the local environment
    - All unit tests passed successfully
    - Verified resolution dates are correctly retrieved in both versioned and non-versioned scenarios

ScreenShots:
* Legal-API OFF
   - Job works successfully in the local environment

image

* Legal-API ON
1.  Created a new FED filing (Change of Address) - Filing ID: 151265
image
2. Verified filing exists in the local DB
image

3. Updated filing status to COMPLETED for testing
image
4. Then run the job locally again and confirm the filing(Filing ID: 151265) was successfully updated
image
5. Verified passed all unit test
image
image

 
### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/colin_sync.py
+++ b/legal-api/src/legal_api/resources/v2/business/colin_sync.py
@@ -287,7 +287,8 @@ def _set_shares(primary_or_holding_business, amalgamation_filing, transaction_id
     share_classes = VersionedBusinessDetailsService.get_share_class_revision(transaction_id,
                                                                              primary_or_holding_business.id)
     amalgamation_filing['shareStructure'] = {'shareClasses': share_classes}
-    business_dates = [item.resolution_date.isoformat() for item in primary_or_holding_business.resolutions]
+    resolutions = Resolution.get_resolution_by_business_id(primary_or_holding_business.id)
+    business_dates = [item.resolution_date.isoformat() for item in resolutions]
     if business_dates:
         amalgamation_filing['shareStructure']['resolutionDates'] = business_dates
 
--- a/jobs/furnishings/flags.json
+++ b/jobs/furnishings/flags.json
@@ -12,8 +12,6 @@
             "dissolutions-job": false,
             "furnishings-job": true,
             "emailer-reminder-job": false,
-            "future-effective-job": false,
-            "update-colin-filings-job": false,
             "update-legal-filings-job": false
         }
     }
--- a/legal-api/flags.json
+++ b/legal-api/flags.json
@@ -18,8 +18,7 @@
             "digital-credentials": false,
             "dissolutions-job": false,
             "furnishings-job": true,
-            "emailer-reminder-job": false,
-            "update-colin-filings-job": false
+            "emailer-reminder-job": false
         }
     }
 }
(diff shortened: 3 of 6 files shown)


legal-api/src/legal_api/resources/v2/business/colin_sync.py
```diff
@@ -287,7 +287,8 @@ def _set_shares(primary_or_holding_business, amalgamation_filing, transaction_id
     share_classes = VersionedBusinessDetailsService.get_share_class_revision(transaction_id,
                                                                              primary_or_holding_business.id)
     amalgamation_filing['shareStructure'] = {'shareClasses': share_classes}
-    business_dates = [item.resolution_date.isoformat() for item in primary_or_holding_business.resolutions]
+    resolutions = Resolution.get_resolution_by_business_id(primary_or_holding_business.id)
```
AimeeGao, line 290: This is a temporary fix to handle the missing relationship in versioned objects. 
I've tested both scenarios (FF ON/OFF), and everything works fine.

But I'm not sure if this approach is acceptable. 
I only found this issue in this ticket and am unsure if it exists elsewhere. (Didn't find this issue from the previous submissions of tickets)

Should we:
1. Investigate and fix our db versioning to handle relationships properly
2. Consider how this change might impact our already completed tickets that use versioning

@leodube-aot would appreciate your thoughts on this!

kzdev420 (approved):
LGTM

legal-api/src/legal_api/resources/v2/business/colin_sync.py
leodube-aot: Seems like this is an issue we have been aware of, but I don't think it is too prevalent in the code.

Screenshot 2024-12-05 at 9 52 09 AM

I created a ticket to investigate and fix the issue. I don’t think it is super high priority, but would be nice to have. We can use the workaround for now (and then clean it up in the new ticket). bcgov/entity#24754

eason-pan-bc (approved):
LGTM 🛩️ 
Nice workaround for the issue!

PR #3114, opened 2024-12-05
24478 Add corp processing logic to tombstone pipeline

argush3:
*Issue #:* /bcgov/entity#24478

*Description of changes:*

* Add `claimed_at` and `flow_run_id` fields to `corp_processing` table
* Add `CorpProcessingQueueService`.  This differs a bit from what we had done previously for firms and old corps pipeline.  The underlying `corp_processing` table is still fully intact but this service has been re-implemented to handle parallel flows.  In particular a concept of reserving corps for a given flow run and claiming a batch of reserved corps has been introduced.  It hasn't been fully implemented but enough to handle the general flow of loading a tombstone corp from start to finish.  Error handling corp processing status updates still needs to be handled.  This work will take place in the future.  The batch logic in tombstone flow is also still a bit messy so this will need to be refactored a bit in future work as well.
* Misc comment updates
* Addition of commented deployment code to test parallel flows easily locally.  Will use prefect.yaml to do real deployments if we host prefect infra in GCP
* Update `get_unprocessed_corps_query` to take into account `flow_run_id` when determining unprocessed corps

**Below is example data and info from a local test run that tested multiple parallel flows at a time**

.env values
```
TOMBSTONE_BATCHES=5
TOMBSTONE_BATCH_SIZE=10
```

used following in `if __name__ == "__main__":` to simulate parallel flows
```
    # Create deployment - only intended to test locally for parallel flows
    deployment = tombstone_flow.to_deployment(
        name="tombstone-deployment",
        interval=timedelta(seconds=8),  # Run every x seconds
        tags=["tombstone-migration"]
    )

    # Start serving the deployment
    serve(deployment)
```

Multiple parallel flows
image

Where to find prefect flow run id
image

Query to show flow run data and businesses processed for flows
```sql
select flow_name, flow_run_id, create_date, processed_status, count(*) as cnt
from corp_processing
where 1 = 1
  and flow_name = 'tombstone-flow'
group by flow_name, flow_run_id, create_date, processed_status
order by create_date desc
;
```
image

Query to show batches are claimed properly for a given flow run via the claimed_at field
```sql
select flow_name, flow_run_id, claimed_at, processed_status, count(*) as cnt
from corp_processing
where 1 = 1
  and flow_name = 'tombstone-flow'
group by flow_name, flow_run_id, claimed_at, processed_status
order by flow_run_id, claimed_at desc
;
```
image

### DIFF
--- a/data-tool/flows/common/corp_processing_queue_service.py
+++ b/data-tool/flows/common/corp_processing_queue_service.py
@@ -1,0 +1,173 @@
+from enum import Enum
+from typing import List
+from sqlalchemy import text
+import logging
+
+class ProcessingStatuses(str, Enum):
+    PENDING = 'PENDING'
+    PROCESSING = 'PROCESSING'
+    COMPLETED = 'COMPLETED'
+    FAILED = 'FAILED'
+
+class CorpProcessingQueueService:
+    def __init__(self, environment: str, db_engine, flow_name: str):
+        self.data_load_env = environment
+        self.db_engine = db_engine
+        self.flow_name = flow_name
+        self.logger = logging.getLogger(__name__)
+
+    def reserve_for_flow(self, base_query: str, flow_run_id: str) -> int:
+        """Reserve corporations for processing in a specific flow run.
+
+        Args:
+            base_query: SQL query that selects corps to be processed
+            flow_run_id: Unique identifier for this flow run
+
+        Returns:
+            Number of corporations successfully reserved for this flow
+        """
+        init_query = f"""
+        WITH candidate_corps AS ({base_query}),
+        available_corps AS (
+            SELECT corp_num, corp_type_cd
+            FROM candidate_corps
+            FOR UPDATE SKIP LOCKED
+        )
+        INSERT INTO corp_processing (
+            corp_num,
+            corp_type_cd,
+            flow_name,
+            processed_status,
+            environment,
+            flow_run_id,
+            create_date,
+            last_modified,
+            claimed_at
+        )
+        SELECT 
+            corp_num,
+            corp_type_cd,
+            :flow_name,
+            :status,
+            :environment,
+            :flow_run_id,
+            NOW(),
+            NOW(),
+            NULL
+        FROM available_corps
+        ON CONFLICT (corp_num, flow_name, environment) 
+        DO NOTHING
+        RETURNING corp_num
+        """
+
+        with self.db_engine.connect() as conn:
+            with conn.begin():
+                result = conn.execute(
+                    text(init_query),
+                    {
+                        'flow_name': self.flow_name,
+                        'status': ProcessingStatuses.PENDING,
+                        'environment': self.data_load_env,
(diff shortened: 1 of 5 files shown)


data-tool/flows/common/corp_processing_queue_service.py
argush3, line 1: there is probably a bit too much logging in here for now but i think for now it's ok while we test out the stability of the corp processing code

data-tool/flows/common/corp_processing_queue_service.py
```diff
@@ -0,0 +1,173 @@
...
+                return claimed_corps
+
+    def update_corp_status(
+        self,
+        flow_run_id: str,
+        corp_num: str,
+        status: ProcessingStatuses,
+        error: str = None
+    ) -> bool:
+        """Update status for a corp."""
```
argush3, line 138: will make this more flexible and take in dynamic properties as we flesh this out in future work

data-tool/flows/common/corp_processing_queue_service.py
```diff
@@ -0,0 +1,173 @@
...
+
+        Returns:
+            Number of corporations successfully reserved for this flow
+        """
+        init_query = f"""
+        WITH candidate_corps AS ({base_query}),
+        available_corps AS (
+            SELECT corp_num, corp_type_cd
+            FROM candidate_corps
+            FOR UPDATE SKIP LOCKED
```
argush3, line 34: temporarily locks rows until these corps are reserved for the current flow run

data-tool/flows/common/corp_processing_queue_service.py
```diff
@@ -0,0 +1,173 @@
...
+            AND environment = :environment
+            AND flow_name = :flow_name
+            AND flow_run_id = :flow_run_id
+            AND claimed_at IS NULL
+            LIMIT :batch_size
+            FOR UPDATE SKIP LOCKED
+        )
+        UPDATE corp_processing
+        SET processed_status = :processing_status,
+            claimed_at = NOW(),
```
argush3, line 102: `claimed_at` is key field to determined when corps in a given flow(`flow_run_id`) were picked up by a batch

data-tool/flows/corps_tombstone_flow.py
```diff
@@ -327,6 +342,11 @@ def tombstone_flow():
                     corp_futures.append(
                         migrate_tombstone.submit(config, lear_engine, corp_num, clean_data, users_mapper)
                     )
+                    processing_service.update_corp_status(
+                        flow_run_id,
+                        corp_num,
+                        ProcessingStatuses.COMPLETED
+                    )
```
chenhongjing, line 349: I think each update of status had better move to the point after each corp_future is complete (tweak line 350+). And it will be something like
```python
for f in corp_futures:
    corp_num = f.result()
    if corp_num:
        processing_service.update_corp_status(...)
    else:
       # error handling? 
```

chenhongjing (approved):
LGTM! 😸

PR #3120, opened 2024-12-11
24825 Business-Pay - update to send pub/sub message to filer

chenhongjing:
*Issue #:* /bcgov/entity#24825

*Description of changes:*
- Update config
- Update to publish Pub/Sub message to filer
- Comment out NATS message publishing to filer

*Local Testing*
Run [Pub/Sub emulator](https://cloud.google.com/pubsub/docs/emulator), and comment out the `verify_gcp_jwt()`
- Pub/Sub
image

- Feature Flag
image

### DIFF
--- a/queue_services/business-pay/src/business_pay/resources/pay_filer.py
+++ b/queue_services/business-pay/src/business_pay/resources/pay_filer.py
@@ -47,7 +47,7 @@ from simple_cloudevent import SimpleCloudEvent, to_queue_message
 from structured_logging import StructuredLogging
 
 from business_pay.database import Filing
-from business_pay.services import (create_email_msg, create_filing_msg,
+from business_pay.services import (create_email_msg, create_filing_msg, flags,
                                    gcp_queue, queue, verify_gcp_jwt)
 
 bp = Blueprint("worker", __name__)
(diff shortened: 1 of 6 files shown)


queue_services/business-pay/src/business_pay/resources/pay_filer.py
```diff
@@ -174,17 +171,38 @@ def publish_to_filer(filing: Filing, payment_token: PaymentToken):
...
+            # queue_message = create_filing_msg(filing.id)
+            # logger.debug(f"filer queue_message: {queue_message}")
+
+            # try:
+            #     # await queue.publish(subject=filer_topic, msg=queue_message)
+            #     queue.publish_json(subject=filer_topic, payload=queue_message)
+            # except Exception as err:
+            #     logger.debug(
+            #         f"Publish to Filer error: {err}, for pay-id: {payment_token.id}")
+
```
vysakh-menon-aot, line 205: We cannot merge back to main If we comment this. Can you change this code in such a way that if the `enable-sandbox` FF is true then publish to gcp else nats.

PS: I created `enable-sandbox` FF yesterday.

chenhongjing: Updated.

I don't import Flags service from legal-api because there're lots of dependency conflicts and I assume we don't want to downgrade Flask3 in business-pay. So I created a new serivce in this component with some tweaks of the original legal-api one, since LaunchDarkly SDK of business-pay is also in higher version.

vysakh-menon-aot (approved):

argush3 (approved):

queue_services/business-pay/src/business_pay/resources/pay_filer.py
chenhongjing: FYI about Flask3 updates if anyone has the error like `app.env does not exist` etc. in the future
image

PR #3136, opened 2024-12-18
22169 - api specs notice of withdrawal

eason-pan-bc:
*Issue #:* /bcgov/entity#22169

*Description of changes:*
**Adding items listed below to the api specs**
- Notice of Withdrawal Schema
- Sample request
- Sample responses

**Specs Render Screenshots:**
- Overview
image
- Sample Request
image
- Sample Request with Options
image
- Sample Response - successful
image
- Sample Response - 400
image
- Sample Response - 401
image
- Sample Response - 404
image

### DIFF
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -1523,7 +1523,6 @@ paths:
                       courtOrder:
                         fileNumber: "A12345"
                         effectOfOrder: planOfArrangement
-                        effectOfArrangement: ''
               voluntary-dissolution-request:
                 summary: Voluntary Dissolution Request
                 value:
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -5387,6 +5387,7 @@ components:
             - dissolution
             - dissolved
             - incorporationApplication
+            - noticeOfWithdrawal
             - putBackOn
             - registration
             - restoration
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -22,6 +22,7 @@ info:
     - Consent Continuation Out 
     - Conversion
     - Dissolution
+    - Notice of Withdrawal
     - Registration (Sole Proprietorship, General Partnership)
     - Special Resolution
 
@@ -696,6 +697,35 @@ paths:
                         paymentToken: '12345'
                         status: 'PENDING'
                         submitter: 'mocked submitter'
+                notice-of-withdrawal-success-response:
+                  summary: Notice of Withdrawal Response
+                  value:
+                    filing:
+                      business:
+                        foundingDate: '2023-07-12T17:31:58.000+00:00'
+                        identifier: BC1234567
+                        legalName: 1234567 B.C. LTD.
+                        legalType: BC
+                      header:
+                        affectedFilings: []
+                        availableOnPaperOnly: false
+                        certifiedBy: Sample Certified Person
+                        colinIds: []
+                        comments: []
+                        data: '2024-12-18T00:10:47.042797+00:00'
+                        email: api.specs@example.com
+                        filingId: 654321
+                        inColinOnly: false
+                        isCorrected: false
+                        isCorrectionPending: false
+                        isPaymentActionRequired: false
+                        name: noticeOfWithdrawal
+                        paymentStatusCode: 'APPROVED'
+                        paymentToken: '12345'
+                        status: 'PENDING'
+                        submitter: 'mocked submitter'
+                      noticeOfWithdrawal:
+                        filingId: 123456
                 voluntary-dissolution-success-response:
                   summary: Voluntary Dissolution Response
                   value:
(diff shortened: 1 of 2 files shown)


severinbeauvais:
Shouldn't it be 404 if not found?

severinbeauvais:
What about the optional data that could be posted / fetched? Like the 2 checkbox values and the court order.

eason-pan-bc:
> Shouldn't it be 404 if not found?

Good point, we will need to adjust the validation code a bit to make it happen. It's kinda a straight forward change, should I change it and include it in this PR/ticket.

eason-pan-bc:
> What about the optional data that could be posted / fetched? Like the 2 checkbox values and the court order.

Gonna add this example

severinbeauvais:
> > Shouldn't it be 404 if not found?
> 
> Good point, we will need to adjust the validation code a bit to make it happen. It's kinda a straight forward change, should I change it and include it in this PR/ticket.

How about: update the spec now and update the code in one of the existing Legal API tickets?

eason-pan-bc:
> > > Shouldn't it be 404 if not found?
> > 
> > 
> > Good point, we will need to adjust the validation code a bit to make it happen. It's kinda a straight forward change, should I change it and include it in this PR/ticket.
> 
> How about: update the spec now and update the code in one of the existing Legal API tickets?

Sounds good, added this task to [24989](https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/24989)

ArwenQin:
I would like to confirm the schema for the two checkbox values. Should they be placed within a sub-object or not?
"noticeOfWithdrawal": {
       "filingId": "153004",
       "courtOrder": {
           "fileNumber": "TT0010045",
           "effectOfOrder": "planOfArrangement"
       },
       "planOfArrangement": {
             "partOfPoa": true,
             "comeIntoEffect": true
   }
}

Or 
"noticeOfWithdrawal": {
       "filingId": "153004",
       "courtOrder": {
           "fileNumber": "TT0010045",
           "effectOfOrder": "planOfArrangement"
       },
         "partOfPoa": true,
         "comeIntoEffect": true
}

eason-pan-bc:
> "noticeOfWithdrawal": { "filingId": "153004", "courtOrder": { "fileNumber": "TT0010045", "effectOfOrder": "planOfArrangement" }, "planOfArrangement": { "partOfPoa": true, "comeIntoEffect": true } }
> 
> Or "noticeOfWithdrawal": { "filingId": "153004", "courtOrder": { "fileNumber": "TT0010045", "effectOfOrder": "planOfArrangement" }, "partOfPoa": true, "comeIntoEffect": true }

I'm a bit confused, I only saw `courtOrder` as part of the NoW schema. https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/notice_of_withdrawal.json#L21C11-L21C122
Is there a NoW schema update? @severinbeauvais @ArwenQin

ArwenQin:
Yes, the schema may need to be updated, as there are 2 new check boxes properties for NoW. 
For now, We are not sure whether we need to include the 2 properties, and whether in a sub-object.
Severin and I discussed, for now, we say Yes

severinbeauvais:
> Is there a NoW schema update? @severinbeauvais @ArwenQin

Not yet. The courtOrder object is still there, and the 2 new boolean values may be optional or used for UI only. Still undecided but we don't need to add those to the schema yet.

eason-pan-bc:
> Not yet. The courtOrder object is still there, and the 2 new boolean values may be optional or used for UI only. Still undecided but we don't need to add those to the schema yet.
> 
> Update: Yes we will use the 2 new properties. I prefer them in the root of the noticeOfWithdrawal object.

Since the api specs follows the schemas, how about I add the option of `courtOrder` for now. If we ended up updating the NoW schema, we can always come back to update the specs?

severinbeauvais:
@ArwenQin Speaking of these 2 new properties, I think I was talking with Mihai and we were unsure whether we should save them or not.

On the one hand, staff should know that if either value is True then the NoW cannot be filed, so they won't even try.

On the other hand, if this filing ever becomes self-serve (ie, not staff-only) then we would use these values to prevent the NoW filing.

Please follow up with Mihai and Jacqueline on this. If staff don't need the properties then there's no need to save them. If regular users won't be allowed to file a NoW when either propery is True then, again, there's no need to save them.

severinbeauvais:
> Since the api specs follow the schemas, how about I add the option of `courtOrder` for now. If we ended up updating the NoW schema, we can always come back to update the specs?

Sure. The courtOrder object is optional and we haven't decided about the 2 new properties yet. They may turn out to be for UI use only, so should not be part of the schema/spec.

eason-pan-bc:
added the sample request with options, and updated the filing not found to 404

docs/business.yaml
```diff
@@ -1498,6 +1503,27 @@ paths:
...
+                      foundingDate: '2023-07-12T17:31:58.000+00:00'
+                      identifier: BC1234567
+                      legalName: 1234567 B.C. LTD.
+                      legalType: BC
+                    noticeOfWithdrawal:
+                      filingId: 123456
+                      courtOrder:
+                        fileNumber: "A12345"
+                        effectOfOrder: planOfArrangement
+                        effectOfArrangement: ''
```
severinbeauvais, line 1526: We aren't using `effectOfArrangement` and it isn't a court order property anyway. Remove it.

Ref: https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/court_order.json

eason-pan-bc: updated

severinbeauvais (approved):

vysakh-menon-aot (approved):

PR #3139, opened 2024-12-19
24952 - Fix not allow to update NoW draft issue

eason-pan-bc:
*Issue #:* /bcgov/entity#24952

*Description of changes:*
Updated the logic for allowable actions to enable update Notice of Withdrawal draft. This change is only within scope for regular/existing businesses. The support of allowable actions for temporary business will be done in https://github.com/bcgov/entity/issues/24987

#### Issue Before
- Can file NoW and NoW draft
- But cannot update the draft using `PUT` method with endpoint `/api/v2/businesses/:identifier/filings/:filing_id?draft=true`

#### After Update
Able to update NoW draft using the endpoint and method mentioned above.
- Create a Draft
image
- Update this Draft
image

#### Other Verifications
- Allowable Actions: there is no NoW or NoW draft, NoW filing is allowed
image
- Allowable Action: after a NoW draft is created, NoW filing is not allowed anymore
image
- Try to create the **second** draft when there is already a draft exist is now allowed
image

### DIFF
--- a/legal-api/src/legal_api/services/authz.py
+++ b/legal-api/src/legal_api/services/authz.py
@@ -746,7 +746,7 @@ def business_blocker_check(business: Business, is_ignore_draft_blockers: bool = 
     if business.in_dissolution:
         business_blocker_checks[BusinessBlocker.IN_DISSOLUTION] = True
 
-    if has_notice_of_withdrawal_filing_blocker(business):
+    if has_notice_of_withdrawal_filing_blocker(business, is_ignore_draft_blockers):
         business_blocker_checks[BusinessBlocker.FILING_WITHDRAWAL] = True
 
     return business_blocker_checks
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/services/authz.py
```diff
@@ -877,8 +877,24 @@ def has_notice_of_withdrawal_filing_blocker(business: Business):
     if business.admin_freeze:
         return True
 
-    filing_statuses = [Filing.Status.DRAFT.value,
-                       Filing.Status.PENDING.value,
+    # Get request method and filing id from request context
```
vysakh-menon-aot, line 880: These checks are not required

eason-pan-bc: Notice of Withdrawal filing is supposed to withdrawal future effective filings. That's why there is a special check for notice of withdrawal. (this was the previous ticket for allowable actions for NoW https://github.com/bcgov/entity/issues/23348, and this was the PR https://github.com/bcgov/lear/pull/3014)

The issue this PR is trying to address is that with the previous implementation, we cannot update the NoW draft using  `PUT` method with endpoint `/api/v2/businesses/:identifier/filings/:filing_id?draft=true`. And this PR is trying to make this happen and keep all other checks remain the same.

vysakh-menon-aot: I did not mean that this function is not required. We only need this check
```
    def has_notice_of_withdrawal_filing_blocker(business: Business):
        """Check if there are any blockers specific to Notice of Withdrawal."""
        if business.admin_freeze:
            return True
      
        now = datetime.now(timezone.utc)
        paid_filings = Filing.get_filings_by_status(business.id, [Filing.Status.PAID.value])
        return not any(f.effective_date and f.effective_date > now for f in paid_filings)
```

eason-pan-bc: Thanks for clarification, I tried that, but it ended up to allow filing multiple drafts

vysakh-menon-aot: Ok I missed that scenario. Let me check

vysakh-menon-aot: Use `is_ignore_draft_blockers` (https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/authz.py#L765) similar to `has_blocker_filing` function

eason-pan-bc: Good call, gonna try it out.

eason-pan-bc: It works, thanks for the inspiration, love it

vysakh-menon-aot (approved):

AimeeGao (approved):
Beautiful solutions! LGTM (^-^)

PR #3158, opened 2025-01-02
Correction Details Added on Register Correction Application pdf

Rajandeep98:
*Issue #:* [/bcgov/entity/24946](https://github.com/bcgov/entity/issues/24946)

*Description of changes:*

Show Details of Correction change on Register Correction Application pdf

### DIFF
--- a/legal-api/report-templates/correction.html
+++ b/legal-api/report-templates/correction.html
@@ -53,8 +53,8 @@
       </div>
       <div class="container business-details-table" >
          <div class="bold"><span class="details-header-text"> Correction Detail</span><span class="correction-label-added">ADDED</span></div>
-         <div class="pt-2">Correction filed by {{correction.type}} on {{effective_date_time}}</div>
-         <div class="pt-2"><span class="bold">{{correction.comment.split('\n')[0]}}. </span><span>{{correction.comment.split('\n')[1]}}</span></div>
+         <div class="pt-2">Correction filed by Registry Staff on {{effective_date_time}}</div>
+         <div class="pt-2 preserve-line-breaks"><span>{{correction.comment}}</span></div>
       </div>
    </body>
 </html>
--- a/legal-api/report-templates/correction.html
+++ b/legal-api/report-templates/correction.html
@@ -51,5 +51,10 @@
          [[correction/rulesMemorandum.html]]
          [[correction/resolution.html]]
       </div>
+      <div class="container business-details-table" >
+         <div class="bold"><span class="details-header-text"> Correction Detail</span><span class="correction-label-added">ADDED</span></div>
+         <div class="pt-2">Correction filed by {{correction.type}} on {{effective_date_time}}</div>
+         <div class="pt-2"><span class="bold">{{correction.comment.split('\n')[0]}}. </span><span>{{correction.comment.split('\n')[1]}}</span></div>
+      </div>
    </body>
 </html>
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/version.py
```diff
@@ -22,4 +22,4 @@
 Development release segment: .devN
 """
 
-__version__ = '2.135.0'  # pylint: disable=invalid-name
+__version__ = '2.136.0'  # pylint: disable=invalid-name
```
severinbeauvais, line 25: This is normally done right before this branch (main) is deployed to Test. Vysakh is handling the deployments this time around. Please check with him on whether to leave this in or not.

legal-api/report-templates/correction.html
```diff
@@ -51,5 +51,10 @@
          [[correction/rulesMemorandum.html]]
          [[correction/resolution.html]]
       </div>
+      <div class="container business-details-table" >
+         <div class="bold"><span class="details-header-text"> Correction Detail</span><span class="correction-label-added">ADDED</span></div>
+         <div class="pt-2">Correction filed by {{correction.type}} on {{effective_date_time}}</div>
```
severinbeauvais, line 56: What are the different values for `type`?

In the UI, if the current user is staff then we show the staff name here; if the current user is not staff then we show "by Registry Staff".

image

In this case, to prevent outputs from accidentally showing a staff's name, I think this can simply be hard-coded to "by Registry Staff" (or "by Registries Staff" if you follow the UI example above).

legal-api/src/legal_api/version.py
vysakh-menon-aot: No need to update. Can revert this change

legal-api/report-templates/correction.html
```diff
@@ -51,5 +51,10 @@
...
+         <div class="pt-2"><span class="bold">{{correction.comment.split('\n')[0]}}. </span><span>{{correction.comment.split('\n')[1]}}</span></div>
```
vysakh-menon-aot, line 57: Its better to not access lines with index. Try this (add `<br>` if needed)
```suggestion
         <div class="pt-2">
         <span class="bold">
          {% for line in correction.comment.split('\n') %}
              {{ line }}
          {% endfor %}
          </span>
          </div>
```

legal-api/report-templates/correction.html
severinbeauvais: We talked about this. Resolving.

legal-api/report-templates/correction.html
```diff
@@ -51,5 +51,10 @@
          [[correction/rulesMemorandum.html]]
          [[correction/resolution.html]]
       </div>
+      <div class="container business-details-table" >
+         <div class="bold"><span class="details-header-text"> Correction Detail</span><span class="correction-label-added">ADDED</span></div>
+         <div class="pt-2">Correction filed by Registry Staff on {{effective_date_time}}</div>
+         <div class="pt-2 preserve-line-breaks"><span>{{correction.comment}}</span></div>
```
severinbeauvais, line 57: Do you have an example of this? Please verify with Jacqueline. I thought we said we'd bold the first comment line but maybe it's OK like this.

Rajandeep98: image

severinbeauvais: This looks good to me but please verify with Jacqueline.

vysakh-menon-aot (approved):

severinbeauvais (approved):

PR #3169, opened 2025-01-09
24715 Updates to support legacy ledger display name

chenhongjing:
*Issue #:* /bcgov/entity#24715

*Description of changes:*
- Add legacy ledger display name mapping
- Populate `colinDisplayName` in filing.meta_data
- Update legal-api to return legacy display name

*Pending Issue*
- CoD

*Testing*

### DIFF
--- a/legal-api/src/legal_api/core/meta/filing.py
+++ b/legal-api/src/legal_api/core/meta/filing.py
@@ -655,6 +655,10 @@ class FilingMeta:  # pylint: disable=too-few-public-methods
     @staticmethod
     def display_name(business: Business, filing: FilingStorage) -> Optional[str]:
         """Return the name of the filing to display on outputs."""
+        # if filing is imported from COLIN and has custom disaply name
+        if filing.meta_data and filing.meta_data.get('isLedgerPlaceholder'):
+            if displayName := filing.meta_data.get('colinDisplayName'):
+                return displayName
         # if there is no lookup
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/core/meta/filing.py
```diff
@@ -655,6 +655,10 @@ class FilingMeta:  # pylint: disable=too-few-public-methods
     @staticmethod
     def display_name(business: Business, filing: FilingStorage) -> Optional[str]:
         """Return the name of the filing to display on outputs."""
+        # if filing is imported from COLIN and has custom disaply name
+        if filing.meta_data and filing.meta_data.get('isLedgerPlaceholder') and\
+                (display_name := filing.meta_data.get('colinDisplayName')):
+            return display_name
```
argush3, line 661: i don't think we can rely on `isLedgerPlaceholder`.  ideally, we eventually remove `isLedgerPlaceholder` when we finish backfilling data for a migrated corp or do it as clean up after all corps are migrated over.  

I think if `colinDisplayName` exists, it should be sufficient condition for returning the colin specific display name.

argush3 (commented):
looks good.  

just the one comment.

I think you mentioned this already but just wanted to confirm.  No changes required for the FE right?

chenhongjing:
> looks good.
> 
> just the one comment.
> 
> I think you mentioned this already but just wanted to confirm. No changes required for the FE right?

Yes, not required for FE.

argush3 (approved):

PR #3172, opened 2025-01-15
25252 - Manual approach to populate TING info

chenhongjing:
*Issue #:* /bcgov/entity#25252

*Description of changes:*
- add subqueries to select TING, TED and others (manual approach)
- fix some data mapping for amalgamation
- raise exception if TED is migrated before its corp TINGs
- fix delete scripit

### DIFF
--- a/data-tool/flows/batch_delete_flow.py
+++ b/data-tool/flows/batch_delete_flow.py
@@ -482,7 +482,10 @@ def execute_query(conn: Connection, template: dict) -> dict:
                 else:
                     # now only consider str and int in the list
                     v_str = ', '.join(map(lambda x: f'\'{x}\'' if isinstance(x, str) else str(x), filter_none(v)))
-                    query += f' AND {k} IN ({v_str})'
+                    if v_str:
+                        query += f' AND {k} IN ({v_str})'
+                    else:
+                        query += ' AND 1 != 1'
 
     results = conn.execute(text(query))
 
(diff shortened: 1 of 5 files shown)


data-tool/flows/tombstone/tombstone_queries.py
```diff
@@ -1,5 +1,88 @@
...
+                    from t3
+                    where t3.corp_num = c.corp_num
+                )
+            """
+        }
+    ]
+    # change index to select subset of corps
+    # 0: all, 1: TING, 2: TED that linked TINGs are migrated, 3: exclude TING & TED
+    subquery = subqueries[2]
+    return subquery['cte'], subquery['where']
```
argush3, line 79: so with this approach, when we do the actual migration in prod, we would run in the following order? 1, 2, 3 or 1, 2, 0?

data-tool/flows/batch_delete_flow.py
```diff
@@ -482,7 +482,10 @@ def execute_query(conn: Connection, template: dict) -> dict:
                 else:
                     # now only consider str and int in the list
                     v_str = ', '.join(map(lambda x: f'\'{x}\'' if isinstance(x, str) else str(x), filter_none(v)))
-                    query += f' AND {k} IN ({v_str})'
+                    if v_str:
+                        query += f' AND {k} IN ({v_str})'
+                    else:
+                        query += ' AND 1 != 1'
```
argush3, line 488: thank for fixing this!

data-tool/flows/tombstone/tombstone_queries.py
chenhongjing: 1, 2, 3 or 2, 1, 3. But you may find fewer eligible corps using the second order. 
0 is for other purpose so that we can tweak queries easily.

argush3: can you add a comment similar to what you wrote here in the code ^^

argush3 (approved):

PR #3174, opened 2025-01-16
24482 Tombstone pipeline - Implement continuation in

leodube-aot:
*Issue #:* /bcgov/entity#24482

*Description of changes:*
- Implement continuation in tombstone pipeline


data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -226,6 +227,44 @@ def format_resolutions_data(data: dict) -> list[dict]:
...
+        formatted_jurisdiction['country'] = 'CA'
+        formatted_jurisdiction['region'] = 'FEDERAL' if can_jurisdiction_code == 'FD' else can_jurisdiction_code
+    # when other jurisdiction and len(othr_juris_desc) = 2, then othr_juris_desc is country code
+    elif can_jurisdiction_code == 'OT' and len(other_jurisdiction_desc) == 2:
+        formatted_jurisdiction['country'] = other_jurisdiction_desc
+    # when other jurisdiction and len(othr_juris_desc) = 6, then othr_juris_desc contains both
+    # region code and country code (like "US, SS"). Ignore any other cases.
+    elif can_jurisdiction_code == 'OT' and len(other_jurisdiction_desc) == 6:
+        formatted_jurisdiction['country'] = other_jurisdiction_desc[:2]
+        formatted_jurisdiction['region'] = other_jurisdiction_desc[4:]
```
leodube-aot, line 263: This logic is pulled from data_mapping.xlsx document.

Screenshot 2025-01-15 at 2 40 21 PM

data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -226,6 +227,44 @@ def format_resolutions_data(data: dict) -> list[dict]:
     return formatted_resolutions
 
 
+def format_jurisdictions_data(data: dict, event_id: Decimal) -> dict:
+    jurisdictions_data = data['jurisdictions']
+
+    matched_jurisdictions = [
+        item for item in jurisdictions_data if item.get('j_start_event_id') == event_id
+    ]
```
leodube-aot, line 235: In COLIN the `jurisdiction` table can have multiple entries with the same `CORP_NUM` but different `START_EVENT_ID`s, so we need to make sure that the jurisdiction entry matches the `event_id`.

data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -226,6 +227,44 @@ def format_resolutions_data(data: dict) -> list[dict]:
...
+    # when other jurisdiction and len(othr_juris_desc) = 2, then othr_juris_desc is country code
+    elif can_jurisdiction_code == 'OT' and len(other_jurisdiction_desc) == 2:
+        formatted_jurisdiction['country'] = other_jurisdiction_desc
+    # when other jurisdiction and len(othr_juris_desc) = 6, then othr_juris_desc contains both
+    # region code and country code (like "US, SS"). Ignore any other cases.
+    elif can_jurisdiction_code == 'OT' and len(other_jurisdiction_desc) == 6:
+        formatted_jurisdiction['country'] = other_jurisdiction_desc[:2]
+        formatted_jurisdiction['region'] = other_jurisdiction_desc[4:]
+
+    return formatted_jurisdiction
```
chenhongjing, line 265: Think you may need to populate values for `expro_identifier`. From what I check it's `A% ` businesses that registered in our system. As for `expro_legal_name`, not sure if we can get this value from extract db. If not, maybe we can leave it out now.

P.S. Data in dev extract seems different so you may find `A%` values in other places. Using test extract is better.

leodube-aot:
> You need to update business info for continuation in too. You can check [filer](https://github.com/bcgov/lear/blob/main/queue_services/entity-filer/src/entity_filer/filing_processors/continuation_in.py#L123) and [this tombstone mapping](https://github.com/bcgov/lear/blob/main/data-tool/flows/tombstone/tombstone_mappings.py#L382)

Updated to populate `expro_identifier` and updated business info. The new corps loaded into dev (from the test Colin extract):
```
C0693901 (C)
C0875873 (CCC)
C0816278 (CUL)
```

data-tool/flows/tombstone/tombstone_utils.py
leodube-aot: For `expro_legal_name` I couldn't find it anywhere in COLIN.

Looks like the `expro_legal_name` is only used in the [business summary report](https://github.com/bcgov/lear/blob/main/legal-api/report-templates/template-parts/business-summary/foreignJurisdiction.html#L10). 

Screenshot 2025-01-16 at 8 19 48 PM

This expro info (identifier and legal name) seems to be optional as not every continued in business has it.

Screenshot 2025-01-16 at 8 08 03 PM

The expro legal name can still be found in an MRAS lookup. I think we're okay to leave out trying to populate `expro_legal_name` entries for now, and if we deem it necessary down the line we can investigate either:

A. Finding out if this data is indeed stored somewhere in COLIN, or
B. Explore whether this info can be retrieved from MRAS

chenhongjing: Another option may be applicable...mark the information `Not Available`. I've marked amalgamation related information `Not Available` for business summary of tombstone corps.

leodube-aot: I like that route, I'll update it to do that.

leodube-aot: Updated to handle the case when `expro_identifier != NULL` and `expro_legal_name == NULL`, which should only occur due to the migration.

Screenshot 2025-01-17 at 10 44 11 AM

legal-api/src/legal_api/reports/business_document.py
```diff
@@ -550,6 +550,12 @@ def _set_continuation_in_details(self, business: dict):
                     'business_id': jurisdiction.business_id,
                     'filing_id': jurisdiction.filing_id,
                     }
+            
+            # Imported from COLIN
+            if continuation_in_filing.source == Filing.Source.COLIN.value:
```
chenhongjing, line 555: Think using `is_tombstone` to check will be better. [Reference](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/reports/business_document.py#L332)

chenhongjing (approved):

argush3 (approved):

PR #3180, opened 2025-01-23
25446 - add check for lear businesses before freezing corps

ketaki-deodhar:
*Issue #:* /bcgov/entity#25446

*Description of changes:*

freeze a business if it exists only in LEAR

Branched off from https://github.com/bcgov/lear/tree/release-2.136.0 (Sev created this yesterday)

- add service to call legal-api business endpoint
- add logic to freeze a business only if found in LEAR

BC Limited company created in in DEV:
https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0887528/?accountid=3040

Record in CDEV after COLIN sync was complete:
image

### DIFF
--- a/colin-api/src/colin_api/config.py
+++ b/colin-api/src/colin_api/config.py
@@ -93,6 +93,15 @@ class _Config:  # pylint: disable=too-few-public-methods
     except (TypeError, ValueError):
         JWT_OIDC_JWKS_CACHE_TIMEOUT = 300
 
+    # legal api
+    LEGAL_API_URL = os.getenv('LEGAL_API_URL')
+
+    # service accounts
+    ACCOUNT_SVC_AUTH_URL = os.getenv('ACCOUNT_SVC_AUTH_URL')
+    ACCOUNT_SVC_CLIENT_ID = os.getenv('ACCOUNT_SVC_CLIENT_ID')
+    ACCOUNT_SVC_CLIENT_SECRET = os.getenv('ACCOUNT_SVC_CLIENT_SECRET')
+    ACCOUNT_SVC_TIMEOUT = os.getenv('ACCOUNT_SVC_TIMEOUT')
+
     TESTING = False
     DEBUG = False
 
--- a/colin-api/src/colin_api/models/filing.py
+++ b/colin-api/src/colin_api/models/filing.py
@@ -1241,7 +1241,7 @@ class Filing:  # pylint: disable=too-many-instance-attributes;
 
     # pylint: disable=too-many-locals,too-many-statements,too-many-branches,too-many-nested-blocks;
     @classmethod
-    def add_filing(cls, con, filing: Filing, bc_identifier: str) -> int:
+    def add_filing(cls, con, filing: Filing, lear_identifier: str) -> int:
         """Add new filing to COLIN tables."""
         try:
             if filing.filing_type not in ['agmExtension', 'agmLocationChange', 'alteration',
@@ -1396,14 +1396,17 @@ class Filing:  # pylint: disable=too-many-instance-attributes;
                                                 Business.TypeCodes.BCOMP.value,
                                                 Business.TypeCodes.BCOMP_CONTINUE_IN.value,
                                             ])
+            is_business_in_lear = cls.is_business_in_lear(lear_identifier)
 
             # Freeze all entities except CP if business exists in lear and
             # 'enable-bc-ccc-ulc' flag is on else just freeze BEN
             is_frozen_condition = (
-                cls.is_business_in_lear(bc_identifier) and
                 flags.is_on('enable-bc-ccc-ulc') and
-                business['business']['legalType'] != Business.TypeCodes.COOP.value
+                business['business']['legalType'] != Business.TypeCodes.COOP.value and
+                is_business_in_lear
             )
+
+            current_app.logger.debug(f'Business {lear_identifier}, is_frozen_condition:{is_frozen_condition}')
 
             is_new_or_altered_ben = is_new_ben or is_new_cben or is_alteration_to_ben_or_cben
 
@@ -1418,9 +1421,9 @@ class Filing:  # pylint: disable=too-many-instance-attributes;
             raise err
 
     @classmethod
-    def is_business_in_lear(cls, bc_identifier: str) -> bool:
+    def is_business_in_lear(cls, lear_identifier: str) -> bool:
         """Check if business is in lear."""
-        response = LegalApiService.query_business(bc_identifier)
+        response = LegalApiService.query_business(lear_identifier)
 
         if response.status_code == HTTPStatus.OK:
             return True
--- a/colin-api/src/colin_api/models/filing.py
+++ b/colin-api/src/colin_api/models/filing.py
@@ -1403,8 +1403,8 @@ class Filing:  # pylint: disable=too-many-instance-attributes;
             # call to legal-api to check if business exists in lear
             response = LegalApiService.query_business(bc_identifier)
 
-            # Freeze all entities if found in LEAR and if 'enable-bc-ccc-ulc' flag is on else just freeze BEN
-            # exclude CP at all times
+            # Check business exists in lear
+            # If found in LEAR, freeze entities except CP if 'enable-bc-ccc-ulc' flag is on else just freeze BEN
             if response.status_code == HTTPStatus.OK:
                 is_frozen_condition = (
                     flags.is_on('enable-bc-ccc-ulc') and
--- a/colin-api/src/colin_api/models/filing.py
+++ b/colin-api/src/colin_api/models/filing.py
@@ -1397,6 +1397,7 @@ class Filing:  # pylint: disable=too-many-instance-attributes;
                                                 Business.TypeCodes.BCOMP_CONTINUE_IN.value,
                                             ])
             is_business_in_lear = cls.is_business_in_lear(lear_identifier)
+            current_app.logger.debug(f'Business {lear_identifier}, is_business_in_lear:{is_business_in_lear}')
 
             # Freeze all entities except CP if business exists in lear and
             # 'enable-bc-ccc-ulc' flag is on else just freeze BEN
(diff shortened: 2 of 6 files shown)


colin-api/src/colin_api/resources/filing.py
```diff
@@ -117,6 +117,9 @@ def post(legal_type, identifier, **kwargs):
                     {'message': 'Error: Identifier in URL does not match identifier in filing data'}
                 ), HTTPStatus.BAD_REQUEST
 
+            # setting this for lear business check as identifier is converted from lear to colin below
+            bc_identifier = identifier
```
ketaki-deodhar, line 121: had to set this here as identifier is converted from lear to colin later and is used subsequently in all calls

colin-api/src/colin_api/models/filing.py
```diff
@@ -1396,11 +1397,19 @@ def add_filing(cls, con, filing: Filing) -> int:
...
-            )
+            # default value set to False
+            is_frozen_condition = False
+
+            # call to legal-api to check if business exists in lear
+            response = LegalApiService.query_business(bc_identifier)
+
+            # Check business exists in lear
+            # If found in LEAR, freeze entities except CP if 'enable-bc-ccc-ulc' flag is on else just freeze BEN
+            if response.status_code == HTTPStatus.OK:
```
argush3, line 1408: i think you should create a separate function called `is_business_in_lear` that wraps the check and returns a bool.

And add the result of `is_business_in_lear` as one of the conditions in `is_frozen_condition`. 

Think it will be cleaner this way.

colin-api/src/colin_api/config.py
severinbeauvais, line 1: Is there an example env file that goes along with this?

colin-api/src/colin_api/services/legal.py
```diff
@@ -0,0 +1,42 @@
...
+
+        # Perform proxy call using the input identifier (e.g. BC 123456)
+        response = requests.get(legal_api_url + '/businesses/' + identifier,
+                                headers={'Content-Type': 'application/json', 'Authorization': 'Bearer ' + token},
+                                timeout=timeout
+                                )
+        try:
+            return response
+        except Exception:  # pylint: disable=broad-except
+            return None
```
argush3, line 42: think we need to think about if we are handling things correctly here.  

do we want to throw the exception up the chain if there's an exception?

colin-api/src/colin_api/models/filing.py
```diff
@@ -1414,6 +1417,15 @@ def add_filing(cls, con, filing: Filing) -> int:
...
             raise err
 
+    @classmethod
+    def is_business_in_lear(cls, bc_identifier: str) -> bool:
+        """Check if business is in lear."""
+        response = LegalApiService.query_business(bc_identifier)
+
+        if response.status_code == HTTPStatus.OK:
+            return True
+        return False
```
argush3, line 1427: if it's a 404, i understand returning false but if legal api is returning a 500 or something, i don't think we should be just assuming the business is not in LEAR?

colin-api/src/colin_api/services/legal.py
ketaki-deodhar: Yes. Looking into it

colin-api/src/colin_api/resources/filing.py
argush3: i think we should name it `lear_identifier` to make things clear.

colin-api/src/colin_api/config.py
severinbeauvais: I thought it was a requirement, but you're suggesting it's not.

Anyway, that would be nice to have. Just take your local .env file and strip out anything secret or obsolete.

severinbeauvais: ^^ If this will take some time, never mind (or do it later). Need your fix out there asap.

colin-api/src/colin_api/models/filing.py
```diff
@@ -1396,8 +1397,10 @@ def add_filing(cls, con, filing: Filing) -> int:
                                                 Business.TypeCodes.BCOMP_CONTINUE_IN.value,
                                             ])
 
-            # Freeze all entities except CP if 'enable-bc-ccc-ulc' flag is on else just freeze BEN
+            # Freeze all entities except CP if business exists in lear and
+            # 'enable-bc-ccc-ulc' flag is on else just freeze BEN
             is_frozen_condition = (
+                cls.is_business_in_lear(bc_identifier) and
```
argush3, line 1403: think this should be the last condition to potentially save having to call LEAR

colin-api/src/colin_api/models/filing.py
severinbeauvais: If the response is a 500 (or anything other than 200 or 404), can this flow be aborted and it will automatically retry later?

Otherwise, what should be done if this isn't a 200 or 404?

colin-api/src/colin_api/models/filing.py
argush3, line 1403: can we actually assign this to a variable and add two log statements?  

it would be nice to be able to tell what the value is for `is_business_in_lear` and `is_frozen_condition` from the logs.

colin-api/src/colin_api/models/filing.py
argush3: i think we can throw an exception in the `LegalApiService` if get business endpoint doesn't return a 404 or 200. 
 And if get business endpoint results in an exception we also re-throw the exception.

If we do this then, for non-404 and non-200 scenarios, it will just bubble up to the exception handler for the add filings endpoint.

https://github.com/bcgov/lear/blob/0502c9f58eb736822485e937001760ca26c4dc21/colin-api/src/colin_api/resources/filing.py#L196-L202

severinbeauvais: So you'd see an error in the logs and this filing wouldn't get processed (synced)?

argush3: yes, the filing wouldn't get synced if we can't determine whether the business in LEAR.  Either because get business endpoint is broken or legal api is down

colin-api/src/colin_api/models/filing.py
argush3: Actually, i'm contradicting myself as that would require doing the `is_business_in_lear` call upfront.

Maybe just log `is_frozen_condition`.

The http call in LEAR might be logged automatically anyways

ketaki-deodhar: okay

colin-api/src/colin_api/models/filing.py
severinbeauvais: Sounds good to me.

colin-api/src/colin_api/config.py
ketaki-deodhar: I will make a note and will do it later

colin-api/src/colin_api/resources/filing.py
ketaki-deodhar: Renamed

colin-api/src/colin_api/config.py
severinbeauvais: OK. Please close this conversation when you're done with it.

colin-api/src/colin_api/services/legal.py
```diff
@@ -0,0 +1,45 @@
...
+
+        try:
+            # Perform proxy call using the input identifier (e.g. BC 123456)
+            response = requests.get(legal_api_url + '/businesses/' + identifier,
+                                    headers={'Content-Type': 'application/json', 'Authorization': 'Bearer ' + token},
+                                    timeout=timeout
+                                    )
+            # If the status code is 200 or 404, return the response
+            if response.status_code in (200, 404):
+                return response
```
argush3, line 41: think we want to throw an exception if it's not 200 or 404 in the scenario when a valid response comes back.

``` python
# If the status code is 200 or 404, return the response
if response.status_code in (200, 404):
    return response
# If we get here, it's an unexpected status code, we should raise an exception
response.raise_for_status()
```

colin-api/src/colin_api/services/legal.py
```diff
@@ -0,0 +1,45 @@
...
+        except requests.exceptions.RequestException as err:
+            current_app.logger.error(err, exc_info=True)
+            return None, str(err)
+        return None
```
argush3, line 45: We'll want to re-throw the exception after logging here. 

```
except requests.exceptions.RequestException as err:
    current_app.logger.error(err, exc_info=True)
    raise  # re-throw the exception after logging
```

Also, don't think we need the `return None` at the end

colin-api/src/colin_api/models/filing.py
```diff
@@ -1395,13 +1396,18 @@ def add_filing(cls, con, filing: Filing) -> int:
...
+            # Freeze all entities except CP if business exists in lear and
+            # 'enable-bc-ccc-ulc' flag is on else just freeze BEN
             is_frozen_condition = (
                 flags.is_on('enable-bc-ccc-ulc') and
-                business['business']['legalType'] != Business.TypeCodes.COOP.value
+                business['business']['legalType'] != Business.TypeCodes.COOP.value and
+                is_business_in_lear
             )
 
+            current_app.logger.debug(f'Business {lear_identifier}, is_frozen_condition:{is_frozen_condition}')
```
argush3, line 1409: so since you end up doing the `cls.is_business_in_lear(lear_identifier)` call upfront, can you add another debug logging statement right after `is_business_in_lear = cls.is_business_in_lear(lear_identifier)`?

argush3:
we'll need to add a `entity-service-account` entry under `applications` in [vaults.json](https://github.com/bcgov/lear/blob/9543eff7d7a7b5de0aa9c7e4d7e6f261c9933893/colin-api/devops/vaults.json#L4-L10).

argush3 (approved):
looks good!

As per our conversation, we can't roll this out into dev for testing until I get the 1pass and OCP secrets setup.  Will let you know when we're good to deploy the hotifx branch!

PR #3186, opened 2025-01-27
24594 Tombstone pipeline - Implement conversion ledger

leodube-aot:
*Issue #:* /bcgov/entity#24594

*Description of changes:*
- Implemented conversion ledger in tombstone pipeline

_Testing:_
The following businesses are loaded in dev and affiliated to account 668.
```
BC0010385
BC0019921
```

### DIFF
--- a/data-tool/flows/tombstone/tombstone_mappings.py
+++ b/data-tool/flows/tombstone/tombstone_mappings.py
@@ -62,6 +62,9 @@ class EventFilings(str, Enum):
     FILE_CONTU = 'FILE_CONTU'
     FILE_CONTC = 'FILE_CONTC'
 
+    # Conversion
+    FILE_CONVL = 'FILE_CONVL'
+
     # Correction
     FILE_CO_AR = 'FILE_CO_AR'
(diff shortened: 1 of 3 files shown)


data-tool/flows/tombstone/tombstone_mappings.py
```diff
@@ -192,6 +195,8 @@ def has_value(cls, value):
     EventFilings.FILE_CONTU: 'continuationIn',
     EventFilings.FILE_CONTC: 'continuationIn',
 
+    EventFilings.FILE_CONVL: 'conversion',
```
chenhongjing, line 198: Can you change `conversion` to `conversionLedger` so that there won't be any conflicts with the subsequent CONV event mapping to `conversion`? The filing type spreadsheet is a bit misleading to show `conversion`, but it should be `conversionLedger`.

Others look good!

leodube-aot: Updated and confirmed that everything checks out in dev

argush3 (approved):

PR #3190, opened 2025-01-29
25393 - Suppress emails for system file BEN Corrections Statement

ketaki-deodhar:
*Issue #:* /bcgov/entity#25393

*Description of changes:*

A little bit of background for this change.
We worked on a Jupyter Notebook to run Corrections for BEN statements. This will be a system file Correction filing with No Fee. This just adds a correction statement as Comment. We need to suppress any emails generated for such Corrections filing.

### DIFF
--- a/queue_services/entity-filer/tests/unit/test_worker/test_correction_bcia.py
+++ b/queue_services/entity-filer/tests/unit/test_worker/test_correction_bcia.py
@@ -228,13 +228,7 @@ BC_CORRECTION_SHORT = {
             'details': 'First correction',
             'correctedFilingId': '123456',
             'correctedFilingType': 'incorporationApplication',
-            'comment': f'''Correction for Incorporation Application filed on 2025-01-01 \n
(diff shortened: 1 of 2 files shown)


queue_services/entity-filer/tests/unit/test_worker/test_correction_bcia.py
```diff
@@ -213,6 +213,26 @@
...
+            'legalType': 'BC'
+        },
+        'correction': {
+            'details': 'First correction',
+            'correctedFilingId': '123456',
+            'correctedFilingType': 'incorporationApplication',
+            'comment': 'Correction for Incorporation Application filed on 2025-01-01 by system'
+        }
+    }
+}
```
argush3, line 234: can we add even one small element to in this correction filing json so we can see that it actually is able to process changes in the correction json?

argush3:
can we fix linting issues?

argush3 (approved):

PR #3191, opened 2025-01-30
25583 - Return Documents for NoW when Status is COMPLETED

meawong:
*Issue #:* /bcgov/entity#25583 

*Description of changes:*
- Add case for NoW when in COMPLETED status 
- Update unit test

image

image

### DIFF
--- a/legal-api/src/legal_api/core/filing.py
+++ b/legal-api/src/legal_api/core/filing.py
@@ -531,7 +531,9 @@ class Filing:  # pylint: disable=too-many-public-methods
             Filing.FilingTypes.AGMEXTENSION.value,
             Filing.FilingTypes.AGMLOCATIONCHANGE.value,
         ]
-        if filing.status in (Filing.Status.PAID, Filing.Status.WITHDRAWN) and \
+        if (filing.status in (Filing.Status.PAID, Filing.Status.WITHDRAWN) or
+                (filing.status == Filing.Status.COMPLETED and
+                    filing.filing_type == Filing.FilingTypes.NOTICEOFWITHDRAWAL.value)) and \
             not (filing.filing_type in no_legal_filings_in_paid_withdrawn_status
                  or (filing.filing_type == Filing.FilingTypes.DISSOLUTION.value and
                      business.legal_type in [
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/core/filing.py
```diff
@@ -531,7 +531,9 @@ def get_document_list(business,  # pylint: disable=too-many-locals disable=too-m
             Filing.FilingTypes.AGMEXTENSION.value,
             Filing.FilingTypes.AGMLOCATIONCHANGE.value,
         ]
-        if filing.status in (Filing.Status.PAID, Filing.Status.WITHDRAWN) and \
+        if (filing.status in (Filing.Status.PAID, Filing.Status.WITHDRAWN) or
```
argush3, line 534: is `WITHDRAWN` actually a filing status for the filing being withdrawn?  I don't think I see this status for latest withdrawn filings in dev.  Or is that

legal-api/tests/unit/resources/v2/test_business_filings/test_filing_documents.py
```diff
@@ -1687,7 +1687,7 @@ def test_temp_document_list_for_now(mocker, session, client, jwt,
     withdrawn_filing.save()
     filing = factory_filing(business, filing_json, filing_date=filing_date)
     filing.skip_status_listener = True
-    filing._status = Filing.Status.PAID
+    filing._status = Filing.Status.COMPLETED
```
argush3, line 1690: so is it the case there are no outputs available when the Now filing is in PAID status?

meawong: They should be available for `PAID `as well, I have updated the unit tests.

legal-api/tests/unit/resources/v2/test_business_filings/test_filing_documents.py
```diff
@@ -1703,3 +1703,17 @@ def test_temp_document_list_for_now(mocker, session, client, jwt,
 
     assert rv.status_code == expected_http_code
     assert rv_data == expected
+
+    filing._status = Filing.Status.PAID
```
argush3, line 1707: this should be `COMPLETED` right?

Also, the expected documents returned are the same for both `PAID` and `COMPLETED` status?

argush3 (approved):

ArwenQin (approved):
Looks good!

kzdev420 (approved):

eason-pan-bc (approved):

PR #3207, opened 2025-02-07
23352 update_filing_status_listen

kzdev420:
*Issue #:* /bcgov/entity#23352

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/models/filing.py
+++ b/legal-api/src/legal_api/models/filing.py
@@ -1158,7 +1158,8 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
         """Set review decision."""
         if filing_status not in [Filing.Status.CHANGE_REQUESTED.value,
                                  Filing.Status.APPROVED.value,
-                                 Filing.Status.REJECTED.value]:
+                                 Filing.Status.REJECTED.value,
+                                 Filing.Status.WITHDRAWN.value]:
             raise BusinessException(
                 error=f'Cannot set this filing status {filing_status}.',
                 status_code=HTTPStatus.FORBIDDEN
@@ -1219,7 +1220,8 @@ def receive_before_change(mapper, connection, target):  # pylint: disable=unused
     # pylint: disable=protected-access
     if (filing._status in [Filing.Status.AWAITING_REVIEW.value,
                            Filing.Status.CHANGE_REQUESTED.value,
-                           Filing.Status.REJECTED.value] or
+                           Filing.Status.REJECTED.value,
+                           Filing.Status.WITHDRAWN.value] or
             (filing._status == Filing.Status.APPROVED.value and not filing.payment_token)):
         return  # should not override status in the review process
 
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/models/filing.py
```diff
@@ -1158,7 +1158,8 @@ def set_review_decision(self, filing_status):
         """Set review decision."""
         if filing_status not in [Filing.Status.CHANGE_REQUESTED.value,
                                  Filing.Status.APPROVED.value,
-                                 Filing.Status.REJECTED.value]:
+                                 Filing.Status.REJECTED.value,
```
eason-pan-bc, line 1161: emmm, since WITHDRAWN status isn't from a staff review. It doesn't make too much sense to have it inside the `set_review_decision`. Maybe have a new function, like, `set_withdrawn_result` and have similar logic to the review_decision function?

eason-pan-bc (commented):
Also, the unit test didn't pass

legal-api/src/legal_api/models/filing.py
eason-pan-bc: And if you made this change, you may want to call this function in the NoW flier processor, instead of manually change filing status and save; otherwise, it won't work

kzdev420: Updated

legal-api/src/legal_api/models/filing.py
```diff
@@ -1219,7 +1219,8 @@ def receive_before_change(mapper, connection, target):  # pylint: disable=unused
     # pylint: disable=protected-access
     if (filing._status in [Filing.Status.AWAITING_REVIEW.value,
                            Filing.Status.CHANGE_REQUESTED.value,
-                           Filing.Status.REJECTED.value] or
+                           Filing.Status.REJECTED.value,
+                           Filing.Status.WITHDRAWN.value] or
             (filing._status == Filing.Status.APPROVED.value and not filing.payment_token)):
         return  # should not override status in the review process
```
eason-pan-bc, line 1225: I think the logic should work.
Maybe update the comment, so ppl can understand the WITHDRAWN status, it's not in the review process.

queue_services/entity-filer/tests/unit/filing_processors/test_notice_of_withdrawal.py
```diff
@@ -25,10 +25,8 @@
 
 
 @pytest.mark.parametrize('test_name, withdrawal_pending,withdrawn_filing_status', [
-    ('Process the Filing', False, False),
-    ('Dont process the Filing', False, True),
-    ('Dont process the Filing', True, False),
-    ('Dont process the Filing', True, True),
+    ('Process the Filing', False, 'PAID'),
+    ('Process the Filing', False, 'COMPLETED')
```
eason-pan-bc, line 29: the unit test is a bit confusing for me, why do we have a withdrawn_filing_status as COMPLETED? did you mean the NoW filing status as COMPLETED at the end?

Also, since you've changed some logics in the worker.py, it's better to add a unit test for that.

Some of my thoughts
| Test case | withdrawal_pending | withdrawn_filing_status | In Which Unit Test |
| --- | --- | --- | --- |
| regular filing processed | FALSE | PAID | Need to add or update a test under tests/unit/test_worker |
| FE filing not processed | TRUE | PAID | Need to add or update a test under tests/unit/test_worker |
|  |  |  |  |
| FE filing status change to WITHDRAWN, withdrawal_pending flag change to False and NoW status to completed | TRUE | PAID | tests/unit/filing_processors/test_notice_of_withdrawal.py |

kzdev420: I will do it another PR

eason-pan-bc (approved):

PR #3214, opened 2025-02-10
24754 Db Versioning - Update one-to-one relationships and directly query list-type relationships

leodube-aot:
*Issue #:* /bcgov/entity#24754

*Description of changes:*
- Updated the sql_versioning library to populate one-to-one relationships. ~For some reason these updates do not work on relationships with a list component (one-to-many, many-to-one, etc.)~
- ~Update legal api to directly query list-type relationships on versioned records.~

_Update:_
- Got one-to-many and many-to-one relationships to work as well. Also got versioned relationships working.
- Removed the workarounds

### DIFF
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -37,7 +37,7 @@ from legal_api.models import (
     OfficeType,
     PartyRole,
     Resolution,
-    ShareClass
+    ShareClass,
 )
 from legal_api.models.business import ASSOCIATION_TYPE_DESC
 from legal_api.reports.registrar_meta import RegistrarInfo
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -33,8 +33,11 @@ from legal_api.models import (
     CorpType,
     Document,
     Filing,
+    Office,
     OfficeType,
     PartyRole,
+    Resolution,
+    ShareClass
 )
 from legal_api.models.business import ASSOCIATION_TYPE_DESC
 from legal_api.reports.registrar_meta import RegistrarInfo
@@ -836,8 +839,8 @@ class Report:  # pylint: disable=too-few-public-methods, too-many-lines
                 self._filing.transaction_id,
                 primary_or_holding_business.id)
         else:
-            officelist = primary_or_holding_business.offices.all()
-            for i in officelist:
+            offices = Office.get_all_by_business_id(primary_or_holding_business.id)
+            for i in offices.all():
                 if i.office_type in [OfficeType.REGISTERED, OfficeType.RECORDS]:
                     offices[i.office_type] = {}
                     for address in i.addresses:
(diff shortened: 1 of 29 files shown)


python/common/sql-versioning/tests/test_versioning.py
```diff
@@ -134,6 +143,22 @@ def test_versioning_insert(db, session):
...
+    user.locations = locations
+    session.add(user)
+    session.commit()
+
+    user_version = version_class(User)
+    result_revision = session.query(user_version)\
+        .filter(user_version.name=='user')\
+        .one_or_none()
+    assert result_revision.address == address
+    # assert result_revision.locations == locations
```
leodube-aot, line 160: These 2 tests illustrate the issue I'm running into with the list-type relationships. The first relationship test passes, however the second one fails when not commented.

legal-api/src/legal_api/reports/report.py
```diff
@@ -836,8 +839,8 @@ def _set_from_primary_or_holding_business_data(self, filing):  # pylint: disable
                 self._filing.transaction_id,
                 primary_or_holding_business.id)
         else:
-            officelist = primary_or_holding_business.offices.all()
-            for i in officelist:
+            offices = Office.get_all_by_business_id(primary_or_holding_business.id)
+            for i in offices.all():
```
leodube-aot, line 843: Thinking about this more, I don't think this workaround works as hoped. Consider the following case:

```
parent_v1
  -> first_child_v1
  -> second_child_v1

# some changes to parent and children ...

parent_v2
  -> first_child_v2
  -> second_child_v2
```

When we do `parent_v1.children` we expect to get `[first_child_v1, second_child_v1]`. However, by using a query we are always getting the latest `child` entries (v2 in this case). I think by using the `transaction_id` with the query we should be able to get the correct version for the `child` entries. 

I am not sure how this worked originally with sqlalchemy_continuum. Was the line `primary_or_holding_business.offices` returning versioned children results (ie. `first_child_v1`)?

chenhongjing: These codes from sqlalchemy_continuum may be helpful.

[replace static relationship with dynamic query](https://github.com/kvesteri/sqlalchemy-continuum/blob/master/sqlalchemy_continuum/relationship_builder.py#L359-L363)
[dymanic query](https://github.com/kvesteri/sqlalchemy-continuum/blob/master/sqlalchemy_continuum/relationship_builder.py#L218)

Another thing is that sqlalchemy_continuum builds relationships after all versioned classes are built, which is different from our method.

leodube-aot: Yeah I'm working on an approach based off the way 'sqlalchemy_continuum' does it, which is mainly a query based approach. I'm trying to adapt it to work with the approach we use. 

I think a query based approach is the only way to get [version relationships](https://sqlalchemy-continuum.readthedocs.io/en/latest/version_objects.html#version-relationships) as that info only persists via the database.

python/common/sql-versioning/tests/test_versioning.py
```diff
@@ -135,6 +144,53 @@ def test_versioning_insert(db, session):
...
+
+    user_version = version_class(User)
+    result_revisions = session.query(user_version)\
+        .filter(user_version.name=='user')\
+        .all()
+    
+    assert user.address.name == 'Some new address'
+    assert len(result_revisions) == 2
+    assert result_revisions[0].address.name == "Some address"
+    # assert result_revisions[1].address.name == "Some new address"
```
leodube-aot, line 191: Still working on getting this test working. Seems like updating the relationship on User doesn't persist on the versioned record (however the new versioned records are still created accordingly).

PR #3221, opened 2025-02-11
25678 Tombstone - Updates for involuntary dissolution

chenhongjing:
*Issue #:* /bcgov/entity#25678 /bcgov/entity#25672

*Description of changes:*
- Add in-dissolution information in tombstone data set
- Updates for conversion involuntary dissolution
- Update delete script
- Misc fix

Testing
The following businesses are loaded in dev and affiliated to account `3446` (add this account into involuntary dissolution filter so that things work, but the legal-api is buggy somewhere)
```python
# D1F
BC0588870 # by SYSD1
BC0546050 # by DISD1 (delay of dissolution 1)
BC0550140 # by DISD2 (delay of dissolution 2)
BC0670410 # by ADCORP
BC0600187 # by FILE (change of address)

# D2F
BC0585930 # by SYSD2
BC0703023 # by DISD1 (delay of dissolution 1)
BC0576833 # by DISD2 (delay of dissolution 2)
BC0927679 # by ADCORP
BC0546311 # by FILE (change of address)

# no records for D1T and D2T
```

```python
# CONVID1
BC0617448 # active
BC0618432 # historical

# CONVID2
BC0617257 # active
BC0618459 # historical
```

### DIFF
--- a/data-tool/flows/tombstone/tombstone_mappings.py
+++ b/data-tool/flows/tombstone/tombstone_mappings.py
@@ -216,8 +216,8 @@ EVENT_FILING_LEAR_TARGET_MAPPING = {
     EventFilings.CONVDSL_NULL: 'conversion',  # TODO: liquidation
     EventFilings.CONVDSO_NULL: ['conversion', ('dissolution', 'unknown')],
     EventFilings.CONVICORP_NULL: 'conversion',
-    EventFilings.CONVID1_NULL: 'conversion',  # TODO: related to invol dissolution
-    EventFilings.CONVID2_NULL: 'conversion',  # TODO: related to invol dissolution
+    EventFilings.CONVID1_NULL: ['conversion', 'putBackOn'],  # TODO: to confirm
+    EventFilings.CONVID2_NULL: ['conversion', 'putBackOn'],  # TODO: to confirm
     EventFilings.CONVILIQ_NULL: 'conversion',  # TODO: liquidation
     EventFilings.CONVLRSTR_NULL: ['conversion', ('restoration', 'limitedRestoration')],
     EventFilings.CONVNC_NULL: ['conversion', 'changeOfName'],
--- a/data-tool/flows/tombstone/tombstone_queries.py
+++ b/data-tool/flows/tombstone/tombstone_queries.py
@@ -720,7 +720,7 @@ def get_filing_comments_query(corp_num):
     return query
 
 
-def get_in_dissolution_query(corp_num, threshold_timestamp):
+def get_in_dissolution_query(corp_num):
     query = f"""
     select
         cs.corp_num         as cs_corp_num,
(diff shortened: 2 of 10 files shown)


data-tool/flows/tombstone/tombstone_mappings.py
```diff
@@ -216,8 +216,8 @@ def has_value(cls, value):
     EventFilings.CONVDSL_NULL: 'conversion',  # TODO: liquidation
     EventFilings.CONVDSO_NULL: ['conversion', ('dissolution', 'unknown')],
     EventFilings.CONVICORP_NULL: 'conversion',
-    EventFilings.CONVID1_NULL: 'conversion',  # TODO: related to invol dissolution
-    EventFilings.CONVID2_NULL: 'conversion',  # TODO: related to invol dissolution
+    EventFilings.CONVID1_NULL: ['conversion', 'putBackOn'],  # TODO: to confirm
+    EventFilings.CONVID2_NULL: ['conversion', 'putBackOn'],  # TODO: to confirm
```
chenhongjing, line 220: The legacy system behaviour for conversion in-dissolution status is weird. And David R has already asked home team to trace the code. I'll keep this change until being told to revert.

data-tool/flows/tombstone/tombstone_queries.py
```diff
@@ -720,6 +720,27 @@ def get_filing_comments_query(corp_num):
...
+        to_char(
+            e.trigger_dts::timestamp at time zone 'UTC', 'YYYY-MM-DD HH24:MI:SSTZH:TZM'
+        )                   as e_trigger_dts_str
+    from corp_state cs
+    join event e on e.event_id = cs.start_event_id
+    where 1 = 1
+    and cs.corp_num = '{corp_num}'
+    and cs.end_event_id is null
+    and cs.state_type_cd in ('D1F', 'D2F', 'D1T', 'D2T')
+    and e.trigger_dts > '{threshold_timestamp}' :: timestamptz
```
argush3, line 739: so i'm assuming we have corp state entries in D1T or D2T where the trigger date is before the threshold timestamp?  and if so, we have confirmed with Dylan what we're doing is ok?

chenhongjing: The decision log suggests that the current state for involuntary dissolution needs to be brought over. So I use this threshold to avoid duplicate processing in both LEAR and COLIN. But it depends on how we fill the gap between COLIN shutdown and migration dry run, which I still have no idea. Any thoughts?

argush3: As per our discussion, I think we can remove the threshold timestamp.  The scenario where LEAR and COLIN have duplicate involuntary dissolution processing shouldn't happen.  It would mean our COLIN extract is not the latest data once COLIN data is frozen for corps.

argush3 (approved):

PR #3231, opened 2025-02-14
24699 & 24700 Tombstone - distinguish FED and non-FED, implement Notice of Withdrawal

chenhongjing:
*Issue #:* /bcgov/entity#24699 /bcgov/entity#24700
 
*Description of changes:*
- Updates for `filing_date`, `completion_date`, `effective_date`
- Implement Notice of Withdrawal
~~- Need discussion for different COLIN/LEAR behaviours of withdrawn IA, amalgamation❗~~ DONE

_Testing_
load the following businesses are loaded in dev and affiliated to account `3446`
```python
BC0920360 # lots of FED CoA + 2 FED alteration
BC1489667 # FED followed by NoW

```

### DIFF
--- a/data-tool/flows/tombstone/tombstone_utils.py
+++ b/data-tool/flows/tombstone/tombstone_utils.py
@@ -278,6 +278,7 @@ def format_filings_data(data: dict) -> list[dict]:
     formatted_filings = []
     state_filing_idx = -1
     idx = 0
+    withdrawn_filing_idx = -1
     for x in filings_data:
         event_file_type = x['event_file_type']
         # TODO: build a new complete filing event mapper (WIP)
@@ -301,6 +302,7 @@ def format_filings_data(data: dict) -> list[dict]:
             filing_subtype = raw_filing_subtype
 
         effective_date = x['ce_effective_dt_str'] or x['f_effective_dt_str'] or x['e_event_dt_str']
+        filing_date = x['ce_effective_dt_str'] or x['e_event_dt_str']
         trigger_date = x['e_trigger_dt_str']
 
         filing_json, meta_data = build_filing_json_meta_data(raw_filing_type, filing_type, filing_subtype,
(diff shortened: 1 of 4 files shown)


data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -319,26 +321,43 @@ def format_filings_data(data: dict) -> list[dict]:
...
+                'incorporationApplication',
+                'continuationIn'
+            ]:
+                raise Exception('Stop migrating withdrawn corp')
+            status = 'WITHDRAWN'
+            completion_date = None
+            withdrawn_filing_idx = idx
+        else:
+            status = 'COMPLETED'
+            completion_date = effective_date
```
argush3, line 336: ok, so I think this looks ok.  It would be probably be nicer if we could distinguish this scenario without having to throw an exception but it works.

Have you tested this scenario out on some businesses that fit this scenario?

And I think you mentioned in the PR description that this scenario needs discussion?

argush3:
Looks good to me.  Just the one comment.

data-tool/flows/tombstone/tombstone_utils.py
chenhongjing: Yes I did test this scenario.

image
image
I'm thinking of how we can distinguish this scenario too. Right now it throws error during data collection so that this corp will be regarded as skip and updated as [FAILED](https://github.com/bcgov/lear/blob/7fc1cbaf4d8fc96ebbb676b87977c13c3064865d/data-tool/flows/corps_tombstone_flow.py#L422-L429) so the next migration step won't start. I've ever thought about the possibility to update it as SKIP, but considering there may be other failure scenario during data collection, I can't do it.

Besides, this reminds me of something that you mentioned - "prefect UI shows flows as green even when a task fails for a given business" in dry run. Yes, this happens, though I haven't looked into the details. One reason may be that I suppress some exception in the script. And one reason may be from the mechanism of how prefect marks a flow as complete without failure.

So I think we need to think about the error tracking more carefully and make every corner consistent.

As for the description that "withdrawn IA, amalgamation, continuation in(no record actually for the 3rd one)" needs discussion, I think we've talked a bit based on what Dylan said that we don't need such record migrated. Can we completely filter them out in another better way? Any thoughts?

argush3: let's leave it as is.  the error message will make it easy to filter out via the `corp_processing` table

argush3: yeah, i guess we could filter it out at extract level.  unless there are a ton of these type of withdrawals, it might not be worth the effort to filter it out of the extract.

chenhongjing: Very small number of filings for each IA/amalgamation filing type in COLIN, under 100.

argush3: ok, let's leave the PR as is then

PR #3233, opened 2025-02-19
24628 Get remove() function working for db versioned relationships

leodube-aot:
*Issue #:* /bcgov/entity#24628

*Description of changes:*
This fix ended up needing 3 things:
1. In order to use `<parent>.remove(<child>)` and have the child be deleted from the table, the relationship needs `cascade='delete-orphan'`. This is already the case for resolutions, but I added it to the versioning tests as well.
2. For whatever reason, the parent side of the relationship also needs `backref=<parent>`, otherwise we always get `operation_type=0` (created). This however results in the child version entries that were removed getting `operation_type=1` (updated).
3. In order to properly assign `operation_type=2` on cascade deleted orphans I needed to add some custom logic to versioning.

### DIFF
--- a/python/common/sql-versioning/sql_versioning/versioning.py
+++ b/python/common/sql-versioning/sql_versioning/versioning.py
@@ -82,7 +82,7 @@ def _is_session_modified(session):
     return False
 
 
-def _get_operation_type(session, obj, delete_orphan: False):
+def _get_operation_type(session, obj, delete_orphan=False):
     """Return the operation type for the given object within the session.
     
     :param session: The database session instance.
--- a/python/common/sql-versioning/sql_versioning/versioning.py
+++ b/python/common/sql-versioning/sql_versioning/versioning.py
@@ -47,6 +47,27 @@ def _is_obj_modified(obj):
     return False
 
 
+def _should_relationship_delete_orphan(session, obj):
+    """
+    Checks if:
+        1. This relationship is a many-to-one relationship
+        2. If the opposite direction one-to-many relationship parent object has changes
+        3. If the opposite direction one-to-many relationship has cascade=delete-orphan
+        
+    :param session: The database session instance.
+    :param obj: The object to inspect for changes.
+    :return: True if the above checks pass, otherwise False.
+    """
+    should_delete = False
+    for r in inspect(obj.__class__).relationships:
+        if r.direction.name == 'MANYTOONE' and r._reverse_property:
+            reverse_rel, *_ = r._reverse_property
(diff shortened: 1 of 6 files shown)


python/common/sql-versioning/sql_versioning/versioning.py
```diff
@@ -61,19 +82,20 @@ def _is_session_modified(session):
     return False
 
 
-def _get_operation_type(session, obj):
+def _get_operation_type(session, obj, delete_orphan: False):
```
chenhongjing, line 85: LGTM 😸 just a very tiny thing here
```suggestion
def _get_operation_type(session, obj, delete_orphan=False):
```

argush3:
@leodube-aot just to confirm, these updates you are making around cascades deletes, it's how things work for the old versioning?

python/common/sql-versioning/sql_versioning/versioning.py
leodube-aot: Ah good catch on this typo! Fixed

chenhongjing (approved):

leodube-aot:
> @leodube-aot just to confirm, these updates you are making around cascades deletes, it's how things work for the old versioning?

Yes, the way the old versioning works is when we call `<parent>.remove(<child>)`, the `child` gets deleted. A good example for this is `ShareClass`

```python
# in main branch
class Business(db.Model, Versioned):
      ...
      share_classes = db.relationship('ShareClass', lazy='dynamic', cascade='all, delete, delete-orphan')
```

During an alteration we [delete **all** existing share classes](https://github.com/bcgov/lear/blob/main/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py#L150-L154), and [replace them with the ones comes from the alteration filing](https://github.com/bcgov/lear/blob/main/queue_services/entity-filer/src/entity_filer/filing_processors/filing_components/shares.py#L50-L64) (even if from the UI perspective we don't actually remove a given share class).

From my testing I have these share classes created/altered for a business with FF OFF, and for a business with FF ON.
Screenshot 2025-02-20 at 9 33 49 AM

When we look at the versioning entries, the ones for the FF ON are not created correctly. Inspecting this further, they're also missing the `end_transaction_id` so I'll have to update this PR to get that working as well:
Screenshot 2025-02-20 at 9 36 49 AM

Ultimately the fact that the operation_type and end_transaction_id are not getting set properly on `<parent>.remove(<child>)` results in this issue when trying to access the Notice of Articles for the FF ON business
Screenshot 2025-02-20 at 9 58 33 AM

argush3 (approved):

PR #3253, opened 2025-02-25
25904 Tombstone pipeline - push contact email to auth

vysakh-menon-aot:
*Issue #:* /bcgov/entity#25904

*Description of changes:*

### DIFF
--- a/data-tool/flows/corps_tombstone_flow.py
+++ b/data-tool/flows/corps_tombstone_flow.py
@@ -1,3 +1,4 @@
+from http import HTTPStatus
 from pathlib import Path
 
 import math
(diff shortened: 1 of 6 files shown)


data-tool/flows/corps_tombstone_flow.py
```diff
@@ -296,12 +297,22 @@ def update_auth(conn: Connection, config, corp_num: str, tombstone_data: dict):
...
+        if entity_status == HTTPStatus.OK:
+            update_email_status = AuthService.update_contact_email(
+                config=config,
+                identifier=business_data['identifier'],
+                email=tombstone_data['admin_email']
+            )
+            if update_email_status != HTTPStatus.OK:
+                raise Exception(f"""Failed to update admin email in auth {business_data['identifier']}""")
+        else:
+            raise Exception(f"""Failed to create entity in auth {business_data['identifier']}""")
```
chenhongjing, line 315: @argush3 a bit hesitant here. Do you think it will be better to raise exception if auth updates fail? It will cause lear updates to rollback together.

argush3: yes, we want to rollbacks in LEAR if auth api calls fail.  It will make it easier to reload businesses that way as opposed to having an business in LEAR but missing auth entries if we need it.  

Is this question about the contact email?  Or the `AuthService.create_entity` block above?

chenhongjing: This is question is for both create entity and update email. But what about the failture of creating affilication?

argush3: i think all 3 need to rollback the LEAR data ideally.

Just a note that this is just useful for dev/test data loading purposes.  For the actual big data migration, we won't be calling auth api at all is what I'm working off of.  For the small number of early adopters we load, I think there is a good chance we create all the auth entries.

argush3 (approved):

chenhongjing (approved):

PR #3255, opened 2025-02-26
23001 - Reduce DB Versioning Debugging Logs to Minimal

eason-pan-bc:
*Issue #:* /bcgov/entity#23001 

*Description of changes:*
- Reduce DB Versioning Debugging Logs to Minimal
    - Only keep showing the current db-versioning type (new / old) to help us quickly identify whether FF switching works or not)

**For a same API call:**
**Before**
Pasted image 20250226083535

**After**
Pasted image 20250226094534

### DIFF
--- a/legal-api/src/legal_api/models/db.py
+++ b/legal-api/src/legal_api/models/db.py
@@ -19,7 +19,7 @@ from datetime import datetime
 
 from flask import current_app
 from flask_sqlalchemy import SignallingSession, SQLAlchemy
-from sql_versioning import TransactionManager, debug
+from sql_versioning import TransactionManager
 from sql_versioning import disable_versioning as _new_disable_versioning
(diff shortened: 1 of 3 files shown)


AimeeGao (commented):
Since we're no longer using debug decorator, should we also consider removing the [debugging.py ](https://github.com/bcgov/lear/blob/main/python/common/sql-versioning/sql_versioning/debugging.py)file?

chenhongjing (approved):

vysakh-menon-aot (approved):

leodube-aot (approved):

eason-pan-bc:
> Since we're no longer using debug decorator, should we also consider removing the [debugging.py ](https://github.com/bcgov/lear/blob/main/python/common/sql-versioning/sql_versioning/debugging.py)file?

I'd keep it there for now, in case we need it later. Since it doesn't throw anything out now

PR #3262, opened 2025-02-27
26262 - Minimize DB Versioning Logging

eason-pan-bc:
*Issue #:* /bcgov/entity#26262

*Description of changes:*
- Removed the remaining `debug` in `__all__`
- Change logging to only print out when a service initialized, and when switching versioning

**Screenshots after change:**
Legal-API:
Pasted image 20250228121225

Emailer:
Pasted image 20250228122223

Filer:
Pasted image 20250228140011

**Local Linting and Testing Results**
Since we currently have some issues in the legal-api CI, here are screenshots for local linting and testings resutls
image
image

### DIFF
--- a/legal-api/src/legal_api/models/db.py
+++ b/legal-api/src/legal_api/models/db.py
@@ -35,6 +35,7 @@ from sqlalchemy_continuum.manager import VersioningManager
 db = SQLAlchemy()  # pylint: disable=invalid-name
 IS_LOGGED = False
 
+
 class Transaction(db.Model):
     """This class manages the transaction."""
 
@@ -51,16 +52,17 @@ class Transaction(db.Model):
 
 
 def print_versioning_info():
-    """Print the current versioning status if not already printed.
+    """
+    Print the current versioning status if not already printed.
+
     This should only be called within an application context.
     """
-    global IS_LOGGED
-    
+    global IS_LOGGED  # pylint: disable=global-statement
+
     if not IS_LOGGED:
         try:
-            from legal_api.services import flags as flag_service
-            from flask import current_app
-            
+            from legal_api.services import flags as flag_service  # pylint: disable=import-outside-toplevel
+
             current_service = current_app.config.get('SERVICE_NAME')
             if current_service:
                 db_versioning = flag_service.value('db-versioning')
@@ -276,8 +278,10 @@ def setup_versioning():
     _new_enable_versioning(transaction_cls=Transaction)
     make_versioned(user_cls=None, manager=versioning_manager)
 
+
 # TODO: enable versioning switching
 # it should be called before data model initialized, otherwise, old versioning doesn't work properly
 setup_versioning()
+
 
 # make_versioned(user_cls=None, manager=versioning_manager)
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/models/db.py
```diff
@@ -50,6 +51,30 @@ class Transaction(db.Model):
     issued_at = db.Column(db.DateTime, default=datetime.utcnow, nullable=True)
 
 
+def print_versioning_info():
+    """
+    Print the current versioning status if not already printed.
+
+    This should only be called within an application context.
+    """
+    global IS_LOGGED  # pylint: disable=global-statement
```
chenhongjing, line 60: is it necessary when `init_db()` is only called once?

eason-pan-bc: Good call, yeah, you are right!! gonna remove this

eason-pan-bc:
> Restarting the failed checks keeps using `setup-python@v1` instead of v5 that I recently merged in. Please rebase and let's see if the latest CI works.

still rebasing, will push up soon

legal-api/src/legal_api/models/db.py
eason-pan-bc: removed

chenhongjing:
Leave a note here to consider an extreme scenario. Currently printed logs during app initialization is based on FF value. But DB versioning is not completely initialized until the first db operation. If the FF is switched after app initialization and before [the real db versioning initialization](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/db.py#L174-L175), there's no logging to detect it.

Try to initialize db versioning earlier, or ignore this scenario for now?

eason-pan-bc:
> Leave a note here to consider an extreme scenario. Currently printed logs during app initialization is based on FF value. But DB versioning is not completely initialized until the first db operation. If the FF is switched after app initialization and before [the real db versioning initialization](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/db.py#L174-L175), there's no logging to detect it.
> 
> Try to initialize db versioning earlier, or ignore this scenario for now?

I feel like it's extremely rare that people switched the FF but actually don't want to do anything and switch again. So, I lean towards ignore it for now. Any ideas? @argush3 @leodube-aot @AimeeGao

AimeeGao (approved):
I think it's good now. LGTM

argush3:
> Leave a note here to consider an extreme scenario. Currently printed logs during app initialization is based on FF value. But DB versioning is not completely initialized until the first db operation. If the FF is switched after app initialization and before [the real db versioning initialization](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/db.py#L174-L175), there's no logging to detect it.
> 
> Try to initialize db versioning earlier, or ignore this scenario for now?

I agree that using the FF flag value as a way of logging what db versioning(old or new) to be inaccurate.  

But given that we had already done a bunch of testing with the more detailed logging and things seemed to be switching between the versions correctly when the FF was toggled, I'm ok with logging like this.

argush3 (approved):

chenhongjing (approved):

leodube-aot (approved):

PR #3276, opened 2025-03-03
26146 - Allow Amalgamations with C Business as Primary or Holding

meawong:
*Issue #:* /bcgov/entity#26146

*Description of changes:*
image
image
image
image
image

### DIFF
--- a/legal-api/src/legal_api/services/authz.py
+++ b/legal-api/src/legal_api/services/authz.py
@@ -537,6 +537,11 @@ def is_allowed(business: Business,
         else:
             is_ignore_draft_blockers = True
 
+    # Special case: handling authorization for amalgamation application
+    # this check is to make sure that amalgamation application is not allowed/authorized with continue in corps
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/services/authz.py
```diff
@@ -537,11 +537,6 @@ def is_allowed(business: Business,
         else:
             is_ignore_draft_blockers = True
 
-    # Special case: handling authorization for amalgamation application
-    # this check is to make sure that amalgamation application is not allowed/authorized with continue in corps
-    if filing_type == 'amalgamationApplication' and legal_type in ['C', 'CBEN', 'CUL', 'CCC']:
-        return False
-
```
vysakh-menon-aot, line 544: Why is it removed? This check is still valid

meawong: won't this check prevent amalgamations with continued-in businesses?

vysakh-menon-aot: Right. This won't allow the new temp business to become a continued in business. It won't block continued in business to become TING

vysakh-menon-aot (approved):

PR #3279, opened 2025-03-04
26262 - change logging approach

eason-pan-bc:
*Issue #:* /bcgov/entity#26262

*Description of changes:*
change logging approach to use `logger.info` to see if we'll be able to see loggings, and `logger.error` when not able to determine versioning type.
Known issue: won't be able to see it in Emailer, since the LOG_Level has been set to ERROR

image

When switching db versioning:
image
image


chenhongjing (approved):

AimeeGao (approved):
LGTM ^ ^

PR #3294, opened 2025-03-10
26359 - Fail the flow if one of the migration tasks fails

chenhongjing:
*Issue #:* /bcgov/entity#26359

*Description of changes:*
- Fail the flow to indicate errors exist if
    - User migration fails
    - Collecting corp data fails
    - Corp migration fails
- Some tweaks related to user migration error to mark corps FAILED

Note that the failed flow is not equal to all corp migrations fail in this flow.

_Testing_

Migrate several corps, and one corp fails. The flow is marked as FAILED.
image

### DIFF
--- a/data-tool/flows/corps_tombstone_flow.py
+++ b/data-tool/flows/corps_tombstone_flow.py
@@ -12,6 +12,7 @@ from prefect import flow, task, serve
 from prefect.futures import wait
 from prefect.context import get_run_context
 from prefect.task_runners import ConcurrentTaskRunner
+from prefect.states import Failed
 from prefect_dask import DaskTaskRunner
 from sqlalchemy import Connection, text
 from sqlalchemy.engine import Engine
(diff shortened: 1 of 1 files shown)


data-tool/flows/corps_tombstone_flow.py
```diff
@@ -437,12 +440,20 @@ def tombstone_flow():
...
-            users_mapper = migrate_corp_users(colin_engine, lear_engine, corp_nums)
-
-            # TODO: skip the following migration or continue?
-            if users_mapper is None:
-                print(f'❗ Skip populating user info for corps in this round due to user migration error.')
-                users_mapper = {}
+            try:
+                users_mapper = migrate_corp_users(colin_engine, lear_engine, corp_nums)
+            except Exception as e:
+                # skip migration if there's user migration error
```
chenhongjing, line 446: Not sure if we should fail and stop the corp migration in this round if the user migration fails.

argush3: yeah, i think we should fail and stop the corp migration if user migration fails

argush3 (approved):

vysakh-menon-aot (approved):

PR #3297, opened 2025-03-11
26409 Tombstone pipeline - timestamp with timezone

vysakh-menon-aot:
*Issue #:* /bcgov/entity#26409

*Description of changes:*
image

### DIFF
--- a/data-tool/flows/tombstone/tombstone_utils.py
+++ b/data-tool/flows/tombstone/tombstone_utils.py
@@ -785,7 +785,7 @@ def build_filing_json_meta_data(raw_filing_type: str, filing_type: str, filing_s
 
     if filing_type == 'annualReport':
         meta_data['annualReport'] = {
-            'annualReportFilingYear': int(effective_date[:4]),
+            'annualReportFilingYear': int(data['f_period_end_dt_str'][:4]),
         }
     elif filing_type == 'dissolution':
         dissolution_date = effective_date[:10]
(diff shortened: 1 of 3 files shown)


argush3 (approved):

vysakh-menon-aot:
> I don't think it's the fix for the issue.

Good catch

data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -785,7 +785,7 @@ def build_filing_json_meta_data(raw_filing_type: str, filing_type: str, filing_s
 
     if filing_type == 'annualReport':
         meta_data['annualReport'] = {
-            'annualReportFilingYear': int(effective_date[:4]),
+            'annualReportFilingYear': int(data['f_period_end_dt_str'][:4]),
```
chenhongjing, line 788: I'll think it would be better to update effective date to use `period_end_dt` for ARs and it's a special case.

The following screenshot is an AR filed in a different year of the AR year in LEAR.
image

vysakh-menon-aot: I don't think that's required. Even in Lear we have different date in effective_date and annualReportDate

vysakh-menon-aot: This is how we sync to colin. https://github.com/bcgov/lear/blob/main/colin-api/src/colin_api/models/filing.py#L613

chenhongjing: I try to trace the code logic back to FE,
1. https://github.com/bcgov/business-filings-ui/blob/main/src/views/AnnualReport.vue#L1033
2. https://github.com/bcgov/business-filings-ui/blob/main/src/views/AnnualReport.vue#L545
3. https://github.com/bcgov/business-filings-ui/blob/main/src/stores/businessStore.ts#L214

chenhongjing (approved):

PR #3311, opened 2025-03-17
Fix 26487 - Fix Unit Test of Amalgamation Out

ArwenQin:
*Issue #:* /bcgov/entity#26487

*Description of changes:*
- Fix the failed unit test in test_business.py


ArwenQin:
> LGTM but how does this fix the unit test?

I believe so, as the failed test shows, only the order is different
https://github.com/bcgov/lear/actions/runs/13864685985/job/38801078927
image


And I ran get('/api/v2/businesses/allowable/BC/ACTIVE'  locally:
image
just need to move Amalgamation Out before Annual Report

PR #3344, opened 2025-03-25
22143 - Add auth check for legal-api business endpoint v2

patrickpeinanw:
*Issue #:* /bcgov/entity#22143

*Description of changes:*
Add auth check for business endpoint v2. Public account with new 'view' role should not access full business information. (the slim version data is still available for public)

### DIFF
--- a/legal-api/src/legal_api/config.py
+++ b/legal-api/src/legal_api/config.py
@@ -185,9 +185,6 @@ class _Config():  # pylint: disable=too-few-public-methods
     # Transparency Register
     TR_START_DATE = os.getenv('TR_START_DATE', '').strip()  # i.e. '2025-02-01'
 
-    # Flag to control if the business endpoint v2 is open to the public (no auth check)
-    BUSINESS_V2_PUBLIC = False
-
     TESTING = False
     DEBUG = False
 
--- a/legal-api/src/legal_api/config.py
+++ b/legal-api/src/legal_api/config.py
@@ -185,6 +185,9 @@ class _Config():  # pylint: disable=too-few-public-methods
     # Transparency Register
     TR_START_DATE = os.getenv('TR_START_DATE', '').strip()  # i.e. '2025-02-01'
 
+    # Flag to control if the business endpoint v2 is open to the public (no auth check)
+    BUSINESS_V2_PUBLIC = False
+
     TESTING = False
     DEBUG = False
 
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -33,9 +33,9 @@ from legal_api.services import (  # noqa: I001;
     AccountService,
     RegistrationBootstrapService,
     check_warnings,
-    authorized
+    flags,
 )  # noqa: I001;
-from legal_api.services.authz import get_allowable_actions, get_allowed, get_could_files
+from legal_api.services.authz import authorized, get_allowable_actions, get_allowed, get_could_files
 from legal_api.utils.auth import jwt
 
 from .bp import bp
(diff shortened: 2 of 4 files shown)


kialj876:
> @kialj876 is this alright? I think you ran into issues previously and commented out the auth check on GET business. Maybe search is good with this change now?
> 
> If we roll this change in, we'll need to be careful. Not sure anything outside of search relies on this.

Yeah Search has been updated to use the 'slim' call everywhere which returns public data only and still returns without view permissions on the business. Last time Search was the only one that had the issue I think. @patrickpeinanw put it behind a config variable, but we could have it behind a FF I think if you prefer? If it breaks something else then we'll need to flip it back. We could also add some extra logging there so that its easier to monitor

legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -67,6 +63,14 @@ def get_businesses(identifier: str):
 
         return jsonify(business=business_json)
 
+    auth_check_required = not current_app.config.get('BUSINESS_V2_PUBLIC')
+
+    # check authorization -- need to implement workaround for business search users before using this #22143
+    if auth_check_required and not authorized(identifier, jwt, action=['view']):
```
kialj876, line 69: Since it'll be hard to verify all the integration pieces that use this, can you add a warning log here that spits out the identifier, username, account id, and 'app-name' from the request headers if it exists?

kialj876: Also you can remove the part of the comment referencing the ticket and business search

kialj876:
> > @kialj876 is this alright? I think you ran into issues previously and commented out the auth check on GET business. Maybe search is good with this change now?
> > If we roll this change in, we'll need to be careful. Not sure anything outside of search relies on this.
> 
> Yeah Search has been updated to use the 'slim' call everywhere which returns public data only and still returns without view permissions on the business. Last time Search was the only one that had the issue I think. @patrickpeinanw put it behind a config variable, but we could have it behind a FF I think if you prefer? If it breaks something else then we'll need to flip it back. We could also add some extra logging there so that its easier to monitor

It's probably something we'll want to have switched on in the sandbox for a while before we flip it in prod too

argush3:
>we could have it behind a FF I think if you prefer? If it breaks something else then we'll need to flip it back.

@kialj876 @patrickpeinanw yeah i think putting it behind a FF would be ideal for now.

legal-api/src/legal_api/config.py
```diff
@@ -185,6 +185,9 @@ class _Config():  # pylint: disable=too-few-public-methods
     # Transparency Register
     TR_START_DATE = os.getenv('TR_START_DATE', '').strip()  # i.e. '2025-02-01'
 
+    # Flag to control if the business endpoint v2 is open to the public (no auth check)
+    BUSINESS_V2_PUBLIC = False
```
argush3, line 189: can remove this now right?

legal-api/tests/unit/resources/v2/test_business.py
```diff
@@ -194,6 +194,36 @@ def test_get_business_info(app, session, client, jwt, requests_mock, test_name,
     assert registry_schemas.validate(rv.json, 'business')
 
 
+@pytest.mark.parametrize('test_name, slim_version', [
+    ('slim business request', True),
+    ('regular business request', False)
+])
+def test_get_business_with_unauthoized_role(app, session, client, jwt, requests_mock, test_name, slim_version):
```
argush3, line 201: think you might need to get your tests working with the feature flag?

legal-api/tests/unit/resources/v2/test_business.py
```diff
@@ -194,6 +196,52 @@ def test_get_business_info(app, session, client, jwt, requests_mock, test_name,
...
+    headers = create_header(jwt, [PUBLIC_USER], identifier)
+    requests_mock.get(f"{app.config.get('AUTH_SVC_URL')}/entities/{identifier}/authorizations", json={'roles': []})
+
+    if slim_version:
+        rv = client.get('/api/v2/businesses/' + identifier + '?slim=true', headers=headers)
+        assert rv.status_code == HTTPStatus.OK
+    else:
+        rv = client.get('/api/v2/businesses/' + identifier, headers=headers)
+        assert flags.is_on('enable-auth-v2-business') == auth_check_on
+        assert rv.status_code == (HTTPStatus.UNAUTHORIZED if auth_check_on else HTTPStatus.OK)
```
kialj876, line 242: Nice test

kialj876 (approved):
Nice work, looks good to me

argush3 (approved):

PR #3345, opened 2025-03-26
26303c lookup data tests

thorwolpert:
*Issue #:* /bcgov/entity#26303

*Description of changes:*
part III
- added all lookup data
- updated tests
- for DateTime columns set default=func.now() and onupdate=func.now()


python/common/business-registry-model/src/business_model/models/batch.py
```diff
@@ -46,7 +48,7 @@ class BatchStatus(BaseEnum):
     batch_type = db.Column('batch_type', db.Enum(BatchType), nullable=False)
     status = db.Column('status', db.Enum(BatchStatus), nullable=False)
     size = db.Column('size', db.Integer, nullable=True)
-    start_date = db.Column('start_date', db.DateTime(timezone=True), default=datetime.utcnow)
+    start_date = db.Column('start_date', db.DateTime(timezone=True),  default=func.now())
```
kialj876, line 51: extra space

python/common/business-registry-model/src/business_model/models/db.py
```diff
@@ -43,7 +43,7 @@ class Transaction(db.Model):
             )
     remote_addr = db.Column(db.String(50), nullable=True)
     issued_at = db.Column(db.DateTime,
-                          default=func.utc_timestamp(),
+                           default=func.now(),
```
kialj876, line 46: extra space

python/common/business-registry-model/src/business_model/models/configuration.py
```diff
@@ -84,7 +87,11 @@ def find_by_name(cls, config_name: str) -> Configuration:
         """Return the configuration matching the name."""
         configuration = None
         if config_name:
-            configuration = cls.query.filter_by(name=config_name).one_or_none()
+            try:
+                configuration = cls.query.filter_by(name=config_name).one_or_none()
+            except Exception as exc:
+                print(exc)
```
kialj876, line 93: Probably want this to be the app logger?

python/common/business-registry-model/tests/models/test_configuration.py
```diff
@@ -20,25 +20,26 @@
...
 
-def test_valid_configuration_save(session):
-    """Assert that a valid configuration can be saved."""
-    configuration = Configuration(
-        name='NEW_TEST_CONFIGURATION',
-        val='100',
-    )
-    configuration.save()
-    assert configuration.id
+# TODO: Do we need to create configs on the fly?
```
kialj876, line 23: Yeah it seems these are only populated by migrations so probably not, but it does look like these would work if you set 'name' to a valid name enum value? I think since you updated the type from string to enum it fails because the enum doesn't contain 'NEW_TEST_CONFIGURATION'

kialj876 (approved):
Looks good to me, I just noted some super minor cleanup items

python/common/business-registry-model/tests/models/test_configuration.py
thorwolpert: yep, hence my question, I couldn't see a place that didn't use the Enum

PR #3346, opened 2025-03-26
26697 Resolve sql_versioning issues after sqlalchemy upgrade

leodube-aot:
*Issue #:* /bcgov/entity#26607

*Description of changes:*
- Fix deprecation warnings.
- Split up `get_or_create_version_class` so that creating a version class happens explicitly before mapper configuration.

### DIFF
--- a/python/common/sql-versioning-alt/src/sql_versioning/versioning.py
+++ b/python/common/sql-versioning-alt/src/sql_versioning/versioning.py
@@ -18,9 +18,8 @@ from sqlalchemy import (BigInteger, Column, DateTime, Integer, SmallInteger,
                         String, and_, event, func, insert, inspect, select,
                         update)
 # from sqlalchemy.ext.declarative import declarative_base, declared_attr
-from sqlalchemy.ext.declarative import declared_attr
-from sqlalchemy.orm import declarative_base
-from sqlalchemy.orm import Session, mapper, relationships
+from sqlalchemy.orm import declarative_base, declared_attr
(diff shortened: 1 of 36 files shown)


python/common/sql-versioning-alt/src/sql_versioning/versioning.py
```diff
@@ -312,11 +311,12 @@ def __versioned_cls__(cls):
 
         :return: The versioned class.
         """
-        return cls.get_or_create_version_class()
+        return cls._version_cls
```
leodube-aot, line 314: The reason version tables weren't getting created is that this `declared_attr` decorator was not getting called prior to/during the mapper configuration. It seems that in sqlalchemy 2.0, declared_attrs are dynamic. This results in nothing being set for `_pending_version_classes`. I've split up `get_or_create_version_class` so that we create the version class before the mapper configuration.

leodube-aot, line 314: The reason version tables weren't getting created is that this `declared_attr` decorator was not getting called prior to/during the mapper configuration. It seems that in sqlalchemy 2.0, declared_attrs are dynamic. This results in nothing being set for `_pending_version_classes`. I've split up `get_or_create_version_class` so that we create the version class before the mapper configuration.

thorwolpert (commented):
Can we add a test to check this, will make maintenance up the versions easier. - thx

chenhongjing (approved):
LGTM

argush3 (approved):

PR #3350, opened 2025-03-27
chore: full ruff lint and fix

thorwolpert:
*Issue #:* /bcgov/entity/issues/26303

*Description of changes:*
- ruff applied
- manually fix additional errors

### DIFF
--- a/python/common/business-registry-model/pyproject.toml
+++ b/python/common/business-registry-model/pyproject.toml
@@ -33,10 +33,168 @@ freezegun = "^1.5.1"
 pytest = "^8.3.5"
 zimports = "^0.6.1"
 black = "^25.1.0"
+pytest-cov = "^6.0.0"
 
+[tool.ruff]
+# Exclude a variety of commonly ignored directories.
+exclude = [
+    ".vscode",
+    "dist",
+    "htmlcov",
+    "venv",
+    "migrations",
+    "devops",
+    ".history"
+]
 
 [tool.poetry.dependencies]
 sql-versioning = {develop = true}
 [build-system]
 requires = ["poetry-core>=2.0.0,<3.0.0"]
 build-backend = "poetry.core.masonry.api"
+
+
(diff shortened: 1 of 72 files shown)


python/common/business-registry-model/pyproject.toml
```diff
@@ -33,10 +33,168 @@ freezegun = "^1.5.1"
...
+    "TCH", # flake8-type-checking
+    "TID", # flake8-tidy-imports
+    "UP",  # pyupgrade
+]
+#ignore = [
+#    "B904", # use 'raise ... from err'
+#    "B905", # use explicit 'strict=' parameter with 'zip()'
+#    "N999"
+#]
+[tool.ruff.lint.per-file-ignores]
```
kialj876, line 113: I think its a good idea to have a comment for each rule or each group so that we can work towards unignoring some of these (particularly for the business_model grp) or maybe a tech debt ticket created to clean these up with the number commented here. Otherwise I think people might copy this config for new projects without realizing we actually prefer most of these as unignored

python/common/business-registry-model/pyproject.toml
```diff
@@ -33,10 +33,168 @@ freezegun = "^1.5.1"
...
+fixable = ["ALL"]
+unfixable = [
+    "ERA", # do not autoremove commented out code
+]
+
+# Allow unused variables when underscore-prefixed.
+#dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
+
+[tool.ruff.lint.isort]
+known-first-party = ["business_model"]
```
kialj876, line 151: Do we want to add 'sql_versioning_alt' and 'registry_schemas' ? It helps to know if its our own stuff when we get dependency issues (mainly for newer devs who won't be familiar with the common libraries)

kialj876 (approved):
Nice!

PR #3367, opened 2025-04-02
26950 Update tombstone pipeline to included ceased party data

argush3:
*Issue #:* /bcgov/entity#26950

*Description of changes:*
* Update parties query in tombstone pipeline to include ceased parties data

**Test notes:**

Have loaded some early adopters corps into dev with this update so we can compare the same corps in test:
'BC1507445','BC1507435','BC1292656','BC1249698','BC1196188'

In particular, I did some testing with BC1196188 as it tested the scenario where an AR was referencing a ceased director.  You can observe that the filing json for the AR is different in dev & test.  The corp in dev correctly files the AR with the ceased director.

image

image

filing json for AR
image

### DIFF
--- a/data-tool/flows/tombstone/tombstone_queries.py
+++ b/data-tool/flows/tombstone/tombstone_queries.py
@@ -458,7 +458,7 @@ def get_parties_and_addresses_query(corp_num):
     --    and e.corp_num = 'BC0006247' -- OFF, DIR, RCC
     --    and e.corp_num = 'BC0883637' -- INC, DIR
         and e.corp_num = '{corp_num}'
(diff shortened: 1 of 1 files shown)


chenhongjing (approved):

chenhongjing:
A heads-up that this is a temporary solution and doesn't work for some liquidation parties.

vysakh-menon-aot (approved):

argush3:
> A heads-up that this is a temporary solution and doesn't work for some liquidation parties.

Ok, can you cut a ticket and add to the tombstone epic?

PR #3379, opened 2025-04-07
Entity-BN gcp job

BrandonSharratt:
*Issue #:* /bcgov/entity#26990

*Description of changes:*
Added Github actions for the gcp entity bn job. 
Added GCP entity bn job converted it to a python job to run the notebooks.


.github/workflows/gcp-entity-bn-ci.yml
kialj876, line 1: Hey Brandon I think there's an issue with this file as its not running. Can you compare it to my PR? I think you may need to update the yaml under 'jobs'. We should see the ci run the linter, tests and build

kialj876 (commented):
Looking good, just one small thing for the CI. Also we probably don't actually want the CD flows in there because they will just fail until SRE builds the infrastructure and adds the config in 1pass for them. We can leave those to patricks team

.github/workflows/gcp-entity-bn-ci.yml
BrandonSharratt: fixed but note, this job doesn't have tests

kialj876: Looks like there's some issues with the build, let me know if you want a second pair of eyes

BrandonSharratt: Builds fixed (for both)

kialj876: awesome, thanks

kialj876 (approved):
Nice work 👍

PR #3384, opened 2025-04-08
27057 Legal API - update get_share_class_revision & get_share_series_revision to support tombstone

chenhongjing:
*Issue #:* /bcgov/entity#27057

*Description of changes:*
- The updates are to make sure NOA of filing can display correct information even after more subsequent filings are complete.
- Update `get_share_class_revision()`
- Update `get_share_series_revision()`
- Misc updates

_Notes & Observations_
- Alteration will completely remove the current share structure and populate new data if it alters share structure. Tombstone share structure will be replaced completely by the new ones after alteration.
- Alteration won't touch the current share structure and populate new data if it doesn't alter share structure. Tombstone share structure won't be replaced completely in this scenario.
- Correction will update the current share structure if necessary. Tombstone share structure will be updated partially.
- This work doesn't support the scenario that correction updates share structure info and we go back to download NOAs before this filings (e.g. CoD), as previous tombstone share structure will be lost due to missing versioned records.
- Transition & Conversion are not taken into considerations as they won't happen for corps in modernized system.

_Testing results for NOAs_

- All the testing businesses are in dev. "Before" means the NOAs downloaded from dev. "After" means the NOAs downloaded from local Legal API.
- All filings are processed with the old versioning, since the new versioning has cascade delete issue which block the validation.

1. CoD

NOA of CoD will only use non-versioned records
    
image

2. CoD + Alteration (no updates for share structure)

NOA of CoD will use non-versioned records
 
image

NOA of Alteration will use non-versioned records

 
image

3. CoD + Alteration (updates for share structure)

NOA of CoD will use deleted versioned records
image

NOA of Alteration will use inserted versioned records

image

### DIFF
--- a/legal-api/src/legal_api/models/filing.py
+++ b/legal-api/src/legal_api/models/filing.py
@@ -1252,6 +1252,16 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
         return filing_exists
 
     @staticmethod
+    def get_tombstone_filing(business_id: int):
+        """Return the tombstone filing for a given business if any."""
+        return (
+            db.session.query(Filing)
+            .filter(Filing.business_id == business_id)
+            .filter(Filing._status == Filing.Status.TOMBSTONE)
+            .one_or_none()
+        )
+
+    @staticmethod
     def is_filing_after_tombstone(filing_id: int, business_id: int) -> bool:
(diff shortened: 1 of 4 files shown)


chenhongjing:
Close as we choose [a new approach](https://github.com/bcgov/lear/pull/3394) to fix the issue.

PR #3385, opened 2025-04-08
26309 - business mailer

hfekete:
*Issue #:26309 * /bcgov/entity26309 ###

*Description of changes:*
- emailer to be used as endpoint for GCP messages
- adding conversion to poetry, package upgrades, ruff, and running tests. 


.github/workflows/business-emailer-ci.yml
```diff
@@ -0,0 +1,43 @@
+name: Business Emailer CI
```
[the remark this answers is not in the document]
hfekete: for me it looks like the only difference is in jobs (I have setup and if statement, and I have some info in workflow_dispatch) which part is wrong. Also, I copy pasted the one you mentioned from the Brandons PR and tried to update it to match emailer project

queue_services/business-emailer/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
```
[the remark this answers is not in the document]
hfekete: OK, will remove it, as am unsure on what to update it at if it needs updating. (I did try to remove stuff I removed from the code, like tracker stuff)

.github/workflows/business-emailer-ci.yml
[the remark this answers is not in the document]
hfekete: updated the way we discussed offline

queue_services/business-emailer/tests/flags.json
kialj876, line 1: You already have this defined at the root level so you shouldn't have this file (the one at root is meant to be used for the tests). You may want to consider changing the way your 'flags' works and follow what I did in my PR (it is a more current version of things and there's a fixture for tests you can use that'll allow you to update it as needed if you need to alter the values from the flags.json). Let me know if you want to chat about it -- would require small change in how you call it / initialize it in create_app

queue_services/business-emailer/src/business_emailer/config.py
```diff
@@ -0,0 +1,137 @@
...
+    LEGAL_API_URL = os.getenv("LEGAL_API_URL", None)
+    PAY_API_URL = os.getenv("PAY_API_URL", None)
+    AUTH_URL = os.getenv("AUTH_URL", None)
+    ACCOUNT_SVC_AUTH_URL = os.getenv("ACCOUNT_SVC_AUTH_URL", None)
+    # secrets
+    ACCOUNT_SVC_CLIENT_ID = os.getenv("ACCOUNT_SVC_CLIENT_ID", None)
+    ACCOUNT_SVC_CLIENT_SECRET = os.getenv("ACCOUNT_SVC_CLIENT_SECRET", None)
+    # variables
+    LEGISLATIVE_TIMEZONE = os.getenv("LEGISLATIVE_TIMEZONE", "America/Vancouver")
+    TEMPLATE_PATH = os.getenv("TEMPLATE_PATH", None)
```
kialj876, line 82: This is where you set the default for TEMPLATE_PATH and it should fix all the current test errors in the CI

queue_services/business-emailer/pyproject.toml
```diff
@@ -0,0 +1,208 @@
...
+[tool.ruff.lint.flake8-quotes]
+docstring-quotes = "double"
+
+[tool.ruff.lint.extend-per-file-ignores]
+"**/__init__.py" = ["F401"] # used for imports
+"src/business_emailer/services/versioned_business_details.py" = [
+    "I001", # ignoring 'unordered' imports
+    "PLR2004", # ignoring magic value comparison in this file
+    "RUF100"
+] # ignore checking unused noqa directives
```
kialj876, line 203: should those be ignored? They can be deleted?

queue_services/business-emailer/src/business_emailer/config.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+
+    # used to identify versioning flag
+    SERVICE_NAME = "emailer"
+    PROJECT_ROOT = os.path.abspath(os.path.dirname(__file__))
+
+    MSG_RETRY_NUM = int(os.getenv("MSG_RETRY_NUM", "5"))
+
+    SENTRY_DSN = os.getenv("SENTRY_DSN") or ""
+    SENTRY_DSN = "" if SENTRY_DSN.lower() == "null" else SENTRY_DSN
```
kialj876, line 66: we're taking out sentry so this should be removed

kialj876 (changes requested):
Looking pretty good, just a few more small things

kialj876:
Looking good to me, but we should replace all the '.utcnow()' to '.now(UTC)' and fix the other depreciation warnings. Let's meet when you're online on that test I have some ideas

kialj876 (approved):
Nice work!

PR #3390, opened 2025-04-09
future-effective-filings

BrandonSharratt:
*Issue #:* /bcgov/entity#26718

*Description of changes:*
added GCP and python 3.13 version of future-effective-filing job
Added CI Action for above job


gcp-jobs/future-effective-filings/README.md
```diff
@@ -0,0 +1,49 @@
...
+```shell
+poetry config virtualenvs.in-project true
+```
+```shell
+poetry install
+```
+
+You can issue any command in the current environment, via poetry's shell
+```shell
+poetry shell
```
kialj876, line 39: Oh so this is no longer the default for poetry. I've been doing `eval $(poetry env activate)` which is what it directed me to do after upgrading to >2. This still works, but you have to install it as an extra action I think?

gcp-jobs/future-effective-filings/src/future_effective_filings/file_future_effective.py
```diff
@@ -0,0 +1,116 @@
...
+# this will load all the envars from a .env file located in the project root
+load_dotenv(find_dotenv())
+run_mode = os.getenv("FLASK_ENV", "production")
+
+def create_app(run_mode=run_mode):
+    """Return a configured Flask App using the Factory method."""
+    app = Flask(__name__)
+    app.config.from_object(config.CONFIGURATION[run_mode])
+    gcp_queue.init_app(app)
+    StructuredLogging().init_app(app)
```
kialj876, line 45: Can you initialize this the same as me here and use current_app for logging instead of a new logger in each file? https://github.com/bcgov/lear/pull/3383/files#diff-77df499b905d0664774e9fc54014094843a730ecc664484eebf47a5569a9f158

We're planning changes to the structured_logging that'll add more context to our logs and expect this type of initialization

gcp-jobs/future-effective-filings/README.md
BrandonSharratt: Yea I haven't been using that either I've just been doing $(poetry env activate) but didn't update any of the readmes which I stole from somewhere else, I'll update it

gcp-jobs/future-effective-filings/src/future_effective_filings/file_future_effective.py
BrandonSharratt: Already made that change yesterday so that one is ready to go

gcp-jobs/future-effective-filings/README.md
```diff
@@ -0,0 +1,49 @@
+# Future Effective Filings Report
+
+Generate notebook report
```
hfekete, line 3: should be notebook renamed to future-effective-filling, like the entity-bn one above ?

BrandonSharratt: Changed

PR #3391, opened 2025-04-09
feat(api): return officers when added to the request QueryString

thorwolpert:
*Issue #:* /bcgov/entity#27088

*Description of changes:*
* added officers back in, but needs to be the requested role

### DIFF
--- a/legal-api/src/legal_api/models/party_role.py
+++ b/legal-api/src/legal_api/models/party_role.py
@@ -128,17 +128,18 @@ class PartyRole(db.Model, Versioned):
     @staticmethod
     def get_party_roles(business_id: int, end_date: datetime = None, role: str = None) -> list:
         """Return the parties that match the filter conditions."""
-        # TODO: remove filter that excludes officer when we have plans to deal with it
         party_roles = db.session.query(PartyRole). \
-            filter(PartyRole.business_id == business_id). \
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/models/party_role.py
```diff
@@ -128,18 +128,19 @@ def get_active_directors(business_id: int, end_date: datetime) -> list:
...
         party_roles = db.session.query(PartyRole). \
-            filter(PartyRole.business_id == business_id). \
-            filter(PartyRole.role != PartyRole.RoleTypes.OFFICER.value)
+            filter(PartyRole.business_id == business_id)
+
+        # TODO: rework the Officers section; change-on-change to still bring back officers
+        if not role or role != PartyRole.RoleTypes.OFFICER.value:
+            party_roles = party_roles.filter(PartyRole.role != PartyRole.RoleTypes.OFFICER.value)
+        if role is not None:
+            party_roles = party_roles.filter(PartyRole.role == role.lower())
```
argush3, line 138: Think there's an issue with this logic.  If you pass in `role` = "director", it will return all non officer roles?

thorwolpert: yes, that's what it used to do.
I just changed it, from the last change that totally removed the ability to get officers.

argush3 (approved):

PR #3399, opened 2025-04-11
26689 email reminder gcp

kialj876:
*Issue #:* /bcgov/entity#26689

*Description of changes:*
NOTE: easier to review if you look at each commit as the first one copies the existing code over and the next one applies the upgrade changes, then tests, then linting


hfekete (approved):
Nice, and thanks for all the examples here that we took in our PRs :D

PR #3401, opened 2025-04-11
27078 Versioning - Cascade delete relationship orphans

leodube-aot:
*Issue #:* /bcgov/entity#27078

*Description of changes:*
- Fix relationships orphans not getting deleted when cascade="delete, delete-orphan" is set

*Analysis:*
- Cascade delete on an orphaned record removes the non versioned records without issue. It properly creates the version record for the orphaned record. It does not however create the version record for cascade deleted records. 
```
# relationships: parent -> child -> grandchild
parent.remove(child)

# result:
# - child deleted
# - grandchild deleted
# - child_version delete record IS created
# - grandchild_version delete record NOT created
```

- This could be because SQLAlchemy emits a DELETE command for the child to grandchild relationship, but the grandchild object does not actually get added to session.delete or session.dirty. In order to resolve this we need to manually add the grandchild object to session.deleted before it is flushed.

_Testing:_

- Verified that all tables created correctly for
```
# relationships: parent -> child -> grandchild
parent.remove(child)
# result: all tables created correctly

# relationships: parent -> child -> grandchild -> great-grandchild
parent.remove(child)
# result: all tables created correctly

```

- Verified `business.share_classes.remove(share_class)` also creates to correct `share_series_version` records
Screenshot 2025-04-10 at 6 40 26 PM

### DIFF
--- a/python/common/sql-versioning/sql_versioning/versioning.py
+++ b/python/common/sql-versioning/sql_versioning/versioning.py
@@ -19,6 +19,7 @@ from sqlalchemy import (BigInteger, Column, DateTime, Integer, SmallInteger,
                         update)
 from sqlalchemy.ext.declarative import declarative_base, declared_attr
 from sqlalchemy.orm import Session, mapper, relationships
+from sqlalchemy.orm.dynamic import AppenderQuery
 
 from .relationship_builder import RelationshipBuilder
 
@@ -48,6 +49,9 @@ def _is_obj_modified(obj):
 
 def _should_relationship_delete_orphan(session, obj):
     """
+    Whether this object is orphaned from a relationship. This is primarily used  when the 
+    relationship structure is something like parent->child and parent.remove(child) is used.
+
     Checks if:
         1. This relationship is a many-to-one relationship
         2. If the opposite direction one-to-many relationship parent object has changes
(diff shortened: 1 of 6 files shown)


python/common/sql-versioning/sql_versioning/versioning.py
```diff
@@ -67,6 +71,41 @@ def _should_relationship_delete_orphan(session, obj):
...
+                r.direction.name == 'ONETOMANY' and
+                'delete' in r._cascade and
+                'delete-orphan' in r._cascade
+            ):
+                with session.no_autoflush:
+                    relationship_objects = getattr(obj, r.key)
+                    if isinstance(relationship_objects, AppenderQuery):
+                        relationship_objects = relationship_objects.all()
+                    for r_obj in relationship_objects:
+                        session.delete(r_obj)
```
leodube-aot, line 106: Couple things to note here:
- It's safe to manipulate session's state, including deleting object in [before_flush](https://docs.sqlalchemy.org/en/13/orm/session_events.html#before-flush).
- We need to use [no_autoflush context manager](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.no_autoflush) or else we get [an error](https://stackoverflow.com/questions/45972401/invalidrequesterror-session-is-already-flushing-has-been-occurred-in-before-f) for dynamic relationships.
- We need to check if the `relationship_objects` variable is an `AppenderQuery`, which is the case for `lazy="dynamic"` relationships. Non-dynamic relationships, like `share_class->share_series`, will return a list.

leodube-aot:
@thorwolpert FYI I've copied this fix into `sql-versioning-alt` (inc. the new unit test) and confirmed the tests pass.

vysakh-menon-aot (approved):

chenhongjing (approved):
Good learning, thanks Leo! 👍

PR #3416, opened 2025-04-17
26730 update colin filings

kialj876:
*Issue #:* /bcgov/entity#26730

*Description of changes:*


gcp-jobs/update-colin-filings/src/update_colin_filings/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+from structured_logging import StructuredLogging
+from update_colin_filings.config import DevelopmentConfig, ProductionConfig, UnitTestingConfig
+
+CONFIG_MAP = {
+    "development": DevelopmentConfig,
+    "testing": UnitTestingConfig,
+    "production": ProductionConfig
+}
+
+def create_app(environment: str = os.getenv("DEPLOYMENT_ENV", "production")):
```
BrandonSharratt, line 50: I don't love all of this being in init it breaks both convention and the standard we have in other jobs

gcp-jobs/update-colin-filings/src/update_colin_filings/config.py
```diff
@@ -0,0 +1,76 @@
...
+load_dotenv(find_dotenv())
+
+
+class _Config:
+    """Base class configuration."""
+
+    PROJECT_ROOT = os.path.abspath(os.path.dirname(__file__))
+    
+    JOB_TOTAL_LIMIT = int(os.getenv("JOB_TOTAL_LIMIT", "500"))
+    JOB_BATCH_LIMIT = int(os.getenv("JOB_BATCH_LIMIT", "50"))
```
BrandonSharratt, line 30: This and the one above should be in the .env.sample

BrandonSharratt (changes requested):
Looks good overall just a couple small changes

gcp-jobs/update-colin-filings/src/update_colin_filings/__init__.py
kialj876: This is the standard we have everywhere? It should be updated in the jobs that aren't like this I think rather than the other way around

gcp-jobs/update-colin-filings/src/update_colin_filings/config.py
kialj876: Updated

gcp-jobs/update-colin-filings/src/update_colin_filings/worker.py
```diff
@@ -0,0 +1,134 @@
...
+from http import HTTPStatus
+
+import requests
+from flask import current_app
+
+from business_account.AccountService import AccountService
+
+
+def get_filings(token, limit, offset):
+    """Get a filing with filing_id."""
```
hfekete, line 24: Is the comment in accordance to what method is doing ?

hfekete (approved):
LGM, 
had one question on doc comments

gcp-jobs/update-colin-filings/src/update_colin_filings/worker.py
kialj876: No its not, I'll update

PR #3426, opened 2025-04-24
27250 Tombstone Pipeline - Update to include DCR offices

leodube-aot:
*Issue #:* /bcgov/entity#27250

*Description of changes:*
After some conversation in the issue ticket, it was discovered that NOCDS (change respecting DCR) filings also update the dissolved company's records (DCR) office. This office has `office_typ_cd = DS`. This PR updates tombstone to bring over this office data.

- Updated tombstone to include DCR offices
- ~Create migration script to add `dcrOffice` to `office_types`~

_Local Testing_
Ran the tombstone pipeline on the following businesses locally. Verified DS offices are brought over.
```python
BC0041611
BC0252435
BC0500825
BC0654126
BC0684431
BC0702912
BC0878472
BC1199563
```

I will also reload these in dev once this PR is merged (since we need the migration script to update the db first)


data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -163,7 +175,7 @@ def format_parties_data(data: dict) -> list[dict]:
 
         formatted_party_roles = party['party_roles']
         for _, r in group.iterrows():
-            if (role_code := r['cp_party_typ_cd']) not in ['DIR', 'OFF', 'RCC', 'RCM']:
+            if (role_code := r['cp_party_typ_cd']) not in role_mapping.keys():
```
leodube-aot, line 178: Updated to check the dict keys instead of needing to maintain more code lists than necessary. Same with `office_type_mapping` above.

legal-api/migrations/versions/4a3cc8d777d3_add_dcroffice_to_office_types.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '4a3cc8d777d3'
+down_revision = 'edb19fafc835'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute('insert into office_types (identifier, description) values(\'dcrOffice\', \'Dissolved Company\'\'s Records Office\')')
```
leodube-aot, line 20: Doubling up the single quotation character will escape it in the sql:
```sql
insert into office_types (..., description) values (..., 'Dissolved Company''s ...`)
-- description output:  Dissolved Company's ...
```

data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -91,11 +91,23 @@ def format_offices_data(data: dict) -> list[dict]:
     offices_data = data['offices']
     formatted_offices = []
 
+    # Map office type codes to office type names
+    # TODO: support other office types
+    office_type_mapping = {
+        'RC': 'recordsOffice',
+        'RG': 'registeredOffice',
+        'DS': 'dcrOffice',
```
vysakh-menon-aot, line 99:
```suggestion
        'DS': 'custodialOffice',
```

ref:
https://github.com/bcgov/lear/blob/main/colin-api/src/colin_api/models/office.py#L34
https://github.com/bcgov/lear/blob/main/queue_services/entity-filer/src/entity_filer/filing_processors/dissolution.py#L75

legal-api/migrations/versions/4a3cc8d777d3_add_dcroffice_to_office_types.py
chenhongjing: Is it not custodial office that already exists in lear?

data-tool/flows/tombstone/tombstone_utils.py
leodube-aot: Ah that simplifies things! I'll update and remove the migration script.

legal-api/migrations/versions/4a3cc8d777d3_add_dcroffice_to_office_types.py
leodube-aot: Yup, see Vysakh's comment. I'm removing the model change and migration

argush3 (approved):

chenhongjing:
does anyone know if the address of dcrOffice will be used as the same as the address for custodian in modernized system(I mean, party address)? When I check RCC party in extract colin db, I find a small number of RCC having address, which is confusing.

leodube-aot:
> does anyone know if the address of dcrOffice will be used as the same as the address for custodian in modernized system(I mean, party address)? When I check RCC party in extract colin db, I find a small number of RCC having address, which is confusing.

Hmm I'm not sure. When I run the `get_parties_and_addresses_query` query for `NOCDS` filings in the test/prod extracts I don't find any `RCC` `corp_parties` with address info. What filing are you seeing this for?

chenhongjing:
> Hmm I'm not sure. When I run the `get_parties_and_addresses_query` query for `NOCDS` filings in the test/prod extracts I don't find any `RCC` `corp_parties` with address info. What filing are you seeing this for?

I found it via queries against corp_party table - filings are amalgamation(AMALV), dissolution(ADVD2) - it's weird though.

Here are the different behaviours for custodian and custodial office between colin and lear:

0. assumption: RCC is mapped to custodian, and DCR is mapped to custodial office.
1. colin asks to collect custodian and address, then stores them separately - RCC without address, DCR offfice using the address (ctst: BC0707737)
2. lear asks to collect custodian and addresss, then stores them together -  custodian with address, custodial office is inheriated from business existing address (dev: BC0887794)

I have no idea on the reason why they look different and think it's a design thing.

leodube-aot:
> Here are the different behaviours for custodian and custodial office between colin and lear:
> 
>     0. assumption: RCC is mapped to custodian, and DCR is mapped to custodial office.
> 
>     1. colin asks to collect custodian and address, then stores them separately - RCC without address, DCR offfice using the address (ctst: BC0707737)
> 
>     2. lear asks to collect custodian and addresss, then stores them together -  custodian with address, custodial office is inheriated from business existing address (dev: BC0887794)
> 
> 
> I have no idea on the reason why they look different and think it's a design thing.

I think it makes sense to store the custodian record in lear with the address. We generally expect `party` records to have at least a `mailing address`. I'll update the PR. https://github.com/leodube-aot/business-schemas/blob/main/src/registry_schemas/schemas/party.json#L122

leodube-aot:
Updated to save DCR office address with the RCC party.

Before:
Screenshot 2025-04-29 at 11 36 09 AM

After:
Screenshot 2025-04-29 at 11 44 16 AM

chenhongjing (approved):

PR #3432, opened 2025-04-25
27251 - Tombstone - implement appoint liquidator & data model updates

chenhongjing:
*Issue #:* /bcgov/entity#27251

*Description of changes:*
- Implement appoint liduidator in pipeline
- Add `liquidationRecordsOffice` into data model (confirmed with Dylan that it's different from `recordsOffice` and `custodialOffice`)

**Local Testing**
(Test it locally because dev db doesn't have new office type and will result in migration error)
- migration
image

- db
image

image


python/common/business-registry-model/src/business_model_migrations/versions/4eff12b18bef_add_business_lookup_data.py
```diff
@@ -35,7 +35,10 @@ def upgrade():
             },
             {
                 'identifier':'businessOffice', 'description':'Business Office'
-            }
+            },
+            {
+                'identifier':'liquidationRecordsOffice', 'description':'Liquidation Records Office'
```
chenhongjing, line 40: FYI @thorwolpert about the updates of common data model.

vysakh-menon-aot (approved):

argush3 (approved):

PR #3275, opened 2025-03-03
26189 Fix misc issue from tombstone prelim dry run & misc updates

chenhongjing:
*Issue #:* /bcgov/entity#26189

*Description of changes:*
- Fix the following issues
    - `psycopg2.errors.NumericValueOutOfRange) integer out of range`
    - `psycopg2.errors.StringDataRightTruncation) value too long for type character varying(10)`
    - `(sqlalchemy.exc.InvalidRequestError) A value is required for bind parameter ...`
    - `psycopg2.errors.NotNullViolation) null value in column "country" of relation "jurisdictions"`
- Misc updates to support tombstone corp loading in modernized system

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #3438, opened 2025-04-28
26614 - bring in share class and CI changes 

eason-pan-bc:
*Issue #:* /bcgov/entity#26614

*Description of changes:*
- As we updated the db for sandbox, we are experiencing some issues when the share_class max value is not None. Changes have already been made in the main branch. 
- Bringing changes related to share_classes, made in https://github.com/bcgov/lear/pull/3275 to the sandbox branch to make it compatible with updated db.

- Also, as Ubuntu 20 has been deprecated, brining in changes for CI workflows (legal-api and Filer) too


eason-pan-bc:
Unit tests failed, due to not having the db versions file in sandbox-branch, but we updated the code for the share_classes model.
Database in GCP sandbox has been updated, so shouldn't be a problem running there.

AimeeGao (commented):
Just a quick question👀 
What is our decision regarding this PR, do we want to pull in so many updates

eason-pan-bc:
if we think that we might have brought in too much from the main. I was thinking about bringing in Alembic scripts up to the commit related to the `share_class` changes. So that we don't need extra DC stuff
image

eason-pan-bc:
close this PR as we resolved the issue in the sandbox using another approach

PR #3439, opened 2025-04-28
27271 Tombstone - implement cease liquidator, ceased liquidator updates & misc

chenhongjing:
*Issue #:* /bcgov/entity#27271

*Description of changes:*
- Implement cease liquidator in pipeline
- Bring over ceased liquidator
    - It's a hacky solution for liquidator only.
    - There's only a small number of liquidator with `end_event_id` in COLIN.
    - This solution may not cover all the edge cases.
    - We need to sort every party type out in another work. e.g. https://github.com/bcgov/entity/issues/26973
- Fix for party group-by key

**⭐ Let me know if anyone has any questions, thanks!**

_Testing_

The following businesses are loaded in dev and affiliated to account `3446`
- Tesing data
```python
# cease liquidator filing
BC0246322
BC0442887

# liquidators got updated by filings, having both active and ceased liquidator
BC0318959
```
- DB
image
image

### DIFF
--- a/data-tool/flows/tombstone/tombstone_utils.py
+++ b/data-tool/flows/tombstone/tombstone_utils.py
@@ -134,7 +134,7 @@ def format_parties_data(data: dict) -> list[dict]:
     }
 
     df = pd.DataFrame(parties_data)
-    df['group_by_key'] = (df['cp_full_name'] + '_' + df['cp_party_typ_cd'] + '_' +
+    df['group_by_key'] = (df['partial_group_key'] + '_' + df['cp_party_typ_cd'] + '_' +
                           df['cp_cessation_dt_str'].fillna('active'))
 
     grouped_parties = df.groupby('group_by_key')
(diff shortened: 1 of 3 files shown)


data-tool/flows/tombstone/tombstone_utils.py
```diff
@@ -185,7 +185,10 @@ def format_parties_data(data: dict) -> list[dict]:
             party_role['appointment_date'] = appointment_date
 
             cessation_date = None
-            if cessation_date := r['cp_cessation_dt_str']:
+            if role_code == 'LIQ':
+                if end_event_date := r['cp_end_event_dt_str']:
+                    cessation_date = end_event_date
```
vysakh-menon-aot, line 190: Lear omit time in party cessation_date, would it be ok to trim the time from end_event_date? what does the output says in colin?

chenhongjing: It's paper-only in colin and I only find one scanned image that shows date without time. I assumed lear would do something like what it does for [cease receiver](https://github.com/bcgov/lear/blob/main/queue_services/entity-filer/src/entity_filer/filing_processors/cease_receiver.py#L36) in the end so I keep the time. But yes I'll update the code to trim it to make it consistent.

argush3 (approved):

vysakh-menon-aot (approved):

PR #3443, opened 2025-04-29
27346 updates to support sync filings from COLIN (BEN to BC businesses)

vysakh-menon-aot:
*Issue #:* /bcgov/entity#27346 

*Description of changes:*


colin-api/src/colin_api/models/filing.py
```diff
@@ -917,6 +919,9 @@ def get_filing(cls, filing: Filing, con=None, year: int = None) -> Dict:
                 else:
                     components = schema['properties'][filing.filing_type].get('properties').keys()
 
+            if filing_event_info['filing_type_code'] == 'CO_DI':
```
chenhongjing, line 922: What about including `AM_DI`? Same for line 974.

vysakh-menon-aot: These changes are limited to the filings filed after the alteration of BEN to BC on the 97 businesses. As of now AM_DI is not filed in any of those business

chenhongjing: k, good to me.

PR #3445, opened 2025-04-30
GCP required changes after testing in dev.

hfekete:
*Issue #:* /bcgov/entity###

*Description of changes:*
- some env variables updates
- removing async


queue_services/business-emailer/.env.sample
```diff
@@ -1,48 +1,48 @@
...
-DATABASE_TEST_USERNAME=
+DATABASE_TEST_USERNAME=.
 DATABASE_TEST_PASSWORD=
 DATABASE_TEST_NAME=
 DATABASE_TEST_HOST=
 DATABASE_TEST_PORT=
 
-
-#DEPLOYMENT_ENV=
+# gcp stuff
```
kialj876, line 46: it is just dev but I don't think we should be checking this in

kialj876 (commented):
Looks good, just one minor thing I noticed

queue_services/business-emailer/.env.sample
```diff
@@ -1,48 +1,48 @@
...
 
-LEGAL_API_URL=
-#LEGAL_API_URL=
+AUTH_API_URL=
+AUTH_API_VERSION=
 NOTIFY_API_URL=
-#NOTIFY_API_URL=
+NOTIFY_API_VERSION=
+BUSINESS_API_URL=
+BUSINESS_API_VERSION2=
```
kialj876, line 14: One last thing, LEGAL_API_URL and LEGAL_API_VERSION_2 are the values that exist in 1pass so if you're going to update it it should be to those (this is what it'll need to be altered to when we hook up the automated CD)

kialj876 (approved):
Looks good to me 👍

kialj876:
Oh needs some updates for the CI. Also can you update the job label in this CI yaml to be 'run-sre-ci' instead of 'furnishings-ci'

PR #3453, opened 2025-05-01
27199 - Use current registrar info for corrections

meawong:
*Issue #:* /bcgov/entity#27199

*Description of changes:*
- Use the current registrar's info instead of the original registrar's info for corrections

Before changes:
image

After:
image

### DIFF
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -341,8 +341,8 @@ class Report:  # pylint: disable=too-few-public-methods, too-many-lines
         if filing.get('correction'):
             original_filing = Filing.find_by_id(filing.get('correction').get('correctedFilingId'))
             original_registrar = {**RegistrarInfo.get_registrar_info(original_filing.effective_date)}
-            filing['registrarInfo'] = original_registrar
             current_registrar = {**RegistrarInfo.get_registrar_info(self._filing.effective_date)}
+            filing['registrarInfo'] = current_registrar
             if original_registrar['name'] != current_registrar['name']:
                 filing['currentRegistrarInfo'] = current_registrar
         elif filing.get('annualReport'):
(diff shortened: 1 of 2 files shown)


legal-api/src/legal_api/reports/report.py
```diff
@@ -341,8 +341,8 @@ def _set_registrar_info(self, filing):
         if filing.get('correction'):
             original_filing = Filing.find_by_id(filing.get('correction').get('correctedFilingId'))
             original_registrar = {**RegistrarInfo.get_registrar_info(original_filing.effective_date)}
-            filing['registrarInfo'] = original_registrar
             current_registrar = {**RegistrarInfo.get_registrar_info(self._filing.effective_date)}
+            filing['registrarInfo'] = current_registrar
```
vysakh-menon-aot, line 345: if `original_registrar` is not required then remove this `if` entirely

legal-api/src/legal_api/reports/report.py
```diff
@@ -339,12 +339,9 @@ def _set_completing_party(self, filing):
 
     def _set_registrar_info(self, filing):
         if filing.get('correction'):
```
vysakh-menon-aot, line 341: Please remove `if filing.get('correction'):`, `else` does the same thing

meawong: don't we want to keep `if filing.get('correction'):` to see if we need to set `filing['currentRegistrarInfo'] `?

vysakh-menon-aot: we are providing the same value for `registrarInfo` and `currentRegistrarInfo`, thats redundant. You can replace `currentRegistrarInfo` with `registrarInfo` in the templates

meawong: how about something like this?
image

PR #3449, opened 2025-05-01
27128 - Record of Amalgamation out Emailer

ArwenQin:
*Issue #:* bcgov/entity#27128

*Description of changes:*
1. Implement Record of Amalgamation out Emailer
2. Update email content for both CAO and AO

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #3418, opened 2025-04-22
27115 - Create Amalgamated Out Business Summary Output

ArwenQin:
*Issue #:* /bcgov/entity#27115

*Description of changes:*
- Filing Outputs are not needed for both CAO and Amalgamation Out, so removed them from document list and email attachments.
- Add Amalgamation Out Information Section in Business Summary

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #3386, opened 2025-04-08
27072 - Consent to Amalgamation Out Emailer

ArwenQin:
*Issue #:* /bcgov/entity#27072

*Description of changes:*
- send email
- display attributes
- attachments: Receipt, Filing Output (not designed yet, created an empty file), Letter of Consent (updated)
- Update Letter of Consent

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #3455, opened 2025-05-02
27128 - Add Amalgamation Out emailers to Business Emailer

ArwenQin:
*Issue #:* /bcgov/entity#27128

*Description of changes:*
- replicate changes in #3386 #3418 and #3449 
For Amalgamation Out and CAO emailers, email templates, unit tests

### DIFF
--- a/queue_services/business-emailer/tests/unit/__init__.py
+++ b/queue_services/business-emailer/tests/unit/__init__.py
@@ -91,8 +91,8 @@ def create_business(identifier, legal_type=None, legal_name=None, parties=None):
     return business
 
 
-def create_filing(token=None, filing_json=None,
-                  meta_data=None, business_id=None, filing_date=EPOCH_DATETIME, bootstrap_id: str = None):
+def create_filing(token=None, filing_json=None, business_id=None,
+                  filing_date=EPOCH_DATETIME, bootstrap_id: str = None, meta_data=None):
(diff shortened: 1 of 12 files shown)


vysakh-menon-aot (approved):

queue_services/business-emailer/tests/unit/__init__.py
```diff
@@ -89,7 +91,8 @@ def create_business(identifier, legal_type=None, legal_name=None, parties=None):
     return business
 
 
-def create_filing(token=None, filing_json=None, business_id=None, filing_date=EPOCH_DATETIME, bootstrap_id: str = None):
+def create_filing(token=None, filing_json=None,
+                  meta_data=None, business_id=None, filing_date=EPOCH_DATETIME, bootstrap_id: str = None):
```
hfekete, line 95: I would add new params at the end of the function call, unless there is specific reason to add it in that specific spot.

I quickly scanned through and think all the places have calls with named params, but to be on the safe side.

hfekete (approved):
I had one comment, that is not mandatory to update, but might be good.

queue_services/business-emailer/tests/unit/__init__.py
ArwenQin: Thank you, I just updated to move the new param to the end

PR #3462, opened 2025-05-05
GCP Migration - business-emailer tweaks

kialj876:
*Issue #:* https://github.com/bcgov/entity/issues/26590

*Description of changes:*
- updates for message integration discrepancies

### DIFF
--- a/queue_services/business-emailer/src/business_emailer/resources/business_emailer.py
+++ b/queue_services/business-emailer/src/business_emailer/resources/business_emailer.py
@@ -34,7 +34,6 @@
 #
 """This Module processes simple cloud event messages for the emailer.
 """
-import json
 from http import HTTPStatus
 
 import requests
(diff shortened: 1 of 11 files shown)


queue_services/business-emailer/src/business_emailer/resources/business_emailer.py
```diff
@@ -101,16 +101,13 @@ def worker():
...
-
-        process_email(email_msg)
+        process_email(ce)
         return {}, HTTPStatus.OK
 
     # ruff: noqa: PGH004
     except QueueException as err:  # noqa B902; pylint: disable=W0703; :
         # Catch Exception so that any error is still caught and the message is removed from the queue
-        current_app.logger.error("Queue Error: %s", json.dumps(email_msg), exc_info=True)
+        current_app.logger.error("Queue Error: %s", json.dumps(ce.data), exc_info=True)
```
hfekete, line 110: will this ce.data have info like type and what, it wont, right ? Maybe it would be good to write out entirety of ce in this error log.
Though we could potentially try to track it inside the logs.

hfekete (approved):
Looks OK to me.
I had only one suggestion for the log.

PR #3469, opened 2025-05-06
27606 - update effectiveDate description regarding CoD filing

eason-pan-bc:
*Issue #:* /bcgov/entity#27606

*Description of changes:*
update the description for `effectiveDate` regarding it as a required property for CoD filings

### DIFF
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -5516,8 +5516,8 @@ components:
         effectiveDate:
           type: string
           format: date-time
-          title: The date and time a filing should become valid and applied. Required for Change of Directors filing.
-          description: Date and time the filing is taking effect.
+          title: The date and time a filing should become valid and applied. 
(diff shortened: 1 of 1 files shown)


docs/business.yaml
```diff
@@ -5516,7 +5516,7 @@ components:
         effectiveDate:
           type: string
           format: date-time
-          title: The date and time a filing should become valid and applied.
+          title: The date and time a filing should become valid and applied. Required for Change of Directors filing.
```
AimeeGao, line 5519: I'm wondering if we should change the description part instead of the title, it seems we can't see the title on the API spec.
image

eason-pan-bc: good point, updated

AimeeGao (approved):
LGTM ^^

argush3 (approved):

PR #3482, opened 2025-05-08
27569 - Update Continuation Out Consent Letter

ArwenQin:
*Issue #:* /bcgov/entity#27569

*Description of changes:*
- Update the Continuation Out Consent Letter
Figma: https://www.figma.com/design/ueBofmytwhDeu82EBkDi6W/Consent-to-Continue-Out---Continuation-Out?node-id=2672-2&m=dev

### DIFF
--- a/legal-api/report-templates/letterOfConsent.html
+++ b/legal-api/report-templates/letterOfConsent.html
@@ -23,19 +23,21 @@
         &nbsp;{{ office.mailingAddress.postalCode }}
       </div>
       <div class="mt-7">Dear Applicant,</div>
-      <div class="mt-5">Re: <span class="underline">{{ business.legalName }} - {{ business.identifier }}</span></div>
+      <div class="mt-5">Re: <span class="bold">{{ business.legalName }} | {{ business.identifier }}</span></div>
       <div class="mt-5">
-        The above company was authorized on {{effective_date}} to continue out of British Columbia, 
-        under section 308 of the Business Corporations Act, to the jurisdiction of {{ jurisdiction }}.
+        {{ business.legalName }} was authorized on {{effective_date}} to continue out of British Columbia
+         to the jurisdiction of <span class="bold">{{ jurisdiction }}</span>.
+      </div>
+      <div class="mt-5 bold">
+        This authorization expires on {{ cco_expiry_date }}.
       </div>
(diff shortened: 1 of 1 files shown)


ArwenQin:
Examples:

Federal Canada
image


Other countries:
image


Province, Canada:
image

ArwenQin:
It seems the errors are caused somewhere else.

vysakh-menon-aot (approved):

legal-api/report-templates/letterOfConsent.html
```diff
@@ -23,19 +23,21 @@
...
+      <div class="mt-5 bold">
+        This authorization expires on {{ cco_expiry_date }}.
       </div>
       <div class="mt-5">
-        Section 311(1) states that "Promptly after the date on which a company is continued into another jurisdiction,
+        Section 311(1) states that “Promptly after the date on which a company is continued into another jurisdiction,
         the continued corporation must file with the registrar a copy of any record issued to it by the other
-        jurisdiction to effect or confirm the continuation." Please forward a copy of this record in order to remove
-        the above company from the companies registered in British Columbia.
+        jurisdiction to effect or confirm the continuation.”
```
severinbeauvais, line 37: Do we use use the "start quotes" and "end quotes" anywhere else?

They aren't in the UI design. Did your IDE add them automatically?

ArwenQin: Strange! I copied them from the Figma Design and paste in VS code, then they become "start quotes" and "end quotes". I type in VS code, then there is no start quotes or end quotes. I think the issue is in copying from Figma. I will correct them now.

severinbeauvais (approved):

legal-api/report-templates/letterOfConsent.html
ArwenQin: Do you think this needs to be corrected too 
image

I searched for it, it's in many places in the code:
image

It has no impact on the actual output

severinbeauvais: That the difference between forward and backward apostrophe, right?

It would be nice to update, but there nothing stopping another developer from adding them back in.

Also, that's a lot results/files to test (though spot-testing might be OK).

In the end, I think you should leave those alone :)

PR #3488, opened 2025-05-12
27476 - Add gcp queue to legal api

hfekete:
*Issue #:27476* /bcgov/entity27476

*Description of changes:*
- adding gcp queu to legal api
- adding tests


legal-api/src/legal_api/resources/v2/admin/administrative_bn.py
```diff
@@ -47,24 +48,37 @@ def publish_entity_event(business: Business,
...
-            'datacontenttype': 'application/json',
-            'identifier': business.identifier,
-            'data': {
-                'header': {
-                    'request': request_name,
-                    'businessNumber': business_number
-                },
-                'business': {'identifier': business.identifier}
-            }
+        # payload = {
```
kialj876, line 51: can remove commented

legal-api/src/legal_api/resources/v2/admin/reviews.py
```diff
@@ -152,11 +153,21 @@ def save_review(review_id: int):
...
+        publish_to_queue(
+            data={'email': {'filingId': filing.id, 'type': filing.filing_type, 'option': filing.status}},
+            subject=current_app.config.get('NATS_EMAILER_SUBJECT'),
+            business=business,
+            event_type='unknown:fixme', # todo: fixme add correct event_type
+            message_id=None,
+            is_wrapped=False
+        )
+    else:
+        current_app.logger.error('Business not found for filing.id=%s; no email sent !', filing.id)
```
vysakh-menon-aot, line 170: We cannot check `if business is not None`. It will be `None` for Continuation In reviews (and we only implemented reviews for Continuation In)

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -1137,10 +1144,16 @@ def submit_filing_for_review(filing: Filing):
...
+        business = Business.find_by_internal_id(filing.business_id)
+
         # emailer notification
-        queue.publish_json(
-            {'email': {'filingId': filing.id, 'type': filing.filing_type, 'option': review.status}},
-            current_app.config.get('NATS_EMAILER_SUBJECT')
+        publish_to_queue(
+            data={'email': {'filingId': filing.id, 'type': filing.filing_type, 'option': review.status}},
+            subject=current_app.config.get('NATS_EMAILER_SUBJECT'),
+            identifier=business.identifier,
```
kialj876, line 1153:
```suggestion
            identifier=business.identifier if business else None,
```

legal-api/src/legal_api/resources/v2/admin/reviews.py
hfekete: Thanks for noticing this. I have added ability for the business to be None and some guardrails in the code for when it happens.

legal-api/test_tax_id_update.http
kialj876, line 1: can remove this file I think or ?

kialj876 (approved):
Nice work, looks good

legal-api/test_tax_id_update.http
hfekete: yes, thanks, removed.

vysakh-menon-aot (approved):
Can you fix the linting errors

hfekete:
> Can you fix the linting errors

What I understood from discussion with Kial there is a plan to move to RUF. And if we move to RUF changes we do here we might need to revert them.

Additional note why I did not update those in this PR is cause they are due to some previous code changes, and doing change here might distract from the logic/code changes in this PR. 

If it is needed that our team work on those fixes, lets bring this up with Melissa and create a new ticket for this. We can clean it up then.

Thanks !

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -587,7 +587,14 @@ def process_colin_filing(identifier: str, filing: Filing, business: Business) ->
...
                 payload = {'filing': {'id': filing.id}}
-                queue.publish_json(payload)
+                publish_to_queue(
+                    data=payload,
+                    subject=current_app.config.get('NATS_FILER_SUBJECT'),
+                    identifier=business.identifier if business else None,
+                    event_type=None,
+                    message_id=None,
+                    is_wrapped=False
+                )
```
argush3, line 597: `payload = {'filing': {'id': filing.id}}` will be a problem for business-filer right?  

This will work for OCP but GCP filer will expect a payload of `{"filingMessage": {"filingIdentifier": filing_id}}`?

hfekete: I am unaware of differences between payload differences (other then where business identifier is added when business is available for the filing). Especially around filer as it was still undergoing changes when this PR was done.

Is this confirmed that it should look like this (being different from OCP payload) when GCP is concerned.

@thorwolpert You are working on a filer, can you confirm there are differences?

Also, what should we use to handle this difference.
I would try to avoid changing publish method, so that means we would need to handle the message differences here.

On top of my mind are two possible solutions:
a) we add both data in here (and with my knowledge), I think there should be no problems, as both OCP and GCP will ignore data that does not fit their mapping.
b) we make this place also GCP/OCP aware and depending on the deployment we provide appropriate message.

Then when removing NATS from the code we will clean it up in this place too either way. (In that regard, maybe safer way would be 2nd option)

kialj876: I like 'a' where we add both, but it's up to you @argush3

argush3: 'a' works for me too

hfekete: Implemented 'a' option and rebased the code. let me know if this should work.
Thanks!

argush3 (approved):

PR #3493, opened 2025-05-15
27621 - Fix Error filing COA

ArwenQin:
*Issue #:* /bcgov/entity#27621

*Description of changes:*
- Fix the error by omitting the contactInfo when the mailing_address is none

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -923,24 +923,11 @@ class ListFilingResource():  # pylint: disable=too-many-public-methods
             if filing.filing_type == Filing.FILINGS['transparencyRegister']['name']:
                 corp_type = 'BTR'
 
-        if mailing_address is None:
-            mailing_address = Address()
-            mailing_address.city = "unknown"
-            mailing_address.postalCode = "unknown"
-            mailing_address.province = "unknown"
-            mailing_address.addressLine1 = "unknown"
-            mailing_address.country = "unknown"
-
         payload = {
             'businessInfo': {
                 'businessIdentifier': f'{business.identifier}',
                 'corpType': f'{corp_type}',
-                'businessName': f'{business.legal_name}',
-                'contactInfo': {'city': mailing_address.city,
-                                'postalCode': mailing_address.postal_code,
-                                'province': mailing_address.region,
-                                'addressLine1': mailing_address.street,
-                                'country': mailing_address.country}
+                'businessName': f'{business.legal_name}'
             },
             'filingInfo': {
                 'filingIdentifier': f'{filing.id}',
@@ -948,6 +935,15 @@ class ListFilingResource():  # pylint: disable=too-many-public-methods
             },
             'details': ListFilingResource.details_for_invoice(business.identifier, corp_type)
         }
+
+        if mailing_address:
+            payload['businessInfo']['contactInfo'] = {
+                'city': mailing_address.city,
+                'postalCode': mailing_address.postal_code,
+                'province': mailing_address.region,
+                'addressLine1': mailing_address.street,
+                'country': mailing_address.country
+            }
         folio_number = filing.json['filing']['header'].get('folioNumber', None)
         if folio_number:
             payload['filingInfo']['folioNumber'] = folio_number
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -923,6 +923,14 @@ class ListFilingResource():  # pylint: disable=too-many-public-methods
             if filing.filing_type == Filing.FILINGS['transparencyRegister']['name']:
                 corp_type = 'BTR'
 
+        if mailing_address is None:
+            mailing_address = Address()
+            mailing_address.city = "unknown"
+            mailing_address.postalCode = "unknown"
+            mailing_address.province = "unknown"
+            mailing_address.addressLine1 = "unknown"
+            mailing_address.country = "unknown"
+
         payload = {
             'businessInfo': {
                 'businessIdentifier': f'{business.identifier}',
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -936,6 +936,7 @@ class ListFilingResource():  # pylint: disable=too-many-public-methods
             'details': ListFilingResource.details_for_invoice(business.identifier, corp_type)
         }
 
+        # mailing_address may not be available for the migrated businesses from COLIN
         if mailing_address:
             payload['businessInfo']['contactInfo'] = {
                 'city': mailing_address.city,


ArwenQin:
The issue is in method create_invoice, when mailing_address is none.
It not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.)
Example: try filing an alteration for:
https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0883825
So I fix it by adding a default invoice address

ArwenQin:
The linting errors are not caused by this change.
When I try to make a test filing locally (local legal api, dev database), I saw this error, may I know how to fix it?
It's for all filings and all businesses, so I believe it's not caused by the empty address issue.
image

severinbeauvais:
> The issue is in method create_invoice, when mailing_address is none. It not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.) Example: try filing an alteration for: https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0883825 So I fix it by adding a default invoice address

Why is mailing address none? Is the filing JSON incomplete?

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -923,6 +923,14 @@ def create_invoice(business: Business,  # pylint: disable=too-many-locals,too-ma
             if filing.filing_type == Filing.FILINGS['transparencyRegister']['name']:
                 corp_type = 'BTR'
 
+        if mailing_address is None:
+            mailing_address = Address()
+            mailing_address.city = "unknown"
+            mailing_address.postalCode = "unknown"
+            mailing_address.province = "unknown"
+            mailing_address.addressLine1 = "unknown"
+            mailing_address.country = "unknown"
```
severinbeauvais, line 932: Just curious, but can the properties be included in the Address constructor?

vysakh-menon-aot, line 932: Does the pay-api accept payload without `contactInfo`? if not can you try an empty address? I don't think its ok to provide "unknown"
```
        if mailing_address is None:
            mailing_address = Address()
```

ArwenQin:
> > The issue is in method create_invoice, when mailing_address is none. It not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.) Example: try filing an alteration for: https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0883825 So I fix it by adding a default invoice address
> 
> Why is mailing address none? Is the filing JSON incomplete?

It tries to get business.mailing_address, which is none. Also, in filings other than COA, even the filing JSON doesn't have mailing address
image

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
ArwenQin: Just tried it, but I'm not able to verify it, as I still saw this error, may I know how to fix it?
image
 Though I didn't see other errors.

severinbeauvais:
> The issue is in method create_invoice, when mailing_address is none. It not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.) Example: try filing an alteration for: https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0883825 So I fix it by adding a default invoice address

I am not able to file an alteration for this business because the business address is missing. How did you test this?

ArwenQin:
> > The issue is in method create_invoice, when mailing_address is none. It not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.) Example: try filing an alteration for: https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0883825 So I fix it by adding a default invoice address
> 
> I am not able to file an alteration for this business because the business address is missing. How did you test this?

That's expected, it's the same issue as filing a COA, caused by the missing mailing_address. It will be fixed by this change.

severinbeauvais:
> The issue is in method create_invoice, when mailing_address is none. It not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.) Example: try filing an alteration for: https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0883825 So I fix it by adding a default invoice address

OK, so you're saying with filings that aren't expected to have an address (such as CCO), the API fails when there is no office address.

The ticket was about filing a COA. I just tried on for this business and it appears to contain all office addresses. Why did this fail to file originally, and is the problem fixed now?

image

ArwenQin:
I mean this issue not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.), since the issue is in create_invoice method.
Because it tries to get business.mailing_address, which is none. It doesn't get it from JSON
image

image

severinbeauvais:
> > > The issue is in method create_invoice, when mailing_address is none. It not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.) Example: try filing an alteration for: https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0883825 So I fix it by adding a default invoice address
> > 
> > 
> > I am not able to file an alteration for this business because the business address is missing. How did you test this?
> 
> That's expected, it's the same issue as filing a COA, caused by the missing mailing_address. It will be fixed by this change.

The alteration doesn't start up at all due to the missing office addresses. @Mihai-QuickSilverDev Should we fix alterations, or is the fixed COA enough for this data migration issue?

Are you saying that another filing will populate the addresses (with empty or Unknown), and after that an alteration would work? I think that's still a problem because then you will assigned an invalid address to the business.

severinbeauvais:
> I mean this issue not only happens when file a COA, but in all filings with a fee (alteration, consent to contOut, etc.), since the issue is in create_invoice method. Because it tries to get business.mailing_address, which is none. It doesn't get it from JSON !

I see. So we're using the old address on the invoice? Does this matter? Maybe it doesn't matter because the address doesn't appear on the receipt. I dunno. @Mihai-QuickSilverDev , thoughts?

ArwenQin:
Strange, I could start an Alteration just now, but now it doesn't work anymore.
I can still start a Consent to ContOut.
I believe it's caused by 
image

I thought it's ignored by the front end. But now since we couldn't start an Alteration, the current change is not enough to fix it.

vysakh-menon-aot:
@severinbeauvais did we decide whether to show warning when data is missing on corps? So that the user will be forced to do a conversion filing? That should solve this issue

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
vysakh-menon-aot: @seeker25 can you answer this?

ArwenQin:
> > Strange, I could start an Alteration just now, but now it doesn't work anymore. I can still start a Consent to ContOut/ court order.
> 
> Different business maybe?
> 
> I just noticed that a CCO fetches the addresses, which fails (404), but the filing seems to work. CCO JSON doesn't need the addresses but the address fetch is part of Filings UI start-up. Anyway, this isn't an issue.
> 

For CCO, when we click file and pay, the filing still fails due to the create_invoice issue, same as COA.
That's the issue I try to fix in this ticket.
My current solution (adding a default unknown address) doesn't seem good.
I'm thinking about the following 2 options:
1. If we decide to force the user to do a COA first, we can modify create_invoice method to get the mailing_address from JSON for COA filing type
2. Otherwise, if pay api accept payload without 'contactInfo', we can omit it when mailing_address is none.

severinbeauvais:
I've create a Teams chat to discuss and try to decide this.

https://teams.microsoft.com/l/message/19:b8a5c45aeae44453887a5c14cf6051ed@thread.v2/1747331011874?context=%7B%22contextType%22%3A%22chat%22%7D

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
seeker25: pay-api can accept without contactInfo

seeker25: > Just tried it, but I'm not able to verify it, as I still saw this error, may I know how to fix it? image Though I didn't see other errors.


I think you're missing the secret for pay-api in there

seeker25: @ArwenQin @vysakh-menon-aot

vysakh-menon-aot: Thanks @seeker25

ArwenQin:
I just updated the current solution to omit contactInfo when mailing_address is none.

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -927,19 +927,23 @@ def create_invoice(business: Business,  # pylint: disable=too-many-locals,too-ma
...
         }
+
+        if mailing_address:
+            payload['businessInfo']['contactInfo'] = {
+                'city': mailing_address.city,
+                'postalCode': mailing_address.postal_code,
+                'province': mailing_address.region,
+                'addressLine1': mailing_address.street,
+                'country': mailing_address.country
+            }
```
severinbeauvais, line 946: Where does this end up -- are there any downstream impacts? Will this data be saved in the db? Will the Filer see this?

And, how do we test this?

vysakh-menon-aot: This change won't impact anywhere else. This payload is specific to pay-api call.

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -927,19 +927,23 @@ def create_invoice(business: Business,  # pylint: disable=too-many-locals,too-ma
...
+                'businessName': f'{business.legal_name}'
             },
             'filingInfo': {
                 'filingIdentifier': f'{filing.id}',
                 'filingTypes': filing_types
             },
             'details': ListFilingResource.details_for_invoice(business.identifier, corp_type)
         }
+
+        if mailing_address:
```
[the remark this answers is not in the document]
ArwenQin: Sure, just added

severinbeauvais (approved):
LGTM and please make sure you file a COA on your test business.

https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0883825?accountid=3040

vysakh-menon-aot (approved):

PR #3505, opened 2025-05-22
Business-pay: config tweaks

kialj876:
*Issue #:* https://github.com/bcgov/entity/issues/26590

*Description of changes:*


queue_services/business-pay/devops/gcp/clouddeploy.yaml
```diff
@@ -26,7 +26,7 @@ serialPipeline:
       verify: false
    deployParameters:
    - values:
-      deploy-env: "development"
+      deploy-env: "production"
```
kialj876, line 29: NOTE: if this is set to anything else then the 'flags' does not get initialized. I could alter that logic instead if it is preferred.

This sets the app to run off the 'ProdConfig' instead of the 'DevConfig'

argush3: think your change should be fine

PR #3517, opened 2025-05-27
28546 bumping version numbers

TVWerdal:
*Issue #:* /bcgov/entity/28546

Version bumps for release. 

### DIFF
--- a/queue_services/business-filer/pyproject.toml
+++ b/queue_services/business-filer/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "business-filer"
-version = "3.0.0"
+version = "3.0.1"
 description = "Business Registry Filer Service"
 authors = [
     {name = "thor",email = "1042854+thorwolpert@users.noreply.github.com"}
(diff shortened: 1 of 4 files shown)


queue_services/business-filer/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "business-filer"
-version = "3.0.0"
+version = "3.0.1"
```
JazzarKarim, line 3: These two are in GCP and I think we're following a different path for them. We also probably don't want to deploy right now any GCP projects so you might want to revert these.

@argush3 Do we want to update the version numbers for business filer and business pay?

argush3: let's leave business-pay and business-filer out for now.  business-pay could theoretically go out as it's shared between OCP and GCP.  but i think the business-pay change is still being tested out and it's not really a change that affects the functionality in OCP

JazzarKarim: Sounds good, thanks Argus!

PR #3234, opened 2025-02-19
25472 - Update Outputs for Expired Limited Restoration

meawong:
*Issue #:* /bcgov/entity#25472

*Description of changes:*
- Add putBackOff to stateFilings
- Update Company Status
- Change expiry time on existing Limited Restoration entry and Limited Restoration Application output to "11:59 pm Pacific time"

image
image
image
image

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #3536, opened 2025-06-04
28584 Remove LIQUIDATION state

chenhongjing:
*Issue #:* /bcgov/entity#28584

*Description of changes:*
- Remove `LIQUIDATION`
- Misc updates

_DB Changes (Local Testing)_
- run alembic script in legal-api
image

- run alembic script in business-registry-model
image


legal-api/report-templates/template-parts/common/businessDetails.html
```diff
@@ -44,8 +44,8 @@
                  <div class="pt-2">
                      <span class="capitalize-text">{{business.state}}</span>
                      {% if business.state in ('HISTORICAL', 'LIQUIDATION') and (stateFilings or amalgamationOut) %}
```
vysakh-menon-aot, line 46: This `if` is not having any data for `LIQUIDATION`. You can change it only for `HISTORICAL`

chenhongjing: I trace the history of changes for this file, and think the logic was completely changed at this point: https://github.com/bcgov/lear/pull/3234/files#diff-afd71bbb4b60c26899a3b451c618bd22e79d50442876b69e6931321ff0dcd2f0
Historical firms won't have state filing info as suffix in business details any more. I don't find any new requirements.
I think we should keep the `LIQUIDATION` while the logic needs to be refined again. Can anyone help confirm this?

legal-api/migrations/versions/0c3eb026efa0_alter_business_state_enum.py
```diff
@@ -0,0 +1,37 @@
...
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+    op.execute("ALTER TYPE state RENAME TO state_old")
+
+    state_new = postgresql.ENUM('ACTIVE', 'HISTORICAL', name='state')
+    state_new.create(bind, checkfirst=True)
+
+    op.execute("ALTER TABLE businesses ALTER COLUMN state TYPE state USING state::text::state")
+    op.execute("ALTER TABLE businesses_version ALTER COLUMN state TYPE state USING state::text::state")
+    if 'businesses_bak' in inspector.get_table_names():
```
chenhongjing, line 30: Can't remember the background of `businesses_bak` table (previous legal name change work?)...seems it was added manually only in dev. But it's causing some issues for alembic script if we don't alter this table and don't make this operation optional.

vysakh-menon-aot (approved):

argush3 (approved):

PR #3540, opened 2025-06-04
28593 Fix correction document for migrated businesses

vysakh-menon-aot:
*Issue #:* /bcgov/entity#28593

*Description of changes:*

### DIFF
--- a/data-tool/flows/tombstone/tombstone_utils.py
+++ b/data-tool/flows/tombstone/tombstone_utils.py
@@ -1032,7 +1032,7 @@ def build_epoch_filing(business_id: int) -> dict:
         'filing_date': now.isoformat(),
         'completion_date': now.isoformat(),
         'effective_date': now.isoformat(),
-        'status': 'TOMBSTONE'
+        'status': 'COMPLETED'
     }
     return filing
 
--- a/data-tool/flows/tombstone/tombstone_utils.py
+++ b/data-tool/flows/tombstone/tombstone_utils.py
@@ -1032,7 +1032,7 @@ def build_epoch_filing(business_id: int) -> dict:
         'filing_date': now.isoformat(),
         'completion_date': now.isoformat(),
         'effective_date': now.isoformat(),
-        'status': 'COMPLETED'
+        'status': 'TOMBSTONE'
     }
     return filing
 
--- a/legal-api/requirements-nats.txt
+++ b/legal-api/requirements-nats.txt
@@ -1,5 +1,4 @@
 nest_asyncio
 asyncio-nats-client==0.11.4
 asyncio-nats-streaming==0.4.0
-protobuf==3.15.8
 requests==2.25.1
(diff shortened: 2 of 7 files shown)


vysakh-menon-aot:
> Why not just update here without changing the `TOMBSTONE` status? https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/filing.py#L1241-L1245
> 
> ```python
> query = db.session.query(Filing). \
>     filter(Filing.business_id == filing.business_id). \
>     filter(Filing._status.in_([Filing.Status.COMPLETED.value, Filing.Status.TOMBSTONE.value]))
> ```
> 
> And think we had better keep this status and it will be helpful during backfill.

Yeah! That was my initial analysis. Then looking into the  filing type `lear_tombstone`, I don't see a need of a status called `TOMBSTONE`. For any identification related to tombstone (including backfill) we can identify using filing type `lear_tombstone`.

chenhongjing:
> > Why not just update here without changing the `TOMBSTONE` status? https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/filing.py#L1241-L1245
> > ```python
> > query = db.session.query(Filing). \
> >     filter(Filing.business_id == filing.business_id). \
> >     filter(Filing._status.in_([Filing.Status.COMPLETED.value, Filing.Status.TOMBSTONE.value]))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And think we had better keep this status and it will be helpful during backfill.
> 
> Yeah! That was my initial analysis. Then looking into the filing type `lear_tombstone`, I don't see a need of a status called `TOMBSTONE`. For any identification related to tombstone (including backfill) we can identify using filing type `lear_tombstone`.

I just think the small code changes like adding `TOMBSTONE` into query will be less risky than updating stuff in prod. We don't know other impact of this lear_tombstone filing when it's changed to `COMPLETED`. And `TOMBSTONE` was introduced at the beginning  so that we can change the status to `BACKFILL` later when we finish backfill pipeline for one business.

vysakh-menon-aot:
> I just think the small code changes like adding `TOMBSTONE` into query will be less risky than updating stuff in prod. We don't know other impact of this lear_tombstone filing when it's changed to `COMPLETED`. And `TOMBSTONE` was introduced at the beginning so that we can change the status to `BACKFILL` later when we finish backfill pipeline for one business.

I agree there is a risk but that should not stop us from implementing it. Yes we introduced `TOMBSTONE` status with the assumption to use it as a point in time for backfill, but we can do the same thing using `lear_tombstone` now

vysakh-menon-aot:
> What I mean about the risk is not only about the `TOMBSTONE` status in prod but more about making an abstract, not real filing `COMPLETED` like a real filing when some subsequent filing may reply on something we don't know.
Make sense

chenhongjing (approved):

PR #3542, opened 2025-06-05
28730 filer issue

kialj876:
*Issue #:* /bcgov/entity#28730

*Description of changes:*


queue_services/business-filer/pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [
 ]
 license = {text = "BSD-3-Clause"}
 readme = "README.md"
-requires-python = ">=3.9,<4"
+requires-python = ">=3.13,<4"
```
kialj876, line 10: NOTE: I think this is okay since nothing is using the filer as a dependency?

python/common/business-registry-model/src/business_model_migrations/alembic.ini
```diff
@@ -2,7 +2,7 @@
 
 [alembic]
 # template used to generate migration files
-# file_template = %%(rev)s_%%(slug)s
+file_template = %%(year)d%%(month).2d%%(day).2d_%%(hour).2d%%(minute).2d%%(second).2d_%%(rev)s_%%(slug)s
```
kialj876, line 5: NOTE: this automatically adds the date to the beginning of the migration file name. I've found this helpful in other APIs, but lmk what you think. It looks like the existing migrations are ordered with a number in front of them, but I think that was a manual edit to the names of the files?

kialj876:
NOTE: the unit tests in this one hang at the end for some reason so I cancelled the run, but you can see they pass up till the hanging point in the logs. I checked the previous PRs to this and it seems to be a long standing issue

vysakh-menon-aot (approved):

PR #3581, opened 2025-06-17
fix: GCP Namex integration issue

kialj876:
*Issue #:* /bcgov/entity#29020

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/config.py
+++ b/legal-api/src/legal_api/config.py
@@ -79,6 +79,10 @@ class _Config:  # pylint: disable=too-few-public-methods
 
     COLIN_URL = f"{os.getenv('COLIN_API_URL', '')}{os.getenv('COLIN_API_VERSION', '')}"
     LEGAL_API_BASE_URL = f'{BUSINESS_API_URL + BUSINESS_API_VERSION_2}/businesses'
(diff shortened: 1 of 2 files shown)


vysakh-menon-aot (approved):

vysakh-menon-aot:
Can you fix the unit test

legal-api/src/legal_api/config.py
```diff
@@ -79,6 +79,10 @@ class _Config:  # pylint: disable=too-few-public-methods
 
     COLIN_URL = f"{os.getenv('COLIN_API_URL', '')}{os.getenv('COLIN_API_VERSION', '')}"
     LEGAL_API_BASE_URL = f'{BUSINESS_API_URL + BUSINESS_API_VERSION_2}/businesses'
+    # Temporary while there is inconsistency between OCP / GCP versions of 1pass env
+    if NAMEX_API_VERSION and NAMEX_API_VERSION[-1] == '/':
+        # remove the slash
+        NAMEX_API_VERSION = NAMEX_API_VERSION[:-1]
```
severinbeauvais, line 85: All other "version" keys in 1Pass do not have a trailing slash. Why don't we fix the key instead?

severinbeauvais: OK, I get it. Resolving.

PR #3592, opened 2025-06-19
Use Document Service

BrandonSharratt:
*Issue #:* /bcgov/entity#28791 and /bcgov/entity#28790

*Description of changes:*
Add document service which uses the document api to store and fetch documents (pdfs)


legal-api/src/legal_api/reports/report.py
```diff
@@ -97,6 +106,8 @@ def _get_report(self):
         if response.status_code != HTTPStatus.OK:
             return jsonify(message=str(response.content)), response.status_code
 
+        self._document_service.create_document(self._business.identifier, self._filing.identifier, self._report_key, jwt.get_account_id(), response.content)
```
vysakh-menon-aot, line 109: Are we moving all documents to document service? I thought the plan is to move only the static documents (uploaded by users) which are in minio

legal-api/src/legal_api/reports/business_document.py
```diff
@@ -58,6 +63,7 @@ def get_pdf(self):
         response = requests.post(url=current_app.config.get('REPORT_SVC_URL'), headers=headers, data=json.dumps(data))
         if response.status_code != HTTPStatus.OK:
             return jsonify(message=str(response.content)), response.status_code
+        # self._document_service.create_document(self._business.identifier, self._document_key, self._report_date_time, jwt.get_account_id(), response.content)
```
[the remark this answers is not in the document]
BrandonSharratt: same as above removed

legal-api/src/legal_api/reports/report.py
BrandonSharratt: My understanding is we are moving all documents that get generated and have a filing id to the document service

vysakh-menon-aot: And I believe we have a separate ticket to replace minio with document service?

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -0,0 +1,101 @@
...
+            raise BusinessException('Document already exists', HTTPStatus.CONFLICT)
+        headers = {
+            'Content-Type': 'application/json',
+            'X-Api-Key': self.api_key,
+            'Account-Id': account_id
+        }
+        post_url = f'{self.url}/application-reports/{self.product_code}/{business_identifier}/{filing_identifier}/{report_type}'
+        response = requests.post(url=post_url, headers=headers, data=binary_or_url)
+        content = self.get_content(response)
+        if response.status_code != HTTPStatus.CREATED:
```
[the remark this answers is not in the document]
BrandonSharratt: yes my understanding is it's only accessible via the api key and the legal api already only returns/generates a document if the user calling it has sufficient permissions

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -0,0 +1,101 @@
...
+        else:
+          document = self.has_document(business_identifier, filing_identifier, report_type)
+          if document is False:
+            raise BusinessException('Document not found', HTTPStatus.NOT_FOUND)
+          get_url = f'{self.url}/application-reports/{self.product_code}/{document.file_key}'
+        
+        if get_url != '':
+          response = requests.get(url=get_url, headers=headers)
+          content = self.get_content(response)
+          if response.status_code != HTTPStatus.OK:
```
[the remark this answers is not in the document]
hfekete: Ah, yes, my mistake here, this is just used as check, and he is building the content, yes, that one is good.

BrandonSharratt:
Linting is failing on a file untouched by this PR

legal-api/src/legal_api/reports/report.py
```diff
@@ -114,7 +114,20 @@ def _get_report(self):
...
-        if account_id is not None:
+        create_document = account_id is not None
+        create_filing_types = [
+          'incorporationApplication',
+          'continuationIn',
+          'amalgamation',
+          'registration'
+        ]
+        if self._filing.filing_type in create_filing_types:
+            create_document = create_document and self._business.idenentifier
```
kialj876, line 125: error referencing the 'identifier' here so it will always be false. I think there should be some test cases added to secure this logic as it is really important and easy to miss the need for it

legal-api/src/legal_api/reports/document_service.py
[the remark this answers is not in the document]
BrandonSharratt: Added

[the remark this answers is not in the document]
flutistar: @BrandonSharratt I also created a similar file for the Corps integration. Would it be possible to move these functions to a common location, like under python/common? I’m using them in both the Legal API and Entity Filer (Business Filer). Thanks!

legal-api/src/legal_api/reports/report.py
```diff
@@ -97,6 +114,30 @@ def _get_report(self):
...
+        create_filing_types = [
+          'incorporationApplication',
+          'continuationIn',
+          'amalgamation',
+          'registration'
+        ]
+        if self._filing.filing_type in create_filing_types:
+            create_document = create_document and self._business is not None
+            create_document = create_document and self._business.identifier
+            create_document = create_document and self._business.tax_id
```
vysakh-menon-aot, line 127: `identifier` will be available when self._business object is not None
```suggestion
            create_document = create_document and self._business and self._business.tax_id
```

legal-api/src/legal_api/reports/report.py
BrandonSharratt: Test case added and above fixed

legal-api/src/legal_api/reports/document_service.py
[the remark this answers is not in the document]
BrandonSharratt: @flutistar I have finished the requested changes in this PR now, but am not intending to merge as we need the devops folks to cover the env variables, so while I'm happy to use the common one, I dont think it'll likely be possible yet as the legal api is on an older version of python so it will likely be tackled with the python upgrade and subsequent migration to the common libraries that's currently underway.

legal-api/.env.sample
```diff
@@ -89,3 +89,8 @@ BUSINESS_CRED_DEF_ID=
 
 # Digital Wallet Attestation credential
 WALLET_CRED_DEF_ID=
+
+DOCUMENT_API_URL=
+DOCUMENT_API_VERSION=/api/v1
+DOCUMENT_PRODUCT_CODE=LEGAL_API
```
kialj876, line 95: Can you set this to the actual entities product code? This is meant to match what is in auth

legal-api/src/legal_api/config.py
```diff
@@ -332,6 +341,10 @@ class TestConfig(_Config):  # pylint: disable=too-few-public-methods
     TRACTION_PUBLIC_SCHEMA_DID = os.getenv('TRACTION_PUBLIC_SCHEMA_DID', 'TRACTION_PUBLIC_SCHEMA_DID')
     TRACTION_PUBLIC_ISSUER_DID = os.getenv('TRACTION_PUBLIC_ISSUER_DID', 'TRACTION_PUBLIC_ISSUER_DID')
 
+    DOCUMENT_API_URL = os.getenv('DOCUMENT_API_URL', 'http://document-api.com')
```
kialj876, line 344: The test config shouldn't ever try to get the actual url. Can you update this to DOCUMENT_API_TEST_URL or just set it to http://document-api.com (I don't think we have a use case where we'd ever want it to actually call anything during testing)

kialj876:
> This also needs the new env variables added to the 'devops' folder for both the GCP and OCP deployments. As Brandon is away till the 2nd and I will be away until the 7th after today, I am thinking we should hold off on merging this in case the integration with DRS has issues while we're away

Still needs the variables added to the devops folder before we merge:
- vault: API
- section: doc-api
  - url: DOC_API_URL
  - version: DOC_API_VERSION
  - key: DOC_API_KEY

legal-api/src/legal_api/reports/document_service.py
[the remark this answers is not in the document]
kialj876: Yeah I think it should not pull the common one because there's other upgrades needed before the legal-api can use the common

vysakh-menon-aot:
Please fix the unit test

legal-api/src/legal_api/config.py
BrandonSharratt: hard set to the value that is above

legal-api/.env.sample
BrandonSharratt: done

kialj876 (approved):
Looks good to me. @argush3 @vysakh-menon-aot I'll leave the final review / merge to you guys

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -0,0 +1,164 @@
...
+        business_identifier: The business identifier.
+        filing_identifier: The filing identifier.
+        report_type: The report type.
+        account_id: The account id.
+        return: The document url (or binary).
+        """
+        token = AccountService.get_bearer_token()
+        headers = {
+            'X-ApiKey': self.api_key,
+            'X-Api-Key': self.api_key,
```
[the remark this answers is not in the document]
BrandonSharratt: Removed the extra one, It expects X-ApiKey, I had left it in as I had seen X-Api-Key previously as the standard but we appear to use X-ApiKey everywhere so I removed the one with the extra -

flutistar (approved):
Great work @BrandonSharratt ! 
🚀 🚀 🚀

PR #3602, opened 2025-06-23
28481 - Improve errors return from allowable actions check

ketaki-deodhar:
*Issue #:* /bcgov/entity#28481

*Description of changes:*

Replicate changes from main branch to update error message if user unable to file as a result of allowable action and update API specs

Reference PRs:
https://github.com/bcgov/lear/pull/3560/files
https://github.com/bcgov/lear/pull/3564/files

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -484,11 +484,13 @@ class ListFilingResource():  # pylint: disable=too-many-public-methods
         legal_type = business.legal_type if business else \
             filing_json['filing'][filing_type]['nameRequest'].get('legalType')
 
-        if not authorized(identifier, jwt, action=['edit']) or \
-                not is_allowed(business, state, filing_type, legal_type, jwt, filing_sub_type, filing):
-            return jsonify({'message':
-                            f'You are not authorized to submit a filing for {identifier}.'}), \
-                HTTPStatus.UNAUTHORIZED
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -484,11 +484,13 @@ def check_authorization(identifier, filing_json: dict,
...
-                not is_allowed(business, state, filing_type, legal_type, jwt, filing_sub_type, filing):
-            return jsonify({'message':
-                            f'You are not authorized to submit a filing for {identifier}.'}), \
-                HTTPStatus.UNAUTHORIZED
+        if not authorized(identifier, jwt, action=['edit']):
+            return jsonify({'message': f'You are not authorized to submit filings for {identifier}.'}), \
+                HTTPStatus.FORBIDDEN
+
+        if not is_allowed(business, state, filing_type, legal_type, jwt, filing_sub_type, filing):
+            return jsonify({'message': f'You are not allowed to submit this type of filing for {identifier}.'}), \
```
eason-pan-bc, line 492: just wondering if this error message will be good enough to help API users figure out the reason of not able to file things.

There are many cases were asked due to have a draft, have a pending filing, etc.

argush3: This should be fine as is.  For now, this is the compromise to detect that a filing was not allowed via the 403 + the difference in messaging.  Previously, it was 401 so it would be confusing if they had a valid token.

argush3 (approved):

PR #3624, opened 2025-07-04
Document Record Service Common Utils

flutistar:
*Issue #:* /bcgov/entity#28971

*Description of changes:*
This PR introduces utility functions for the Document Record Service (DRS) integration.


doug-lovett (approved):
Looks good so far. You may want to remove unused document classes. The max allowed files size could be an env variable.

python/common/document-record-service/src/document_record_service/document_service.py
```diff
@@ -0,0 +1,218 @@
...
+                json={"removed": True},
+                headers=self.headers,
+            ).json()
+            current_app.logger.info(f"Delete document from document record service {document_service_id}")
+            return response
+        except Exception as e:
+            current_app.logger.info(f"Error on deleting document {e}")
+            return {}
+
+    def get_document(self, request_info: RequestInfo) -> dict:
```
kialj876, line 126: I think this name is misleading as you are 'searching' for a document. I would expect the 'get_document' to retrieve the document by document id (a different drs endpoint which is also used by the legal api).

Also, will this document exist in the lear db documents table if it exists? Just wondering if it makes more sense to check the db and then retrieve the document via the id. You can also get the pdf directly this way by passing 'application/pdf' in the headers to drs

flutistar: Thank you, that makes sense — checking the lear DB first before making the API call is a good approach. However, for the search functionality, there are some DRS records that don’t have an associated document record (e.g. we create a DRS record for staff filings to capture the barcode number). So for the download function, I’m going to check the LEAR DB instead.

flutistar:
This PR is ready for review.

python/common/document-record-service/src/document_record_service/document_service.py
```diff
@@ -0,0 +1,235 @@
...
+        Args:
+            file: Uploaded file object (must have a `filename` attribute).
+            content_length (int): Size of the file in bytes.
+            document_type (str): Type of document to determine validation rules.
+
+        Returns:
+            list | None: List of validation errors if any, otherwise None.
+        """
+        msg = []
+        verify_paper_size = document_type in [DocumentTypes.CNTI.value]
```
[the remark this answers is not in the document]
flutistar: Oh, thank you sir!

BrandonSharratt (approved):

python/common/document-record-service/src/document_record_service/constants.py
```diff
@@ -0,0 +1,54 @@
+from enum import Enum
+
+
+class DocumentClasses(Enum):
+    """Render an Enum of the document service document classes."""
+
+    COOP = "COOP"
+    CORP = "CORP"
+    SOCIETY = "SOCIETY"
```
argush3, line 9: think we need an entry for firms(business type = SP/GP) here

python/common/document-record-service/src/document_record_service/constants.py
```diff
@@ -0,0 +1,54 @@
...
+    CORP = "CORP"
+    REGN = "REGN"
+    REGO = "REGO"
+    SYSR = "SYSR"
+
+
+# Map between filing and DRS document type
+# SYSR: alteration, appointReceiver, ceaseReceiver, changeOfDirectors,
+# incorporationApplication, restoration, noticeOfWithdrawal
+DOCUMENT_TYPES = {
```
argush3, line 34: So, some document types will apply to different business types.  I can see that `DocumentClasses` is used to represent the groups of business types(corps, coop and firms).

Some questions around that:
- there are no firm(SP/GP) related entries.  think these will need to be added
- how do you represent documents that will be used across multiple document classes?  for example, annual reports will apply coops, firms and corps.

python/common/document-record-service/src/document_record_service/document_service.py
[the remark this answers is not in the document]
flutistar: Fixed

python/common/document-record-service/src/document_record_service/constants.py
```diff
@@ -0,0 +1,63 @@
...
+
+# Map between legal type and DRS document class
+# For all other legal types not listed below, the document class defaults to CORP
+DOCUMENT_CLASSES = {
+    "CP": DocumentClasses.COOP.value,
+    "XCP": DocumentClasses.COOP.value,
+    "SP": DocumentClasses.FIRM.value,
+    "GP": DocumentClasses.FIRM.value,
+    "FM": DocumentClasses.FIRM.value,
+    "MF": DocumentClasses.FIRM.value,
```
argush3, line 59: I believe these 2 can be removed.  Only SP and GP are firms if the key is only intended to be legal types

python/common/document-record-service/src/document_record_service/constants.py
```diff
@@ -0,0 +1,63 @@
...
+    "consentAmalgamationOut": DocumentTypes.AMLO.value,
+    "consentContinuationOut": DocumentTypes.CNTO.value,
+    "continuationIn": DocumentTypes.CNTI.value,
+    "continuationOut": DocumentTypes.CNTO.value,
+    "conversion": DocumentTypes.CNVS.value,
+    "correction": DocumentTypes.CORR.value,
+    "courtOrder": DocumentTypes.CRT.value,
+    "registrarsNotation": DocumentTypes.REGN.value,
+    "registrarsOrder": DocumentTypes.REGO.value,
+    "systemIsTheRecord": DocumentTypes.SYSR.value,
```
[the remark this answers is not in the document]
flutistar: We have a document type called `System is the record`. For any other documents not listed, the document type will be SYSR.
For better clarity, I’ll send you the spreadsheet link via direct message.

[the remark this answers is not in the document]
argush3: As per our conversation, there will need to be some additional work done in subsequent PRs to ensure all the existing `DOCUMENT_TYPE` entries are accurate and any missing ones are added.

Things that should be addressed in subsequent PRs should include:
- are consent entries accurate?  two entries map to CNTO
- does staff continuation authorization entry need to be added
- spreadsheet has missing document type entries for IA and other filings.  if they are not SYSR, these entries will need to be added
- there are no entries in spreadsheet for firms.  i'm assuming there will be non SYSR entries.  if so they need to be added

vysakh-menon-aot (approved):

PR #3634, opened 2025-07-11
Hotfix for Search

Rajandeep98:
*Issue #:* /bcgov/entity#

*Description of changes:*
Hotfix for search endpoint
- updated early return if no filings being searched.
limit is always being set by default so would be always search filters, no identifiers then return early


legal-api/src/legal_api/services/search_service.py
```diff
@@ -195,6 +195,9 @@ def get_search_filtered_filings_results(business_json,
                                             identifiers=None,
                                             search_filters: AffiliationSearchDetails = None):
         """Return contact point from business json."""
+        if not search_filters or not identifiers:
```
seeker25: `if not identifiers:`

vysakh-menon-aot (approved):

PR #3670, opened 2025-07-28
27028 - Update Logic to Retrieve COOP Docs and Info from Correction

meawong:
*Issue #:* /bcgov/entity#27028 

*Description of changes:*
- Also check correction filings for any resolutions when retrieving COOP Docs and Info
- Refactored logic into helpers due to `too-many-locals` lint warning 

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_documents.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_documents.py
@@ -118,43 +118,46 @@ def _get_coop_documents_and_info(business):
     info['certifiedRules'], info['certifiedMemorandum'] = {}, {}
 
     sr_filings = Filing.get_filings_by_types(business.id, ['specialResolution'])
-    sr_rules_resolution = [sr for sr in sr_filings
-                           if sr.filing_json['filing'].get('alteration', {}).get('rulesInResolution') is True]
-    sr_memorandum_resolution = [sr for sr in sr_filings
-                                if sr.filing_json['filing'].get('alteration', {}).get('memorandumInResolution') is True]
-    if rules_document := Document.find_by_business_id_and_type(business.id, DocumentType.COOP_RULES.value):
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/services/business/validations/validator.py
```diff
@@ -48,7 +48,8 @@ def validate_document_request(document_type, business: Business):
         if (status := document_rules.get('status', None)) and business.state != status:
             errors.append({'error': babel('Specified document type is not valid for the current entity status.')})
 
-        if (good_standing := document_rules.get('goodStanding', None)) and not business.good_standing:
```
meawong, line 51: image

legal-api/src/legal_api/resources/v2/business/business_documents.py
```diff
@@ -118,43 +118,46 @@ def _get_coop_documents_and_info(business):
...
-        }
-    if sr_memorandum_resolution:
-        info['certifiedMemorandum']['includedInResolution'] = True
-        info['certifiedMemorandum']['includedInResolutionDate'] = sr_memorandum_resolution[0].filing_date.isoformat()
+    cr_filings = Filing.get_filings_by_types(business.id, ['correction'])
+
+    def filter_resolutions(filings, filing_type, key):
+        return [
+            f for f in filings
+            if f.filing_json['filing'].get(filing_type, {}).get(key) is True
```
vysakh-menon-aot, line 126: I don't think referring `filing_json` like this is right, we should be using `meta_data` instead. I can see for correction (https://github.com/bcgov/lear/blob/main/queue_services/business-filer/src/business_filer/filing_processors/filing_components/correction.py#L175) its available but not for alteration. It require changes in filer and update existing data in db. Since the previous implementation was same I am approving this PR. Please create a new ticket to update this

PR #3686, opened 2025-08-01
29901 - Add paymentDate to API Spec

meawong:
*Issue #:* /bcgov/entity#29901 

*Description of changes:*

### DIFF
--- a/docs/business.yaml
+++ b/docs/business.yaml
@@ -6304,6 +6304,10 @@ components:
         isFutureEffective:
           type: boolean
           description: Identify if filing was submitted with a future effective date (yes/no).
+        paymentDate:
+          type: string
+          minLength: 1
+          description: 'Date the payment was made for the filing.'
         paymentStatusCode:
           type: string
           minLength: 1


docs/business.yaml
```diff
@@ -6304,6 +6304,10 @@ components:
         isFutureEffective:
           type: boolean
           description: Identify if filing was submitted with a future effective date (yes/no).
+        paymentDate:
+          type: string
+          minLength: 1
+          description: 'Date the payment was made for the filing.'
```
vysakh-menon-aot, line 6310: add `format: date-time`. same as effective date above

docs/business.yaml
```diff
@@ -6304,6 +6304,11 @@ components:
         isFutureEffective:
           type: boolean
           description: Identify if filing was submitted with a future effective date (yes/no).
+        paymentDate:
+          type: string
+          minLength: 1
```
vysakh-menon-aot, line 6309: paymentDate can be null. @argush3 do we need to remove this?

argush3: maybe add a note indicating that filings from legacy filings won't have a value

PR #3715, opened 2025-08-19
29852 - Adding FF to Filings

Rajandeep98:
*Issue #:* /bcgov/entity#29852

*Description of changes:*
Added API FF to enable the filings

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/agm_location_change.py
+++ b/legal-api/src/legal_api/services/filings/validations/agm_location_change.py
@@ -19,6 +19,7 @@ from flask_babel import _ as babel  # noqa: N813, I004, I001; importing camelcas
 # noqa: I003
 from legal_api.errors import Error
 from legal_api.models import Business
+from legal_api.services import flags
 from legal_api.services.utils import get_int
 from legal_api.utils.legislation_datetime import LegislationDatetime
 # noqa: I003
@@ -29,6 +30,11 @@ def validate(business: Business, filing: Dict) -> Optional[Error]:
     if not business or not filing:
         return Error(HTTPStatus.BAD_REQUEST, [{'error': babel('A valid business and filing are required.')}])
 
+    if flags.is_on('supported-agm-location-change-entities'):
+        enabled_filings = flags.value('supported-agm-location-change-entities').split()
+        if not (business.legal_type in enabled_filings):
+            return Error(HTTPStatus.BAD_REQUEST, [{'error': babel('A valid legal type is required.')}])
+        
     msg = []
 
     agm_year_path: Final = '/filing/agmLocationChange/year'
--- a/legal-api/src/legal_api/services/filings/validations/amalgamation_out.py
+++ b/legal-api/src/legal_api/services/filings/validations/amalgamation_out.py
@@ -19,6 +19,7 @@ from flask_babel import _ as babel  # noqa: N813, I004, I001; importing camelcas
 # noqa: I003
 from legal_api.errors import Error
 from legal_api.models import Business, ConsentContinuationOut
+from legal_api.services import flags
 from legal_api.services.filings.validations.common_validations import (
     validate_court_order,
     validate_foreign_jurisdiction,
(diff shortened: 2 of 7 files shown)


legal-api/src/legal_api/services/filings/validations/agm_location_change.py
```diff
@@ -29,6 +30,12 @@ def validate(business: Business, filing: Dict) -> Optional[Error]:
     if not business or not filing:
         return Error(HTTPStatus.BAD_REQUEST, [{'error': babel('A valid business and filing are required.')}])
 
+    if flags.is_on('supported-agm-location-change-entities'):
```
severinbeauvais, line 33: What does `is_on()` tell you? That it's not empty?

As long as the FF is always an array, I'm thinking you don't need this line. Thoughts?

Rajandeep98: added check on nd off string should work unless empty

legal-api/src/legal_api/services/business/validations/validator.py
```diff
@@ -40,7 +41,10 @@ def validate_document_request(document_type, business: Business):
     """Validate the business document request."""
     errors = []
     # basic checks
-
+    enabled_filings = flags.value('enabled-business-summary-entities').split()
+    if enabled_filings and business.legal_type not in enabled_filings:
+        return Error(HTTPStatus.BAD_REQUEST,
+                     [{'error': babel(f'{business.legal_type} is not enabled business summary.')}])
```
severinbeauvais, line 47: Is this code only for Business Summary requests?

If not then this code will block all documents based on the business-summary FF, which is incorrect.

Rajandeep98: added for summary document

legal-api/src/legal_api/services/business/validations/validator.py
```diff
@@ -40,7 +41,10 @@ def validate_document_request(document_type, business: Business):
     """Validate the business document request."""
     errors = []
     # basic checks
-
+    enabled_filings = flags.value('enabled-business-summary-entities').split()
+    if enabled_filings and business.legal_type not in enabled_filings:
+        return Error(HTTPStatus.BAD_REQUEST,
```
severinbeauvais, line 46: Is HTTP 400 (bad request) the best choice? What do you think about HTTP 403 (forbidden) instead? Thoughts?

Rajandeep98: yes this would be better 403, updating

severinbeauvais (commented):
LGTM but what about "AGM Extension" filing?

severinbeauvais:
There is a file with the fallback FF values, that I don't see updated here.

vysakh-menon-aot (approved):

legal-api/flags.json
```diff
@@ -27,6 +27,15 @@
...
+    },
+    "enabled-business-summary-entities": "CP",
+    "supported-amalgamation-out-entities": "BC",
+    "supported-consent-amalgamation-out-entities": "BC",
+    "supported-put-back-on-entities": "BC",
+    "supported-continuation-in-entities": "BC",
+    "support-agm-extension-entities": "BC",
+    "supported-agm-location-change-entities": "BC",
+    "supported-continuation-out-entities": "BC",
+    "supported-consent-continuation-out-entities": "BC"
```
severinbeauvais, line 39: These fallback values will be used if LEAR can't get the FFs from LaunchDarkly.

I think most of these should be empty strings, since we haven't fully released these features, and especially not for BC Limited companies.

vysakh-menon-aot: These are not fallback values. It is for local only. Ref: https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/flags.py#L71

legal-api/flags.json
```diff
@@ -28,14 +28,14 @@
...
-    "supported-consent-continuation-out-entities": "BC"
+    "enabled-business-summary-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL",
+    "supported-amalgamation-out-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL",
+    "supported-consent-amalgamation-out-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL",
+    "supported-put-back-on-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL",
+    "supported-continuation-in-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL",
+    "support-agm-extension-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL",
+    "supported-agm-location-change-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL",
+    "supported-continuation-out-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL",
+    "supported-consent-continuation-out-entities": "CP BEN SP GP CBEN BC CC ULC C CCC CUL"
```
severinbeauvais, line 39: CP does not support all of these.

Please check in LaunchDarkly for whatever the "all" variation is.

severinbeauvais (approved):

PR #3722, opened 2025-08-22
30272 - Submitter roles data fix for migrated staff filings

AimeeGao:
*Issue #:* /bcgov/entity#30272

*Description of changes:*
- Added a notebook to fix migrated corps staff filings issue
   - Updated their `filings.submitter_roles` in LEAR to `staff`

Local Test:

Before:
image

After:
image


eason-pan-bc (approved):
LGTM 🛩️

data-tool/notebooks/fix_submitter_roles/fix_submitter_roles.ipynb
```diff
@@ -0,0 +1,294 @@
...
+    "for _, row in filings_results_df.iterrows():\n",
+    "    values_list.append(f\"({row['filing_id']}, '{row['role_typ_cd']}')\")\n",
+    "\n",
+    "values_str = ','.join(values_list)\n",
+    "\n",
+    "upsert_query = f\"\"\"\n",
+    "    INSERT INTO filings (id, submitter_roles)\n",
+    "    VALUES {values_str}\n",
+    "    ON CONFLICT (id)\n",
+    "    DO UPDATE SET submitter_roles = 'staff'\n",
```
vysakh-menon-aot, line 261: Any specific reason to use this query instead of a direct update query?

eason-pan-bc: We were thinking about if we found more filings from colin_extract db than the ones already in lear db. 
But since we changed to find filings directly from lear db, maybe we can change this back to direct update?

AimeeGao: Yeah, we thought we were pulling filings from migrated corps. Some of them might have submitter_roles as NULL or blank. With `INSERT ... ON CONFLICT`, it would insert if missing, or update if it already existed.

Maybe `update` would be simpler and clearer here. We can tweak a bit here.

argush3 (approved):

AimeeGao:
> Does this change need to be deployed to Test or Prod? Are any special steps required?

I’m testing this in Dev now. Thinking we could try moving it to Test? @argush3 @vysakh-menon-aot what do you think 👀

severinbeauvais:
I hear that `data-tool` doesn't need to be released, so I think you can ignore my question :)

AimeeGao:
> > Does this change need to be deployed to Test or Prod? Are any special steps required?
> 
> I’m testing this in Dev now. Thinking we could try moving it to Test? @argush3 @vysakh-menon-aot what do you think 👀

Just saw Argus’s reply in the Teams channel, please ignore my answer

PR #3751, opened 2025-09-04
30276 - Set submitter_roles to staff if filing was done by staff

JazzarKarim:
*Issue #:* /bcgov/entity#30276

*Description of changes:*
Before:
before migration submitter roles

After:
after migration submitter roles

The submitter_roles is now set to staff if the filing was done as staff in COLIN

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,7 @@ celerybeat-schedule
 .venv
 env/
 venv/
+venv-*/
 ENV/
 env.bak/
 venv.bak/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -83,6 +83,7 @@ celerybeat-schedule
 .venv
 env/
 venv/
+venv-*/
```
JazzarKarim, line 86: `venv-prefect` files were trying to be committed. So I decided to add this here.

vysakh-menon-aot (approved):

PR #3741, opened 2025-08-29
29967 - Non Filing Permissions

Rajandeep98:
*Issue #:* /bcgov/entity###

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #3755, opened 2025-09-05
30412 - update to skip any check for email or phone based on permission

ketaki-deodhar:
*Issue #:* /bcgov/entity#30413

*Description of changes:*

update to validate nr endpoint to skip any check for email or phone based on permission

Unit tests run for the test suite:

image

### DIFF
--- a/legal-api/tests/unit/resources/v2/test_name_requests.py
+++ b/legal-api/tests/unit/resources/v2/test_name_requests.py
@@ -19,15 +19,19 @@ Test-Suite to ensure that the /nameRequests endpoint is working as expected.
 from http import HTTPStatus
 
 import copy
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import datedelta
+from pytest_mock import mocker
 import pytz
 
 from legal_api.services import flags, namex
 from legal_api.utils.legislation_datetime import LegislationDatetime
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/services/permissions.py
```diff
@@ -28,6 +28,7 @@
 class ListFilingsPermissionsAllowed(str, Enum):
     """Define an enum for permissions checks."""
 
+    ADD_ENTITY_NO_AUTHENTICATION = 'ADD_ENTITY_NO_AUTHENTICATION'
```
vysakh-menon-aot, line 31: hmm. i don't think its correct either. This enum seems to be specific to filing permissions. Its better to be in ListActionsPermissionsAllowed (from Rajan's PR https://github.com/bcgov/lear/pull/3741/files#diff-aded57016dcbf8c9a8a0e8e2c87543f60cf2eeda674a32cfd01be70506398b36R61)

ketaki-deodhar: Okay. My misunderstanding! I have added the same enum as Rajan and added the permission in it.

ketaki-deodhar:
> please add unit test

There isn't actually any unit test for this particular functionality.  I will add unit tests for it.

PR #3790, opened 2025-09-16
30531 - Block Officers Filing

Rajandeep98:
*Issue #:* /bcgov/entity#30531

*Description of changes:*
FF Officer Filing


legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -540,6 +540,20 @@ def check_authorization(identifier, filing_json: dict,
...
+        ):
+            return (
+                jsonify({
+                    'message': (
+                        f'Officer Filing is currently not available for: '
+                        f'{identifier}.'
+                    )
+                }),
+                HTTPStatus.BAD_REQUEST
+            )
```
vysakh-menon-aot, line 556: Create a new function in `PermissionService` and move this, so that we can extend for other filings

PR #3792, opened 2025-09-16
30698 Fix flag import in DBC utils

loneil:
*Issue #:* /bcgov/entity#30698

*Description of changes:*
Change flag import to get this working after recent flag changes. Locally when using LD SDK key I can now get the correct flag values. Could only get flags loading in this file that also imports current_app if I do the import in the function as done in [some other example spots](https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/models/db.py#L60) now...

### DIFF
--- a/legal-api/Makefile
+++ b/legal-api/Makefile
@@ -57,12 +57,12 @@ pylint: ## Linting with pylint
 	fi
 
 flake8: ## Linting with flake8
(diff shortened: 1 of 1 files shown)


loneil:
Tests (relevant DC ones):

image

argush3 (approved):

PR #3805, opened 2025-09-19
29005 updates for publically available data

kialj876:
*Issue #:* /bcgov/entity#29005

*Description of changes:*
- add state filing id into slim json (public search users need it)
- add slim call to get filing (allows public to see a few things about a filing)
- add auth check on get filing behind new ff

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -321,8 +321,11 @@ class ListFilingResource():  # pylint: disable=too-many-public-methods
         if sub_type := filing.filing_sub_type:
             filing_json[filing_name]['type'] = sub_type
 
-        if filing.filing_json and (reason := filing.filing_json.get('filing', {}).get(filing_name, {}).get('reason')):
+        if filing.meta_data and (reason := filing.meta_data.get(filing_name, {}).get('reason')):
             filing_json[filing_name]['reason'] = reason
+
+        if filing.meta_data and (expiryDate := filing.meta_data.get(filing_name, {}).get('expiryDate')):
+            filing_json[filing_name]['expiryDate'] = expiryDate
 
         return jsonify({'filing': filing_json})
 
--- a/legal-api/tests/unit/resources/v2/test_business_filings/test_filings.py
+++ b/legal-api/tests/unit/resources/v2/test_business_filings/test_filings.py
@@ -17,6 +17,7 @@
 Test-Suite to ensure that the /businesses endpoint is working as expected.
 """
 import copy
+import json
 from datetime import datetime, timedelta, timezone
 from http import HTTPStatus
 from typing import Final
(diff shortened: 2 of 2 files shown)


legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -300,6 +300,43 @@ def get_single_filing(identifier: str, filing_id: int):
...
+            public_root_filing_keys = ['header', filing_name]
+            filing_json['filing'] = get_redacted_json(filing_json['filing'], public_root_filing_keys)
+            
+            # only need 'name' and 'effectiveDate' from 'header'
+            public_header_keys = ['name', 'effectiveDate']
+            filing_json['filing']['header'] = get_redacted_json(filing_json['filing']['header'], public_header_keys)
+
+            # only need the following keys when they are available in the nested filing info
+            public_filing_keys = ['dissolutionType', 'dissolutionDate', 'type', 'reason']
+            filing_json['filing'][filing_name] = get_redacted_json(filing_json['filing'][filing_name], public_filing_keys)
```
vysakh-menon-aot, line 325: Is this for COMPLETED filing? If so, I wouldn't suggest to depend on filing_json for these data (we try not to refer filing_json after processing, unless its not available in filing table or meta_data). It won't be available for migrated filings.

for instance
dissolutionType and type is in filing_sub_type
dissolutionDate is in metadata https://github.com/bcgov/lear/blob/main/queue_services/business-filer/src/business_filer/filing_processors/dissolution.py#L106

legal-api/src/legal_api/models/business.py
```diff
@@ -636,13 +636,18 @@ def _slim_json(self):
 
         if self.tax_id:
             d['taxId'] = self.tax_id
+        
+        if self.state_filing_id:
```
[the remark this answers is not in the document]
kialj876: Yes, correct

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
kialj876: So I'm trying to make sure this follows the same flow as the regular call. Right now this endpoint returns the CoreFiling json, which already takes previous versions of the filing json into account. This is returning the same data as the regular call through the same flow and then redacting most of it. I don't think we want this to have its own logic for putting together the filing json? Anything like what you are describing should be handled in the 'CoreFiling -> get_json' so that its handled for the non slim cases as well right so I don't think this part should change?
Alternatively I could put this at the end of the call so that its more clear that this is just redacting data from the existing call, but that will be a bit messier due to the auth check and the original/raw flow which are only needed for the non slim response

kialj876: The business dashboard is using this endpoint right now to get data on the state filing for its tombstone, which is why I was planning to use the same endpoint for the tombstone information in search, but if this endpoint is deprecated then I'll look at the other endpoint

vysakh-menon-aot: You can still use this endpoint, the only point i am raising is we cannot depend on filing json

kialj876: Ah okay I see what you mean. And then it won't return values for the migrated businesses. Okay I will just update it to use the model where possible for the 'slim' flow

legal-api/src/legal_api/models/business.py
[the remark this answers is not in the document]
severinbeauvais: Does this conversation still need to be addressed?

PR #3807, opened 2025-09-22
30809 - Certify Name Validation Removed

Rajandeep98:
*Issue #:* /bcgov/entity#30809

*Description of changes:*
Holding certify Name change till Certify statement work is finalized

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/validation.py
+++ b/legal-api/src/legal_api/services/filings/validations/validation.py
@@ -80,13 +80,6 @@ def validate(business: Business,  # pylint: disable=too-many-branches,too-many-s
         error = PermissionService.check_user_permission(required_permission, message)
         if error:
             return error
-
-    if validate_certify_name(filing_json) and flags.is_on('enabled-deeper-permission-action'):
-        required_permission = ListActionsPermissionsAllowed.EDITABLE_CERTIFY_NAME.value
-        message = f'Permission Denied - You do not have permissions to change certified by in this filing.'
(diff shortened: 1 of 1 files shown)


legal-api/src/legal_api/services/filings/validations/validation.py
```diff
@@ -80,13 +80,6 @@ def validate(business: Business,  # pylint: disable=too-many-branches,too-many-s
         error = PermissionService.check_user_permission(required_permission, message)
         if error:
             return error
-
-    if validate_certify_name(filing_json) and flags.is_on('enabled-deeper-permission-action'):
-        required_permission = ListActionsPermissionsAllowed.EDITABLE_CERTIFY_NAME.value
-        message = f'Permission Denied - You do not have permissions to change certified by in this filing.'
-        error = PermissionService.check_user_permission(required_permission, message)
-        if error:
-            return error
```
severinbeauvais, line 89: I thought this would only be disabled (commented out) while a decision is pending on whether to fully remove this or not?

severinbeauvais:
Why is there a lint error?

vysakh-menon-aot (approved):

legal-api/src/legal_api/services/filings/validations/validation.py
severinbeauvais: Thanks.

The background on this is that the current code blocks some currently-valid scenarios (ie, where I change the certify name). But, this is pending a re-design. We may get rid of certify name altogether.

I prefer to keep this code in place, but commented out, in case we need it for the redesign.

PR #4129, opened 2026-03-03
32594 - Update Emailer to Retrieve email recipient from Auth when ContactPoint is not Provided

meawong:
*Issue #:* /bcgov/entity#32594

*Description of changes:*

### DIFF
--- a/queue_services/business-emailer/src/business_emailer/email_processors/__init__.py
+++ b/queue_services/business-emailer/src/business_emailer/email_processors/__init__.py
@@ -53,10 +53,7 @@ def get_recipients(option: str, filing_json: dict, token: str | None = None, fil
     recipients = ""
     filing_type = filing_type if filing_type else "incorporationApplication"
     if filing_json["filing"].get(filing_type):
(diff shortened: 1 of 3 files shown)


queue_services/business-emailer/src/business_emailer/email_processors/__init__.py
```diff
@@ -53,7 +53,7 @@ def get_recipients(option: str, filing_json: dict, token: str | None = None, fil
     recipients = ""
     filing_type = filing_type if filing_type else "incorporationApplication"
     if filing_json["filing"].get(filing_type):
-        recipients = filing_json["filing"][filing_type]["contactPoint"]["email"]
+        recipients = filing_json["filing"][filing_type].get("contactPoint", {}).get("email")
```
loneil: Think this should still be `.get("email", "")`
the return type for the function is `-> str` but this could yield `None` right? Not sure of downstream expectations but the references to this would expect a string result.
(also down below in `recipients = f"{recipients}, {comp_party_email}"` would get `None` put in the comma separated list

vysakh-menon-aot (approved):

loneil (approved):

PR #4235, opened 2026-03-30
32211: Currency validation

loneil:
*Issue #:* /bcgov/entity#32211

*Description of changes:*

Adds support for the legacy OTHER currency type on share classes migrated from COLIN and validates that currency values match recognized currency codes. 
- Additional to OTHER type handling, currency validation was previously just done on UI and missing from API (tracked in previous FE/BE validation mismatch analysis and left to this point to handle)

**Model**

- Return `currencyAdditional` in ShareClass.json so UIs can display the OTHER currency text

**Validation** (`validate_share_currency`)

-Validate that share class currency is a recognized currency code (matching the standard currency list used by the UI dropdowns), only when `hasParValue` is true
-Grandfathered **OTHER** share classes pass through when the incoming share class ID matches an existing OTHER record in the DB — this handles the round-trip scenario where an alteration payload includes unchanged OTHER classes alongside modified non-OTHER classes
-New series under grandfathered OTHER share classes are blocked

**Filing coverage**

- Alteration, Correction, Transition: pass business for ID matching (allows existing OTHER pass-through)
- Incorporation Application, Continuation In, Amalgamation (regular): no business context, all currencies must be valid

### DIFF
--- a/legal-api/src/legal_api/models/share_class.py
+++ b/legal-api/src/legal_api/models/share_class.py
@@ -67,6 +67,7 @@ class ShareClass(db.Model, Versioned):  # pylint: disable=too-many-instance-attr
             "hasParValue": self.par_value_flag,
             "parValue": self.par_value,
             "currency": self.currency,
+            "currencyAdditional": self.currency_additional,
             "hasRightsOrRestrictions": self.special_rights_flag
         }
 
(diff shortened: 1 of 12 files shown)


Rajandeep98 (approved):
lgtm

meawong (approved):

PR #4262, opened 2026-04-15
33028 - Integrate Outputs Flag Flip For Migration Candidates

Rajandeep98:
*Issue #:* /bcgov/entity#33028

*Description of changes:*
Process Paper Filing  Status from Extract to Target DB


argush3:
think this looks good but we need to add the config that controls whether we want to actually show the available legacy outputs.  for example, most of the time we won't have these outputs available in dev/test.

Take a look at https://github.com/bcgov/lear/pull/4287.  I think this experimental PR i created is essentially what you have but with the config.  I've default the config to not show legacy outputs.  In Prod, we would need to ensure we set `TOMBSTONE_USE_SOURCE_PAPER_ONLY` = True when we migrate.

argush3 (approved):

PR #4311, opened 2026-04-23
Create flow to set AR reminder to N in COLIN

ketaki-deodhar:
*Issue #:* /bcgov/entity#32596

*Description of changes:*

- Create flow to set AR reminder to N in COLIN
- Tested in CTST with a batch of size 1382. Reverted the flag in CTST after testing

### DIFF
--- a/data-tool/flows/update_colin_ar_ind_flow.py
+++ b/data-tool/flows/update_colin_ar_ind_flow.py
@@ -58,7 +58,7 @@ def chunk_list(data: List[str], size: int = 1000):
     for i in range(0, len(data), size):
         yield data[i:i + size]
 
-@task(retries=2, retry_delay_seconds=5)
+@task
 def update_chunk(ids_chunk: List[str]) -> int:
     corps_str = ",".join(f"'{corp}'" for corp in ids_chunk)
     sql = UPDATE_AR_IND_QUERY.format(corps=corps_str)
--- a/data-tool/flows/update_colin_ar_ind_flow.py
+++ b/data-tool/flows/update_colin_ar_ind_flow.py
@@ -1,0 +1,98 @@
+from typing import List
+
+from prefect import flow, task
+from prefect.cache_policies import NO_CACHE
+from sqlalchemy import Engine, bindparam, text
+
+from common.init_utils import colin_extract_init, colin_oracle_init, get_config
+
+# --- MIG corp selection (from COLIN) used when USE_MIGRATION_FILTER = True ---
+MIG_CORP_FILTER_BASE = """
+    SELECT DISTINCT mcb.corp_num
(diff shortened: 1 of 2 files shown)


data-tool/flows/update_colin_ar_ind_flow.py
```diff
@@ -0,0 +1,88 @@
...
+        # Strip "BC" prefix if present
+        colin_identifiers = [corp[2:] if corp.startswith("BC") else corp for corp in candidates]
+        return colin_identifiers
+
+def chunk_list(data: List[str], size: int = 1000):
+    # chunk size is set to 1000 to avoid hitting limit on oracle database
+    for i in range(0, len(data), size):
+        yield data[i:i + size]
+
+@task(retries=2, retry_delay_seconds=5)
```
argush3: let's avoid retry logic for now.  when COLIN times or there are connection issues, we probably want to investigate what the issue is first.

data-tool/flows/update_colin_ar_ind_flow.py
```diff
@@ -0,0 +1,89 @@
...
+    JOIN mig_batch b ON b.id = mcb.mig_batch_id
+    JOIN mig_group g ON g.id = b.mig_group_id
+    WHERE 1 = 1
+"""
+
+UPDATE_AR_IND_QUERY = """
+    UPDATE corporation c
+    SET c.SEND_AR_IND = 'N'
+    WHERE c.corp_num IN ({corps})
+    AND c.SEND_AR_IND <> 'N'
```
ketaki-deodhar: Added this so that we will get exact number of records updated

argush3: Maybe update to `AND (c.SEND_AR_IND IS NULL OR c.SEND_AR_IND <> 'N')`.

I believe `c.SEND_AR_IND <> 'N'` skips NULL values

data-tool/flows/update_colin_ar_ind_flow.py
```diff
@@ -0,0 +1,89 @@
...
+
+@task
+def update_chunk(ids_chunk: List[str]) -> int:
+    corps_str = ",".join(f"'{corp}'" for corp in ids_chunk)
+    sql = UPDATE_AR_IND_QUERY.format(corps=corps_str)
+
+    with colin_oracle_init(get_config()).connect() as conn:
+        result = conn.execute(text(sql))
+        conn.commit()
+        return result.rowcount
```
argush3: how about we just initialize the db connection once from the start of the flow and use it for updating each chunk.

e.g.
Image

argush3 (changes requested):
looks good.

just the two things i've commented on.

argush3 (approved):

PR #4323, opened 2026-04-24
API enable replacing of ledger filing reports, update report config.

doug-lovett:
*Issue #:* /bcgov/entity#33217

*Description of changes:*
Add the ability to regenerate a filing report when the report data changes, such as when a BN is assigned to a new business. Add a new optional boolean request parameter regenerate=true to GET filing report requests. If this parameter is true the API will:

Regenerate the report with a report service call.
With the existing DRS record ID replace the pdf data with a recently added DRS API PUT endpoint call: PUT /doc/api/v1/application-reports/BUSINESS/{identifier}
Also update the API reports configuration so additional filings work properly - as a distinct DRS API report type.

*Issue #:* /bcgov/entity#33171

*Description of changes:*
Not all migrated Colin filings have receipts. For example, in TEST the 2025 AR for BC1252030 has no receipt: a screenshot of the PROD colin ledger is attached to the ticket. The business api assumes that all filings have receipts, so if a colin migrated filing has no receipt the business api is trying to get the receipt from the pay api and returning an error.

Fix: update the business api to conditionally remove the receipt link from the ledger filing documents when the filing documents have migrated from colin and no receipt exists in the DRS.

### DIFF
--- a/legal-api/src/legal_api/reports/document_service.py
+++ b/legal-api/src/legal_api/reports/document_service.py
@@ -17,6 +17,7 @@ from typing import Optional
 import requests
 from flask import current_app, jsonify
 
+from legal_api.core.meta.filing import FILINGS
 from legal_api.exceptions import BusinessException
 from legal_api.models import Business, Document, Filing
 from legal_api.services import AccountService
@@ -24,6 +25,7 @@ from legal_api.utils.base import BaseEnum
 
 BUSINESS_DOCS_PATH: str = "{url}/application-reports/history/{product}/{business_id}?includeDocuments=true"
 GET_REPORT_PATH: str = "{url}/application-reports/{product}/{drsId}"
+PUT_REPORT_PATH: str = "{url}/application-reports/{product}/{drsId}"
 GET_REPORT_CERTIFIED_PATH: str = "{url}/application-reports/{product}/{drsId}?certifiedCopy=true"
 POST_REPORT_PATH: str = "{url}/application-reports/{product}/{bus_id}/{filing_id}/{report_type}"
 POST_REPORT_PARAMS: str = "?consumerFilingDate={filing_date}&consumerFilename={filename}"
@@ -68,6 +70,9 @@ class ReportTypes(BaseEnum):
 
     CERT = "CERT"
     FILING = "FILING"
+    FILING_2 = "FILING-2"
+    FILING_3 = "FILING-3"
+    FILING_4 = "FILING-4"
     NOA = "NOA"
     RECEIPT = "RECEIPT"
 
@@ -251,7 +256,7 @@ class DocumentService:
         except Exception:
             return []
 
-    def update_document_list(self, drs_docs: list, document_list: list) -> list:
+    def update_document_list(self, drs_docs: list, document_list: list, filing: Filing) -> list:
         """
         Update document_list urls with DRS information if a filing document maps to a DRS output or document.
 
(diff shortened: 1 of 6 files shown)


legal-api/src/legal_api/reports/document_service.py
```diff
@@ -495,3 +554,41 @@ def _get_request_headers(self, account_id: str, accept_mime_type = None) -> dict
         if accept_mime_type:
             headers["Accept"] = accept_mime_type
         return headers
+
+    def _update_legal_filing(self, doc_list: dict, drs_params: str) -> dict:
+        """Add the DRS params to the legal filing document url."""
+        if doc_list and doc_list.get("legalFilings"):
+            legal_filing = doc_list["legalFilings"][0]
```
kialj876, line 561: Just to clarify do we never expect this to be empty or more than 1 item?

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -495,3 +554,41 @@ def _get_request_headers(self, account_id: str, accept_mime_type = None) -> dict
...
+            filing_key = next(iter(legal_filing.keys()))
```
kialj876, line 562: So is this method to specifically update the first item in the legal_filing dict only - are we assuming there's always just one key in the dict?
If I am following correctly we are relying on this always being the specific filing document right (i.e. if it was an incorporation filing we would expect this to only contain _incorporationApplication_)?

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -495,3 +554,41 @@ def _get_request_headers(self, account_id: str, accept_mime_type = None) -> dict
...
+            legal_filing[filing_key] = legal_filing[filing_key] + drs_params
+        return doc_list
+
+    def _update_additional_filing(self, doc_list: dict, drs_params: str) -> dict:
+        """Conditionally add the DRS params to the additional filing document url."""
+        if doc_list.get("specialResolutionApplication"):  # Not in configuration
```
kialj876, line 568: Might be good to us the walrus here and the other 'ifs' below to try reduce the reading in here
i.e.
```
        if spec_res := doc_list.get("specialResolutionApplication"):  # Not in configuration
            spec_res = spec_res + drs_params
```

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -495,3 +554,41 @@ def _get_request_headers(self, account_id: str, accept_mime_type = None) -> dict
...
+            doc_list["specialResolutionApplication"] = doc_list["specialResolutionApplication"] + drs_params
+            return doc_list
+        if doc_list.get("legalFilings"):
+            legal_filing = doc_list["legalFilings"][0]
+            filing_key = next(iter(legal_filing.keys()))
+            if FILINGS.get(filing_key) and FILINGS[filing_key].get("additional"):
```
kialj876, line 574: Just to clarify I am following correctly: 
- we are checking the hardcoded config of the 'filing_key' filing to check if there's expected additional outputs
- For each expected additional output, if it exists in the doc_list and its not an noa or certificate, then we add the drs params to it.

We are ignoring noa and certificate because they are updated specifically so otherwise this would end up adding extra drs_params right?

What is an example of an additional output that would be actioned here?

kialj876 (commented):
Hey Doug, this is looking good I think, but I've got a couple questions for you below.

Also, does it make sense to add a new test for the regenerate flow here?

legal-api/src/legal_api/reports/document_service.py
doug-lovett: So far it is always exactly 1. If that ever changes the config for outputs/documents will need to be updated. I am not sure what the purpose is in distinguishing one document from the others.

legal-api/src/legal_api/reports/document_service.py
doug-lovett: Yes, so far there is exactly 1 legalFilings document. From my above comment, if for some reason a filing type is added/modified to have 2 legalFilings, then I think the scope of the change is larger and a filing outputs config update will be needed.

legal-api/src/legal_api/reports/document_service.py
doug-lovett: Yes you are following correctly, the additional documents for a filing type are determined by looking at the config, ignoring NOA and certificates which are already identified as distinct report types. For example, a firm changeOfRegistration has an additional amendedRegistrationStatement document.

legal-api/src/legal_api/reports/document_service.py
doug-lovett: I considered this but am on the fence because of the conditions excluding noa's and cert's may make it more confusing so decided to keep as is?

kialj876: yeah no worries if you think its better as is

kialj876 (approved):
Looks good to me, but please wait for another approval from one of the others before merging.

I also think it would be good to add a unit test for the colin receipt removal and another one for regenerating the drs pdf if you can

doug-lovett:
@kialj876 I totally agree with the unit test comment. There are no existing ones to build off of and the mock service set up is involved, so for want of time I only locally tested the regenerate and colin receipt removal.

legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
```diff
@@ -133,6 +137,13 @@ def _get_drs_params() -> dict:
     return params
 
 
+def _regenerate(legal_filing_name: str) -> bool:
+    """Determine if individual report request should regenerate and update the DRS."""
+    if legal_filing_name and legal_filing_name.lower().startswith("receipt"):
+        return False
+    return request.args.get(PARAM_REGENERATE, False)
```
argush3: think any non-empty string passed into the "regenerate" param for the GET request will result in doc being regenerated right now.

doug-lovett: Thanks Argus, I have added some param value checking.

argush3 (approved):

PR #4328, opened 2026-04-27
33224 business-registry-model: Add additional currency serialization

loneil:
*Issue #:* /bcgov/entity#33224

*Description of changes:*
Match the changes made to the legal-api models for `currency_additional`. Some were never made in this `business-registry-model` duplicate.

### DIFF
--- a/python/common/business-registry-model/pyproject.toml
+++ b/python/common/business-registry-model/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "business-model"
-version = "3.3.24"
+version = "3.3.25"
 description = ""
 authors = [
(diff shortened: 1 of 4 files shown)


python/common/business-registry-model/pyproject.toml
kialj876, line 11: Just merged Anish's dependency update PR for sql-versioning so we may need to specify the version of this package if the new one causes any issues

kialj876 (approved):
Looks good to me 👍

meawong (approved):

python/common/business-registry-model/pyproject.toml
loneil: K, will merge this and see if there's any build problem from that when importing into a reliant service

PR #4350, opened 2026-04-30
32792 - Add Validation for Custodian First Name in VD

meawong:
*Issue #:* /bcgov/entity#32792 

*Description of changes:*
- Refactor existing `validate_custodian_org_name` and add logic for VD party first name validation
- Update unit tests

Examples:
image
image
image

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/dissolution.py
+++ b/legal-api/src/legal_api/services/filings/validations/dissolution.py
@@ -405,14 +405,13 @@ def _validate_custodian_name(parties, dissolution_type, legal_type) -> list:
 
         else:
             first_name = get_str(party, "/officer/firstName")
-            stripped_first_name = first_name.strip()
 
-            if not stripped_first_name:
+            if first_name is None or not first_name.strip():
                 msg.append({
                     "error": "Custodian first name is required.",
                     "path": f"/filing/dissolution/parties/{idx}/officer/firstName"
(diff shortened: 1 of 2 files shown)


kialj876 (approved):
Nice work, lgtm

legal-api/src/legal_api/services/filings/validations/dissolution.py
```diff
@@ -405,6 +403,21 @@ def validate_custodian_org_name(parties, dissolution_type, legal_type) -> list:
                     "path": f"/filing/dissolution/parties/{idx}/officer/organizationName"
                 })
 
+        else:
+            first_name = get_str(party, "/officer/firstName")
+            stripped_first_name = first_name.strip()
```
loneil: Tried quick locally on this branch and Llooks like can get through to this point without providing `firstName` at all so this excepts with `AttributeError: 'NoneType' object has no attribute 'strip'` and falls back with the generic handler.
Something like
`if first_name is None or not first_name.strip():` could handle that

loneil (changes requested):

legal-api/src/legal_api/services/filings/validations/dissolution.py
meawong: updated:
image

loneil (approved):

PR #4351, opened 2026-05-01
Legal-api: business/search filtering bug fixes and test suite add

kialj876:
*Issue #:* /bcgov/entity#31582

*Description of changes:*
- return draft temp regs against their nr number in the identifier filter
- updated service method typing to be consistent
- fixed a bug in the process where filters would match on any existing business if only temp reg identifiers were supplied (i.e. if an account only has a draft incorp and they filter for some reason - super unlikely scenario - then the filter will return results on all businesses in the db
- test suite for this endpoint added

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business.py
+++ b/legal-api/src/legal_api/resources/v2/business/business.py
@@ -262,7 +262,6 @@ def search_businesses():
             identifiers=business_identifiers,
             search_filters=search_filters) or ([], False)
         draft_results, draft_hasmore = BusinessSearchService.get_search_filtered_filings_results(
-            business_json=json_input,
             identifiers=temp_identifiers,
             search_filters=search_filters) or ([], False)
         has_more = bus_hasmore or draft_hasmore
--- a/legal-api/src/legal_api/services/search_service.py
+++ b/legal-api/src/legal_api/services/search_service.py
@@ -141,29 +141,30 @@ class BusinessSearchService:  # pylint: disable=too-many-public-methods
     # pylint: disable=too-many-locals # pylint: disable=too-many-branches # pylint: disable=too-many-statements
     @staticmethod
     def get_search_filtered_businesses_results(business_json,
-                                               identifiers=None,
-                                               search_filters: AffiliationSearchDetails = None):
+                                               identifiers: list[str],
+                                               search_filters: AffiliationSearchDetails) -> Optional[tuple[list, bool]]:
         """Return contact point from business json."""
+        if not identifiers:
+            return None
+
         name = search_filters.name
         types = search_filters.type
         statuses = search_filters.status
         identifier = search_filters.identifier
 
         valid_types, _ = BusinessSearchService.separate_legal_types(types or [])
-
         # Edge case: if type searched for doesnt belong to table business such as 'ATMP' Returns early
         if types and not valid_types:
-            return []
+            return None
 
         states, _ = BusinessSearchService.separate_states_by_type(statuses or [])
-
         # If status was provided but none of them are valid, return no results
         if statuses and not states:
-            return []
+            return None
+
         filters = [
             expr for expr in [
-                Business._identifier.in_(identifiers)  # pylint: disable=protected-access
-                if identifiers else None,
+                Business._identifier.in_(identifiers),  # pylint: disable=protected-access
                 Business._identifier.ilike(f"%{identifier}%")  # pylint: disable=protected-access
                 if identifier else None,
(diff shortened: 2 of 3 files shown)


legal-api/src/legal_api/resources/v2/business/business.py
```diff
@@ -262,7 +262,6 @@ def search_businesses():
             identifiers=business_identifiers,
             search_filters=search_filters) or ([], False)
         draft_results, draft_hasmore = BusinessSearchService.get_search_filtered_filings_results(
-            business_json=json_input,
```
kialj876, line 265: is unused

legal-api/src/legal_api/services/search_service.py
```diff
@@ -141,29 +141,30 @@ def separate_legal_types(legal_types: list[str]) -> tuple[list[str], list[str]]:
     # pylint: disable=too-many-locals # pylint: disable=too-many-branches # pylint: disable=too-many-statements
     @staticmethod
     def get_search_filtered_businesses_results(business_json,
-                                               identifiers=None,
-                                               search_filters: AffiliationSearchDetails = None):
+                                               identifiers: list[str],
+                                               search_filters: AffiliationSearchDetails) -> Optional[tuple[list, bool]]:
         """Return contact point from business json."""
+        if not identifiers:
+            return None
```
kialj876, line 148: bug fix

legal-api/src/legal_api/services/search_service.py
```diff
@@ -208,42 +209,50 @@ def get_search_filtered_businesses_results(business_json,
...
-                Filing.temp_reg.ilike(f"%{identifier}%") if identifier else None,
+                or_(
+                    Filing.temp_reg.ilike(f"%{identifier}%"),
+                    func.jsonb_extract_path_text(
+                        Filing._filing_json,  # pylint: disable=protected-access
+                        "filing",
+                        Filing._filing_type,  # pylint: disable=protected-access
+                        "nameRequest",
+                        "nrNumber"
+                    ).ilike(f"%{identifier}%"))
```
kialj876, line 254: NR match for draft filings on identifier filter

legal-api/src/legal_api/services/search_service.py
```diff
@@ -265,33 +274,30 @@ def get_search_filtered_filings_results(business_json,
...
-                                      .get("nameRequest", {})
-                                      .get("legalName"))
-
-                if draft["legalName"] is None:
-                    # Fallback to a generic legal name based on the legal type if no specific legal name is found
-                    draft["legalName"] = (Business.BUSINESSES
-                                          .get(draft_dao.json_legal_type, {})
-                                          .get("numberedDescription"))
-                draft_results.append(draft)
+        for draft_dao in draft_query[:limit]:
```
kialj876, line 277: just spacing change as the 'if identifiers:' was redundant

legal-api/src/legal_api/services/search_service.py
loneil: There's no index on `temp_reg` from what it looks like so I was thinking about whether this is introducing heavy load or anything but I _think_ the json extract part is only a consideration for records in a small enough set (`temp_reg IN (...)`). So the JSONB path only gets evaluated on the handful of rows that already passed the IN check, not the whole draft pool.

However, the missing temp_req index might be something someone would want to take into account at some point (not sure if there's DBAs that profile system-wide things like this or anything). 1.3 mil filings in prod and that doesn't account for upcoming migrations and possibility of lots more drafts with usage...

Not related to this PR though it looks like.

loneil (approved):
Looks good I think. I don't know the search queries too well but change looks like it should do the trick.
Commented an aside on the filtering but something to think of outside this PR

kialj876:
> Looks good I think. I don't know the search queries too well but change looks like it should do the trick. Commented an aside on the filtering but something to think of outside this PR

Yeah this is a good point. It is probably okay due to 'temp_reg' being null for most filings and that it needs to be draft filings (even with the influx of corps I think this should make it a fairly low amount), but this is definitely something to keep an eye on

legal-api/src/legal_api/services/search_service.py
kialj876: Yeah this is a good point. It is probably okay due to 'temp_reg' being null for most filings and that it needs to be draft filings (even with the influx of corps I think this should make it a fairly low amount), but this is definitely something to keep an eye on

meawong (approved):
Nice LGTM!

PR #4352, opened 2026-05-01
33002 - Remove certified copy stamp from Letter of Consent

mruff-aeq:
#33002 : /bcgov/entity#33002

*Description of changes:*
The "Certified Copy" stamp is triggered when the code calls the Document Retrieval Service (DRS) using a URL that includes the `?certifiedCopy=true` query parameter. By creating the FILING-3 report type, we are able to adjust the logic to omit that specific query parameter from the URL for `letterOfConsent`, which prevents DRS from injecting the stamp into the PDF.

For verification, implemented a parameterized unit test that enforces the correct URL structure is generated for each specific report type, so it is ensured that the `?certifiedCopy=true` query parameter is only applied to ` FILING, FILING-2, and NOA`, and explicitly omitted for `letterOfConsent` using FILING-3.

### DIFF
--- a/legal-api/src/legal_api/reports/document_service.py
+++ b/legal-api/src/legal_api/reports/document_service.py
@@ -344,7 +344,7 @@ class DocumentService:
         headers = self._get_request_headers(BUSINESS_API_ACCOUNT_ID, APP_PDF)
         url: str = self.url.replace(DOC_PATH, "")
         get_url = GET_REPORT_PATH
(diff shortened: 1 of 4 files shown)


mruff-aeq:
image

Only the letterOfConsent has been applied to the new FILING-3 type, and it is now correctly generating PDFs without the "Certified Copy" watermark.

kialj876 (approved):
Nice work!

doug-lovett (approved):
Looks good.

meawong (approved):

PR #4360, opened 2026-05-06
Business model - test setup fixes

kialj876:

*Description of changes:*
- prevents data leaking between tests when multiple nested commits are called during the test setup
  - note that the tests in here currently avoid/hide this issue by applying unique keys in every test (i.e. with new random ids every time)
- updated naics tests to work (they were failing before due to referencing test data from an older migration that wasn't brought into the common model migrations)

### DIFF
--- a/python/common/business-registry-model/src/business_model/models/filing.py
+++ b/python/common/business-registry-model/src/business_model/models/filing.py
@@ -1363,10 +1363,11 @@ class Filing(db.Model):  # pylint: disable=too-many-instance-attributes,too-many
                 ~Business.legal_type.in_(excluded_businesses),
                 ~Filing._filing_type.in_(excluded_filings),
                 ~and_(Filing._filing_type == 'dissolution', Filing._filing_sub_type == 'delay'),
-                Filing.colin_event_ids == None,  # pylint: disable=singleton-comparison
+                or_(~Filing.lear_only, Filing.lear_only == None),
+                Filing.colin_event_ids == None,
                 Filing._status == Filing.Status.COMPLETED.value,
                 Filing._source == Filing.Source.LEAR.value,
-                Filing.effective_date != None   # pylint: disable=singleton-comparison
+                Filing.effective_date != None
             ).order_by(Filing.transaction_id).limit(limit).offset(offset).all()
 
         return filings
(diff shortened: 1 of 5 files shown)


loneil (approved):
Pulled locally and tests pass. Still passing on CI on this PR

python/common/business-registry-model/tests/models/test_naics_structure.py
```diff
@@ -51,12 +47,11 @@ def test_exact_match_search_naics(app, session, client):
     # check
     assert results
     assert len(results) == 1
-    assert len(results[0].naics_elements) == 15
+    assert len(results[0].naics_elements) == 10
```
argush3, line 50: interesting that the result changed

kialj876: yeah see my description above - the test results are based on the test data which is pulled via migration files and the data these original tests were based on was not included when it was moved to common. The other thing that's important to know about this is that the migration files pull in test data when the app environment is set to testing so I believe this was purely a test setup misconfiguration issue

kialj876: that's why the 'naics_key' is different and its the reason all these were skipped before and tagged as not working

python/common/business-registry-model/src/business_model/models/filing.py
```diff
@@ -1363,10 +1363,11 @@ def get_completed_filings_for_colin(limit=20, offset=0):
                 ~Business.legal_type.in_(excluded_businesses),
                 ~Filing._filing_type.in_(excluded_filings),
                 ~and_(Filing._filing_type == 'dissolution', Filing._filing_sub_type == 'delay'),
-                Filing.colin_event_ids == None,  # pylint: disable=singleton-comparison
+                or_(~Filing.lear_only, Filing.lear_only == None),
```
kialj876, line 1366: consuming the recent change from vysakh

vysakh-menon-aot: Thanks. I was preparing another pr for this one

vysakh-menon-aot (approved):

argush3 (approved):

PR #3409, opened 2025-04-15
25840 - update Legal API endpoint and  Expired limited Restoration job to add legal type

ketaki-deodhar:
*Issue #:* /bcgov/entity#25840

*Description of changes:*

**Issue**: Businesses were not being synced back to COLIN as Historical after an expired limited restoration job run was complete.
Syncing was failing because legal type was not being set in filing json and Update Colin filings jobs expects it
Ref: https://github.com/bcgov/lear/blob/main/jobs/update-colin-filings/update_colin_filings.py#L80

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #4386, opened 2026-05-12
Business Model - minor consolidation between legal-api model and common

kialj876:

*Description of changes:*
- looks like this is the last change needed for the legal-api to use the model fully in python 3.13

### DIFF
--- a/python/common/business-registry-model/src/business_model/models/business.py
+++ b/python/common/business-registry-model/src/business_model/models/business.py
@@ -851,7 +851,7 @@ class Business(db.Model, Versioned):  # pylint: disable=too-many-instance-attrib
     @classmethod
     def get_expired_restoration(cls):
         """Return all identifier with an expired restoration_expiry_date."""
-        businesses = (db.session.query(Business.identifier)
+        businesses = (db.session.query(Business.identifier, Business.legal_type)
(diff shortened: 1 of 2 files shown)


python/common/business-registry-model/src/business_model/models/business.py
```diff
@@ -851,7 +851,7 @@ def get_all_by_no_tax_id(cls):
     @classmethod
     def get_expired_restoration(cls):
         """Return all identifier with an expired restoration_expiry_date."""
```
loneil: Super minor, but in legal-api in the change where the extra param was introduced the docstring was changed to `"""Return business identifiers with an expired restoration_expiry_date."""`

https://github.com/bcgov/lear/pull/3409/changes#diff-9b455ec1b04a8c3f09e4d25c08604d5ddbec50dd4cff328febc08734b90ad1a0

loneil (approved):

argush3 (approved):

vysakh-menon-aot (approved):

meawong (approved):

python/common/business-registry-model/src/business_model/models/business.py
kialj876: Good point - will update it to include mentioning the legal type too

PR #3341, opened 2025-03-25
26611 fix ar min/max date calculation

vysakh-menon-aot:
*Issue #:* /bcgov/entity#26611

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #4387, opened 2026-05-13
33247 legal api dep upgrade

kialj876:

*Description of changes:*
- upgrade legal-api to 3.13 and all major deps
- replace model with common model
- see below comments for specific notes on things


legal-api/tests/unit/models/__init__.py
kialj876, line 1: NOTE: I am planning to remove all the model tests before merging as they should be secured in the common model tests. 
I didn't want to do this before verifying that they were all passing here with minimal changes in case there were unknown differences etc

legal-api/tests/unit/models/test_configuration.py
kialj876, line 1: NOTE: updates in here follow what was accepted in the common model - I think this is okay, but let me know if you disagree

legal-api/tests/unit/models/test_naics_structure.py
kialj876, line 1: NOTE: these updates and other naics related test changes were already applied and approved in the common model as well and are just due to the test data populated being a little different in the migrations of the common model

legal-api/src/legal_api/logging.conf
kialj876, line 1: NOTE: logging is setup via our structured logging dependency - we removed this in our newer projects as well - I didn't dive too deep into whether or not to keep the older logging stuff around so let me know if you think I should bring it back

legal-api/src/legal_api/version.py
kialj876, line 1: NOTE: we use the version in the toml file instead now (same as newer projects)

legal-api/tests/conftest.py
kialj876, line 1: NOTE: similar changes made here that I made in the common model recently (full rollback of all transactions between each test) with one separate change needed for app_request

legal-api/tests/unit/services/utils.py
```diff
@@ -59,10 +60,32 @@ def helper_create_jwt(jwt_manager: JwtManager, roles: List[str] = [], username:
...
+    headers = {
+        'content-type': 'application/json',
+        'Authorization': 'Bearer ' + token,
+        **kwargs,
+    }
     return headers
 
 
+@contextmanager
+def jwt_request_context(app,
```
kialj876, line 72: NOTE: there was a breaking change in the upgraded version of our flask-jwt-oidc package where we need to pass in custom request context data to some of the functions we use regularly in this API so this ensures that specific request context data is available for tests that skip over the client

kialj876:
~Not sure why the CI isn't running the lint or unit tests here. It's supposed to run against 'feature*' branches~ *edit - running now*.
Below are my local runs:

Screenshot 2026-05-13 at 11 20 37 AM

Screenshot 2026-05-13 at 11 23 52 AM


I've also tested the major functionality locally: 
- getting business/address/party/filing/tasks data
- posting filings+drafts
- removing drafts
- getting filing document lists
- getting filing pdfs
- getting permissions
- will expand on this

vysakh-menon-aot (approved):

meawong (approved):

loneil (approved):
Can't see any obvious blockers or fundamental issues. I've never had a chance to need to dig into migrations but assume that part of it is handled from the note here.
Probably comes down to regression testing then.

legal-api/migrations/env.py
argush3, line 1: The migration env was explicitly setting a db role.  I think we're moving away from this but not sure if it will still be required for now.

```
owner_role = os.getenv("DATABASE_OWNER_ROLE")
...
connection.execute(sa.text(f'SET ROLE "{safe_role}"'))
```

legal-api/src/legal_api/models/filing.py
```diff
@@ -1,1532 +0,0 @@
...
-    def get_previous_completed_filing(filing):
-        """Return the previous completed filing."""
-        query = db.session.query(Filing). \
-            filter(Filing.business_id == filing.business_id). \
-            filter(Filing._status.in_([Filing.Status.COMPLETED.value, Filing.Status.TOMBSTONE.value]))
-
-        if filing.transaction_id:  # transaction_id will be None for the pending filings (intermediate state)
-            query = query.filter(Filing.transaction_id < filing.transaction_id)
-
-        return query.order_by(Filing.transaction_id.desc()).first()
```
argush3, line 1383: there is some [drift](https://github.com/bcgov/lear/blame/main/python/common/business-registry-model/src/business_model/models/filing.py#L1392-L1401) here that needs to be applied to the common model

legal-api/src/legal_api/models/business.py
```diff
@@ -1,996 +0,0 @@
...
-            year_2020: Final = 2020
-            if next_ar_year == year_2020:
-                # For year 2020, set the max date as October 31th next year (COVID extension).
-                ar_max_date = datetime(next_ar_year + 1, 10, 31).date()
-            else:
-                # If this is a CO-OP, set the max date as April 30th next year.
-                ar_max_date = datetime(next_ar_year + 1, 4, 30).date()
-        elif self.legal_type in self.CORPS:
-            # For BCOMP min date is next anniversary date.
-            _founding_date = LegislationDatetime.as_legislation_timezone(self.founding_date)
```
argush3, line 459: not sure if we need this line in common model.

@vysakh-menon-aot you had added this line via this [PR](https://github.com/bcgov/lear/pull/3341).  do we need to add this to common business model?

vysakh-menon-aot: yes it is required

argush3:
looks good.  just a few comments

legal-api/migrations/env.py
kialj876: hmm okay I will look into this and make sure

legal-api/src/legal_api/models/filing.py
kialj876: nice catch, thanks! Will check it out

legal-api/src/legal_api/models/business.py
kialj876: Okay sounds good - will update

legal-api/src/legal_api/models/filing.py
kialj876: pr: https://github.com/bcgov/lear/pull/4424

legal-api/src/legal_api/models/business.py
kialj876: pr: https://github.com/bcgov/lear/pull/4424

argush3 (approved):

kialj876:
Tests are green - code coverage is at 75% which is why it shows red

PR #4395, opened 2026-05-17
32911 - Add validation for foreign business legal name length

mruff-aeq:
#32911 : /bcgov/entity#32911

*Description of changes:*
We discovered that the backend wasn’t validating the length of a foreign amalgamating business’s legalName. The frontend correctly blocked names shorter than 3 characters or longer than 150, but the backend didn’t — so invalid values could still slip through and get saved.

The backend now enforces the same 3–150 character limit. If the name is outside that range, it returns a 400 error with the message “Length of foreign business legal name must be from 3 to 150 characters.”

Example:
when making a POST to url `/api/v2/businesses/<T-id>/filings?only_validate=true`
`"legalName": "AA"` in payload throws 400 error “Length of foreign business legal name must be from 3 to 150 characters.”
`"legalName": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"` in payload throws 400 error “Length of foreign business legal name must be from 3 to 150 characters.”

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
+++ b/legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
@@ -267,6 +267,16 @@ def _validate_foreign_businesses(  # noqa: PLR0913
         amalgamating_business,
         amalgamating_business_path) -> list:
     msg = []
+
+    # Frontend enforces these bounds on the foreign business legal name; mirror them here.
+    min_legal_name_length: Final = 3
+    max_legal_name_length: Final = 150
+    if not min_legal_name_length <= len(foreign_legal_name) <= max_legal_name_length:
+        msg.append({
+            "error": "Length of foreign business legal name must be from 3 to 150 characters.",
+            "path": f"{amalgamating_business_path}/legalName"
+        })
+
     if is_staff:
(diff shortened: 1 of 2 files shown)


meawong (approved):

legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
```diff
@@ -267,6 +267,16 @@ def _validate_foreign_businesses(  # noqa: PLR0913
         amalgamating_business,
         amalgamating_business_path) -> list:
     msg = []
+
+    # Frontend enforces these bounds on the foreign business legal name; mirror them here.
+    min_legal_name_length: Final = 3
+    max_legal_name_length: Final = 150
+    if not min_legal_name_length <= len(foreign_legal_name) <= max_legal_name_length:
+        msg.append({
+            "error": "Length of foreign business legal name must be from 3 to 150 characters.",
```
loneil, line 276: error could f-string format with the constants above rather than hardcoding 3 and 150

PR #4417, opened 2026-05-19
33425 - Auto Prune Updated Corps

Rajandeep98:
*Issue #:* /bcgov/entity#33425

*Description of changes:*

- [x] Added updated logic for creating chunks of 999
- [x] - Add function to run refresh views
- [x] - Add function to check saf criteria
- [x] - Delete from corp_processing and corp batch entries
- [x]  - Update timestamp Logic


data-tool/scripts/subset/subset_transfer_chunk.sql
```diff
@@ -32,6 +32,9 @@
 -- vset target_corp_num_predicate=target_corp_num in ('BC1111585','BC1226175');
 -- vset oracle_corp_num_predicate=c.CORP_NUM in ('1111585','1226175');
 
+-- cuttoff timestamp
+insert into public.colin_subset_extract (extracted_at)
+values (current_timestamp);
```
argush3: subset_transfer_chunk.sql is the template used to generate the subset scripts.  if the delta is large enough, there could be more than one chunk sql file.  This would mean that you could end up with multiple entries in `colin_subset_extract table` with the insert statement here.

argush3 (approved):

PR #4422, opened 2026-05-20
33174 incorp app output change

kialj876:
*Issue #:* /bcgov/entity#33174

*Description of changes:*
- updates to incorporation application output for corps
  - remove completing party
  - update completing party statement
  - decouple coop completing party template as its no longer similar to corps version
- changes feature flagged


legal-api/report-templates/template-parts/incorporation-application/completingPartyCoop.html
kialj876, line 1: No changes in here, just splitting it out of the combined 'old' version so there isn't any if legal type condition

legal-api/report-templates/template-parts/incorporation-application/completingPartyOld.html
kialj876, line 1: this is the cause of the large duplication and will be removed as part of https://github.com/bcgov/entity/issues/33493

legal-api/src/legal_api/reports/report.py
```diff
@@ -382,14 +388,38 @@ def _set_corp_flag(self, filing):
...
-        completing_party_role = PartyRole.get_party_roles_by_filing(
-            self._filing.id, datetime.utcnow(), PartyRole.RoleTypes.COMPLETING_PARTY.value)
-        if completing_party_role:
-            filing["completingParty"] = completing_party_role[0].party.json
-            with suppress(KeyError):
-                self._format_address(filing["completingParty"]["deliveryAddress"])
-            with suppress(KeyError):
-                self._format_address(filing["completingParty"]["mailingAddress"])
+        request_context = get_request_context()
+        enabled_new_features: list[str] = (flags.value("enable-new-feature",
```
kialj876, line 392: Made the flag similar to the UI in that we can reuse it for other features and have multiple separate features released dynamically without needing separate flags. I thought this would help us declutter our feature flag projects, but let me know if you disagree with that. I still like having separate flags for more permanent things, but I thought this would be good for our short term stuff

mruff-aeq: i agree. looks reasonable and useful.

loneil: Separate flags for more permanent stuff does make sense yeah. So this would be for toggles of things to integrate in but not release yet. Yes I think that's better for short term (will get removed) things.
`enabled_new_features` as the name is a minor quibble as I got a bit confused looking at that as to what it is compared to the other `enabled_` things in that file...
I can't think of a better name of the top of my head though haha, so might be fine

legal-api/flags.json
```diff
@@ -9,6 +9,7 @@
     "enable-document-records": false,
     "enable-involuntary-dissolution-filter": false,
     "enable-new-ben-statements": false,
+    "enable-new-feature": ["incorporationApplication-completingParty"],
```
loneil, line 12: `completingParty-incorporationApplication-outputs` or something _could_ make more sense to drive home the toggle is part of the Completing Party feature. Though can see that grouping by filing could be useful too

legal-api/src/legal_api/reports/report.py
loneil, line 392: Would enable-new-feature being missing have `flags.value` return a None?

Think that could cascade a breaking exception on any output load with
`incorp_compparty_stmnt_enabled = "incorporationApplication-completingParty" in enabled_new_features`

Could safeguard to be paranoid with
`enabled_new_features = flags.value("enable-new-feature", blah blah blaah) or []`

legal-api/tests/unit/reports/templates/test_completing_party.py
```diff
@@ -0,0 +1,73 @@
...
+    template = get_template(template_path)
+    rendered = template.render(
+        parties=parties,
+        header={"certifiedBy": certified_by}
+    )
+    if template_path == COMPLETING_PARTY_COOP_TEMPLATE:
+        assert "Mailing Address" in rendered
+        assert completing_party["mailingAddress"]["streetAddress"] in rendered
+        assert "Completing Party Statement" in rendered
+        assert f"I, {completing_party_name.upper()} certify that I have relevant knowledge of the Cooperative and I am authorized to make this filing."
```
loneil: Sonar warning is right I think, will always assert true (missing `in rendered`?)

legal-api/src/legal_api/reports/report.py
```diff
@@ -382,14 +388,38 @@ def _set_corp_flag(self, filing):
...
+                )
+        else:
+            completing_party_role = PartyRole.get_party_roles_by_filing(
+                self._filing.id, datetime.utcnow(), PartyRole.RoleTypes.COMPLETING_PARTY.value)
+            if completing_party_role:
+                filing["completingParty"] = completing_party_role[0].party.json
+                with suppress(KeyError):
+                    self._format_address(filing["completingParty"]["deliveryAddress"])
+                with suppress(KeyError):
+                    self._format_address(filing["completingParty"]["mailingAddress"])
```
vysakh-menon-aot, line 422: Is this applicable for the existing filings as well?

legal-api/src/legal_api/reports/report.py
kialj876: do you mean the other 'enable_' flags in ldarkly? I don't see any other 'enabled_' things in this file ?
This variable is meant to be the response from the flag which would contain all enabled features that we use the flag for so I think the name makes sense, but its only used for the line below so I don't mind changing it if you prefer something

legal-api/flags.json
kialj876: I would prefer to keep this consistent with the naming we have on the frontend flag if that's okay with you? I do agree that having completingParty first would've been a better naming convention, but I don't think it'll cause any problems as it shouldn't be there for long. And it does apply to more than just the outputs (should be referenced for the new validation, etc. as well for this feature)

legal-api/src/legal_api/reports/report.py
kialj876: yep good idea, will do

legal-api/src/legal_api/reports/report.py
kialj876: Well existing filings should have a DRS record already so they wouldn't enter this code, but you're right I should probably verify its not an existing filing getting regenerated that predates the release of the new completing party changes. 

I am thinking I'll alter this so that it checks for a completing party in the json first and still uses that for corps so that if an older filing pdf is regenerated it stays the same. This does mean we'll need to keep the original template around forever so I'll double check this in standup

legal-api/tests/unit/reports/templates/test_completing_party.py
kialj876: Oh yeah oops - will update. I'll have to change this because I think the rendered version will actually have html in this string

legal-api/src/legal_api/reports/report.py
kialj876: Pausing on this for now - PO is confirming whether or not we can rely on the previously saved pdfs for incorporation applications in DRS

kialj876: Okay so a raid ticket was opened about how we can handle regenerating outputs and I've got confirmation that they are okay with us NOT keeping the old templates around. In light of that I think its okay for this to stay as is, but I'll open a pre-release ops ticket to ensure that all existing corp incorp filings have a corresponding DRS output already.
Pre-release ops ticket: https://github.com/bcgov/entity/issues/33567

vysakh-menon-aot: Sounds good to me

vysakh-menon-aot: FYI: COLIN will still show the old data in the output as we are not making any data fixes (better to add a note in the raid ticket)

legal-api/src/legal_api/reports/report.py
kialj876: updated

legal-api/tests/unit/reports/templates/test_completing_party.py
kialj876: updated

legal-api/flags.json
kialj876: let me know what you think!

legal-api/src/legal_api/reports/report.py
kialj876: let me know if this is good to go @loneil !

legal-api/src/legal_api/reports/report.py
kialj876: I let Melissa know

legal-api/src/legal_api/reports/report.py
meawong: good idea! will definitely help prevent clutter in LD as you mentioned

meawong (approved):

vysakh-menon-aot (approved):

legal-api/flags.json
loneil: Yeah sounds good to me

legal-api/tests/unit/reports/templates/test_completing_party.py
```diff
@@ -0,0 +1,80 @@
...
+    """Completing party template renders as expected for coops and corps."""
+    certified_by = "Test Certified"
+
+    # assert incorporation data from schema is what we expect
+    parties = INCORPORATION["parties"]
+    assert parties
+    completing_party = next((party for party in parties if any(role for role in party["roles"] if role["roleType"] == "Completing Party")), None) 
+    assert completing_party
+    first_name = completing_party["officer"]["firstName"]
+    middle_name = completing_party["officer"]["lastName"]
```
loneil: middle_name set from last_name here?

kialj876: oops nice catch - just updated

loneil (approved):
Good struff!

PR #4465, opened 2026-06-08
33249 (Legal API Python 3.13 feature branch) Legal API DBC use python common

loneil:
*Issue #:* /bcgov/entity#33249 

*Description of changes:*
Migrates the Legal API's Digital Business Card implementation to consume the shared `business-registry-digital-credentials` common package rather than maintaining its own copy. 

The Traction/aca-py service, access rules, credential-data helpers, and utility code, etc are removed from legal-api and replaced by imports from the common package, so both legal-api and the queue service now operate on a single shared `DigitalCredentialsService` instance. 
The only DBC logic that remains in legal-api is the thin digital_credentials_auth wrapper, which resolves the dbc-enabled-business-types LaunchDarkly flag and forwards the result into the shared API (The FF is kept as legal-api's responsibility as it controls issuance availability, the queue service does not utilize them) 


legal-api/src/legal_api/services/digital_credentials_auth.py
```diff
@@ -12,42 +12,66 @@
...
+def _resolve_allowed_business_types() -> list[str]:
+    """Read the ``dbc-enabled-business-types`` LD flag and return the resolved list.
+
+    Returns ``[]`` if the flag is off or the flag value is malformed.
+    """
+    flags = _get_flags()
 
-    if not (user := User.find_by_jwt_token(g.jwt_oidc_token_info)):
-        return False
+    if not flags.is_on(DBC_ENABLED_BUSINESS_TYPES_FLAG):
```
kialj876: I think you should pass in the request context here and below on line 59 so that we can target people by account or user. I think it will also fix the test issues with all the 'test_get_allowed_actions' failures (the mock expects flags.value to be passing the user and account id).
i.e.
```
from legal_api.services.request_context import get_request_context
...
request_context = get_request_context()
if not flags.is_on(DBC_ENABLED_BUSINESS_TYPES_FLAG, request_context.user, request_context.account_id):
...
flag_obj = flags.value(DBC_ENABLED_BUSINESS_TYPES_FLAG, request_context.user, request_context.account_id)
```

kialj876:
Looks like the errors in test_decorators are just failing due to the traction stuff being moved out of the api - can probably just remove the traction tests as that should be being tested inside the common code now?

kialj876 (approved):
Nice work!

PR #4479, opened 2026-06-11
33127 - Block intentToLiquidate filing when business is already in liquidation

deetz99:
*Issue #:* /bcgov/entity#33127

*Description of changes:*
- Add BusinessBlocker.IN_LIQUIDATION to intentToLiquidate filing when business is already in liquidation

### DIFF
--- a/legal-api/pyproject.toml
+++ b/legal-api/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "legal-api"
-version = "3.0.4"
+version = "3.0.3"
 description = ""
 authors = [
     {name = "thor",email = "1042854+thorwolpert@users.noreply.github.com"}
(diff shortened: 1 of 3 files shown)


doug-lovett (approved):

legal-api/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "legal-api"
-version = "3.0.3"
+version = "3.0.4"
```
kialj876: Can you revert this actually? We update the version elsewhere for now and we do it once per release, but I'm working on merging the upgraded api and after that's in we'll start updating it here normally

argush3 (approved):

kialj876 (approved):

PR #4482, opened 2026-06-12
33690 - return staff-uploaded Continuation Out documents in filing documents

mruff-aeq:
*Issue #:* /bcgov/entity#33690

*Description of changes:*

Persist the supporting documents uploaded with a Continuation Out filing and surface them on the filing's documents endpoint so the dashboard ledger can render download links (staff-only, matching the existing staticDocuments gate).
*Description of changes:*
- business-filer: create_uploaded_documents() creates Document rows from continuationOut.documents and records them in filing meta_data.
- legal-api: get_static_documents() returns a continuationOut branch.
- models: add DocumentType.CONTINUATION_OUT.
- tests: filer + documents endpoint coverage.

*Testing*
[RUN_FILER_LOCALLY_&_CALL_API.md](https://github.com/user-attachments/files/28896455/RUN_FILER_LOCALLY_._CALL_API.md) - copy and paste scripts
`run_filer.py` runs the business-filer's process_filing() directly against the local database 
The run_filer.py script proves the write path (the filer persists the uploaded Continuation Out documents as rows plus meta_data in the database)
<pre>
  Query 1 — the documents rows:
  1092|continuation_out|demo-key-0001|Continuation Out Support 1.pdf
  1093|continuation_out|demo-key-0002|Continuation Out Support 2.pdf
  (Two rows = the two uploaded files. type is continuation_out; the ids are auto-assigned so they'll differ per run.)

  Query 2 — the filing's meta_data.continuationOut.uploadedDocuments:
  [{"fileKey": "demo-key-0001", "fileName": "Continuation Out Support 1.pdf"}, {"fileKey": "demo-key-0002", "fileName": "Continuation Out Support 2.pdf"}]
</pre>

The curl to `http://127.0.0.1:5000/api/v2/businesses/BC0791952/filings/3080798/documents` proves the read path (the API returns those documents, staff-gated, in the shape the dashboard ledger renders as download links).
<pre>
  curl -sS "http://127.0.0.1:5000/api/v2/businesses/BC0791952/filings/3080798/documents" \
    -H "Authorization: Bearer $TOKEN" \
    -H "Accept: application/json" \
    -w $'\nHTTP %{http_code}\n'

  The result (staff token)

  {
    "documents": {
      "receipt": "/api/v2/businesses/BC0791952/filings/3080798/documents/receipt",
      "staticDocuments": [
        {
          "name": "BC0791952 - Continuation Out Filer Doc 1.pdf",
          "url": "/api/v2/businesses/BC0791952/filings/3080798/documents/static/33690-FILER-aaaa-0001"
        },
        {
          "name": "BC0791952 - Continuation Out Filer Doc 2.pdf",
          "url": "/api/v2/businesses/BC0791952/filings/3080798/documents/static/33690-FILER-aaaa-0002"
        }
      ]
    }
  }
  HTTP 200

  For contrast, the same path with a non-staff token

  {
    "documents": {
      "receipt": "/api/v2/businesses/BC0791952/filings/3080798/documents/receipt"
    }
  }
  HTTP 200
</pre>
[run_filer.py](https://github.com/user-attachments/files/28897484/run_filer.py)

### DIFF
--- a/legal-api/src/legal_api/core/meta/filing.py
+++ b/legal-api/src/legal_api/core/meta/filing.py
@@ -1079,6 +1079,14 @@ class FilingMeta:  # pylint: disable=too-few-public-methods
                         "name": file.get("fileName"),
                         "url": f"{url_prefix}/{file_key}"
                     })
+        elif filing.filing_type == "continuationOut":
+            continuation_out = filing.meta_data.get("continuationOut", {})
+            for file in continuation_out.get("uploadedDocuments", []):
+                if file_key := file.get("fileKey"):
+                    outputs.append({
+                        "name": file.get("fileName"),
+                        "url": f"{url_prefix}/{file_key}"
+                    })
         return outputs
 
     @staticmethod
--- a/queue_services/business-filer/src/business_filer/filing_processors/continuation_out.py
+++ b/queue_services/business-filer/src/business_filer/filing_processors/continuation_out.py
@@ -35,11 +35,46 @@
 from contextlib import suppress
 
 import dpath
-from business_model.models import Business, Comment, Filing
+from business_model.models import Business, Comment, Document, DocumentType, Filing
 
 from business_filer.common.legislation_datetime import LegislationDatetime
 from business_filer.filing_meta import FilingMeta
 from business_filer.filing_processors.filing_components import filings
+
+# DocumentType.CONTINUATION_OUT is added to business-registry-model in this change, but the
+# filer installs that package from bcgov/lear@main, which won't have it until this lands.
+# Use the enum once it's available, otherwise fall back to the literal value (both are
+# "continuation_out"). This can drop to DocumentType.CONTINUATION_OUT.value once the model bumps.
+CONTINUATION_OUT_DOCUMENT_TYPE = (
+    DocumentType.CONTINUATION_OUT.value
+    if getattr(DocumentType, "CONTINUATION_OUT", None)
+    else "continuation_out"
+)
+
+
+def create_uploaded_documents(continuation_out: dict,
+                              business: Business,
+                              filing: Filing,
+                              filing_meta: FilingMeta):
+    """Create document records for the supporting documents uploaded with the continuation out filing."""
+    files = []
+    for file in continuation_out.get("documents", []):
+        document = Document()
+        document.type = CONTINUATION_OUT_DOCUMENT_TYPE
+        document.file_key = file.get("fileKey")
+        document.file_name = file.get("fileName")
+        document.filing_id = filing.id
+        business.documents.append(document)
+        files.append({
+            "fileKey": document.file_key,
+            "fileName": document.file_name
+        })
+
+    if files:
+        filing_meta.continuation_out = {
+            **filing_meta.continuation_out,
+            "uploadedDocuments": files,
+        }
 
 
 def process(business: Business, continuation_out_filing: Filing, filing: dict, filing_meta: FilingMeta):
@@ -85,3 +120,5 @@ def process(business: Business, continuation_out_filing: Filing, filing: dict, f
                 staff_id=continuation_out_filing.submitter_id
             )
         )
+
+    create_uploaded_documents(continuation_out_json, business, continuation_out_filing, filing_meta)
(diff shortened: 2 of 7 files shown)


legal-api/src/legal_api/core/meta/filing.py
```diff
@@ -1079,6 +1079,14 @@ def get_static_documents(filing, url_prefix):
                         "name": file.get("fileName"),
                         "url": f"{url_prefix}/{file_key}"
                     })
+        elif filing.filing_type == "continuationOut":
+            continuation_out = filing.meta_data.get("continuationOut", {})
+            for file in continuation_out.get("uploadedDocuments", []):
```
kialj876, line 1084: I think this needs some validation too right? For staff only?

queue_services/business-filer/src/business_filer/filing_processors/continuation_out.py
```diff
@@ -35,12 +35,47 @@
...
+        business.documents.append(document)
+        files.append({
+            "fileKey": document.file_key,
+            "fileName": document.file_name
+        })
+
+    if files:
+        filing_meta.continuation_out = {
+            **filing_meta.continuation_out,
+            "uploadedDocuments": files,
```
kialj876, line 76: Where do we upload these files to DRS? Does that happen in an existing process already?

queue_services/business-filer/src/business_filer/filing_processors/continuation_out.py
```diff
@@ -35,12 +35,47 @@
...
+        document = Document()
+        document.type = CONTINUATION_OUT_DOCUMENT_TYPE
+        document.file_key = file.get("fileKey")
+        document.file_name = file.get("fileName")
+        document.filing_id = filing.id
+        business.documents.append(document)
+        files.append({
+            "fileKey": document.file_key,
+            "fileName": document.file_name
+        })
```
kialj876, line 71: Not needed for this PR, but would be better if this was in a util (we've got the same sort of thing copy/pasted in a few places in the filer)

kialj876 (commented):
Looking good, just a couple minor comments/questions for ya

legal-api/src/legal_api/core/meta/filing.py
mruff-aeq: It is actually already staff-only. It inherits the same access controls as the existing `continuationIn` docs.
> Here is how the gating works under the hood:
> * **Data Path:** `get_static_documents` doesn't need its own auth. It’s only called by `core/filing.py`, which already enforces the `staff` role before including `staticDocuments` in the response.
> * **Download Path:** The `GET .../documents/static/<file_key>` endpoint runs `_is_document_available`, which triggers the same staff check. Non-staff users will get a 404 before any bytes are sent.
> 
> 
 I did add new tests to prove the gate works—verifying that basic and public users only get the receipt (no static docs), while staff can access them.

queue_services/business-filer/src/business_filer/filing_processors/continuation_out.py
mruff-aeq: To clarify, these aren't routed through DRS. They are uploaded straight to MinIO via a pre-existing presigned-URL flow.

- Frontend (Already handled): The FileUploadPdf component gets a presigned URL and PUTs the bytes directly into object storage. This is the exact same component and flow we’ve used for court orders, etc. It simply attaches the [{fileKey, fileName}] to the filing payload.
- 
- Backend (This PR): By the time the filer runs, the file is already in MinIO. This PR is purely the backend half—it takes those existing fileKeys, creates the Document database rows with the correct metadata, and surfaces them for staff via the API.

queue_services/business-filer/src/business_filer/filing_processors/continuation_out.py
mruff-aeq: Agreed, not for this PR.  I'll create a follow-up ticket to move this into a shared util.

queue_services/business-filer/src/business_filer/filing_processors/continuation_out.py
kialj876: Ah okay thanks for the explanation, I didn't realize we still relied on minio for this. We'll need to rework this eventually to work with DRS

legal-api/src/legal_api/core/meta/filing.py
kialj876: okay perfect!

kialj876 (approved):
Nice work!

PR #4496, opened 2026-06-15
33515 updating the old COOPS outputs to the modern templates

vysakh-menon-aot:
*Issue #:* /bcgov/entity#33515

*Description of changes:*
image

image

image


legal-api/report-templates/template-parts/addresses.html
kialj876, line 1: It looks like this is the exact same as the bc address layout now? Can we combine cp / bc (add the records if clause here and update bc to use this template / remove 'bc-address-change' folder?

legal-api/report-templates/template-parts/directors.html
kialj876, line 1: same thing for this? It looks like the only differences could be applied to the BC version as well?

legal-api/report-templates/annualReport.html
kialj876, line 1: And this too up until line 46?

legal-api/report-templates/changeOfAddress.html
kialj876, line 1: And here - exact same as bcAddressChange

legal-api/report-templates/changeOfDirectors.html
kialj876, line 1: and same here with bcDirectorChange

kialj876 (commented):
Hey Vysakh, can we condense the templates? It looks like BC and CP are almost the exact same now. All my comments below are to do with that

doug-lovett (approved):

legal-api/report-templates/template-parts/addresses.html
vysakh-menon-aot: That was my original plan. Waiting for few UX confirmation regarding the alignment of corps and coop design

vysakh-menon-aot: updated to use common template for coop and corps

kialj876 (approved):
Thanks for the updates!

PR #4515, opened 2026-06-18
32957 Business Emailer - Incorp email updates

kialj876:
*Issue #:* /bcgov/entity#32957

*Description of changes:*
I have met with Janis and explained the limitations on the markdown rendering for GC Notify (this is what renders our emails when we don't provide any html in the template - cheaper/faster/SRE wants us to use it) so the result here is not the exact same as the figma ux (we can't align the bullets or right align the text as it shows in there)

Examples:
future effective paid:
[future-effective-paid-email.pdf](https://github.com/user-attachments/files/29099013/future-effective-paid-email.pdf)

regular completed:
[regular-completed.pdf](https://github.com/user-attachments/files/29099030/regular-completed.pdf)


queue_services/business-emailer/src/business_emailer/email_processors/filing_notification.py
```diff
@@ -41,194 +42,230 @@
...
+            if not business_name:
+                # assume it is numbered for this case (following current logic)
+                numbered_description = Business.BUSINESSES.get(legal_type, {}).get("numberedDescription")
+                business_name = numbered_description
+
+    else:
+        template = Path(
+            f'{current_app.config.get("TEMPLATE_PATH")}/BC-{FILING_TYPE_CONVERTER[filing_type]}-{status}.html'
+        ).read_text()
+        filled_template = substitute_template_parts(template, "md")
```
vysakh-menon-aot: this should be html instead of md?

kialj876: ah yes it should - this should be failing a test. Will dig and update

kialj876: Okay updated and added a simple test for the other filings in this flow

meawong (approved):

vysakh-menon-aot (approved):

PR #4534, opened 2026-06-24
33936 legal-api, business filer, business emailer - update versions

kialj876:
*Issue #:* /bcgov/entity#33936

*Description of changes:*


legal-api/src/legal_api/version.py
kialj876, line 1: Version now goes off of toml file so this is no longer used. Toml file version was already updated during the api upgrade

vysakh-menon-aot (approved):

PR #4542, opened 2026-06-29
32958 Business Emailer - updates for reg and cor

kialj876:
*Issue #:* /bcgov/entity#32958

*Description of changes:*
- moved registration and change of registration into the filing_notification branch and updated their templates to the new markdown design


queue_services/business-emailer/src/business_emailer/email_processors/__init__.py
```diff
@@ -48,26 +48,36 @@ def get_filing_info(filing_id: str) -> tuple[Filing, dict, dict, str, str]:
     return filing, business_json, leg_tmz_filing_date, leg_tmz_effective_date
 
 
+def get_party_emails(parties: list[dict], roles: list[str]) -> list[str]:
+    """Return the emails for the specified party types."""
+    recipients = []
+    for party in parties:
+        for role in party["roles"]:
+            if role["roleType"] in roles:
+                recipients.append(party["officer"].get("email"))
```
mruff-aeq: get_party_emails appends None when a matched party has no email, which later crashes get_recipients at ', '.join(party_emails) (line 76) with TypeError: sequence item 0: expected str instance, NoneType found. filter it out, e.g. return [email for email in recipients if email].

kialj876: nice catch, thanks! Will update

vysakh-menon-aot (approved):

queue_services/business-emailer/src/business_emailer/email_processors/filing_notification.py
```diff
@@ -145,13 +148,15 @@ def process(email_info: dict, token: str) -> dict:
...
         business_name=business_name,
         business_identifier=business_identifier,
         business_number=business_number,
         filing_name=filing_name,
         filing_name_short=filing_name_short,
         future_attachments_list=future_attachments,
         office_name=OFFICE_NAME.get(legal_type_key),
+        number_description="Registration" if legal_type_key == "FIRM" else "Incorporation",
         show_effective_date=show_effective_date,
     )
```
mruff-aeq, line 164: get_recipients is called without filing_type here, so registration/changeOfRegistration fall through to the get_recipient_from_auth else branch and the completing-party/partner/proprietor/contactPoint emails the old processors sent to are dropped. 

is that intentional, or should filing_type be passed?

kialj876: another good catch, thanks!

queue_services/business-emailer/src/business_emailer/email_processors/__init__.py
```diff
@@ -48,26 +48,36 @@ def get_filing_info(filing_id: str) -> tuple[Filing, dict, dict, str, str]:
...
-            for party in parties:
-                for role in party["roles"]:
-                    if role["roleType"] == "Completing Party" and \
-                        (comp_party_email := party["officer"].get("email")):
-                        recipients = f"{recipients}, {comp_party_email}"
-                        break
+        # add filing contact point email
+        recipients = filing_json["filing"][filing_type].get("contactPoint", {}).get("email", "") or ""
+        # add relevant party emails
+        if (option != "mras"
```
mruff-aeq: Changing the guard from option in [PAID, "bn"] to option != "mras" means COMPLETED IA/amalgamation/continuationIn filings now CC the Completing Party (and Partner/Proprietor) where they previously didn't. 

Is that intended for those shared callers, or should the scope be limited to reg/cor?

queue_services/business-emailer/src/business_emailer/email_processors/filing_notification.py
kialj876: updated this and tests now secure it

queue_services/business-emailer/src/business_emailer/email_processors/__init__.py
kialj876: Yeah thank you for calling this out, I reworked this. When I originally combined the logic here for the different filings I got a little confused, but I think what I have now is right and makes the intention more clear without needing check everywhere its called.
Note: correction and dissolution are referenced here, but they do not use this flow yet. I just added them to get a good idea of how this logic would look once they are added. Currently this would change the correction when correcting a dissolution to also include the Custodian to the recipient list (I think this is desirable, but I will verify with design)

queue_services/business-emailer/src/business_emailer/email_processors/__init__.py
```diff
@@ -48,27 +48,42 @@ def get_filing_info(filing_id: str) -> tuple[Filing, dict, dict, str, str]:
...
-                    if role["roleType"] == "Completing Party" and \
-                        (comp_party_email := party["officer"].get("email")):
-                        recipients = f"{recipients}, {comp_party_email}"
-                        break
-    else:
-        identifier = filing_json["filing"]["business"]["identifier"]
-        if identifier[:2] != "CP":
-            # only add recipients if not coop
-            recipients = get_recipient_from_auth(identifier, token)
+    identifier = filing_json["filing"]["business"]["identifier"]
```
mruff-aeq, line 66: get_recipients reads filing_json["filing"]["business"]["identifier"] directly (line 66), but new business filings like incorporation and registration don't have a business block yet (your own prep_registration_filing deletes it for PAID), so this'll throw a KeyError on something like a future-effective incorporation 

can we guard it with .get("business", {})?

kialj876: This is actually required by the schema so it will always be there. In the case of a bootstrap filing (incorp/amalg/cont in/registration) this would be a temporary identifier (regardless if its future effective or not). I will update the tests so that its realistic / more clear

kialj876: Okay I adjusted a couple things to make it more clear about the temp reg and made the tests setup more realistic

mruff-aeq (approved):

PR #4553, opened 2026-07-03
32956 email update amalg contin

kialj876:
*Issue #:* /bcgov/entity#32956

*Description of changes:*
- updated email flows for amalg and continuation in
- condensed tests with incorp (similar flow and setup) 


queue_services/business-emailer/src/business_emailer/email_processors/filing_notification.py
```diff
@@ -74,16 +79,30 @@ def _get_attachments_and_extra_pdf_types(status: str, filing_type: str, filing:
...
+    return invalid_status or invalid_data or invalid_coop_filing
+
+
 def process(email_info: dict, token: str) -> dict | None:
     """Build the email the filing notification."""
     current_app.logger.debug("filing_notification: %s", email_info)
     filing_type, status = email_info["type"], email_info["option"]
+
+    regenerate = False
+    # TODO: need to confirm this case for continuationIn
```
kialj876: Okay so after checking with design it seems I will need to keep some special statuses like this in the old flow - continuationIn has some special review stages before it gets filed

queue_services/business-emailer/src/business_emailer/email_processors/continuation_in_notification.py
```diff
@@ -34,72 +30,21 @@
...
-                    "attachOrder": str(attach_order)
-                }
-            )
-            attach_order += 1
-
-    elif status == "RESUBMITTED":
-        # add filing pdf
-        filing_pdf_type = "continuationIn"
-        filing_pdf_encoded = get_filing_document(business["identifier"], filing.id, filing_pdf_type, token)
+        filing_pdf_encoded = get_filing_document(business["identifier"], filing.id, filing_pdf_type, token, True)
```
mruff-aeq, line 47: This RESUBMITTED path now passes regenerate=True, so the request hits legal-api's _should_regenerate() branch, which returns 401 "Unauthorized to regenerate" unless the caller has the staff/system role. Does the emailer's service token carry that role? If not, the resubmitted email would silently drop the Continuation Application PDF (401 → None → attachment skipped). Pre-change this path fetched the stored PDF with no regenerate.

queue_services/business-emailer/src/business_emailer/email_processors/continuation_in_notification.py
```diff
@@ -186,7 +103,7 @@ def process(email_info: dict, token: str) -> dict:  # pylint: disable=too-many-l
         header=(filing.json)["filing"]["header"],
         filing_date_time=leg_tmz_filing_date,
         effective_date_time=leg_tmz_effective_date,
-        entity_dashboard_url=get_entity_dashboard_url(business.get("identifier"), token),
+        entity_dashboard_url=current_app.config.get("DASHBOARD_URL") + business["identifier"],
```
mruff-aeq, line 106: In continuation_in_notification.py, entity_dashboard_url was changed from get_entity_dashboard_url(...) to DASHBOARD_URL + business["identifier"]. In the review flow business["identifier"] is the temp reg (T...), and the old helper resolved temp identifiers to the auth-web account/{org_id}/business page. The APPROVED/CHANGE_REQUESTED/REJECTED templates render this as the "My Business Registry" link — should that still point to the account page for temp identifiers, or is DASHBOARD_URL + T... intended here?

queue_services/business-emailer/src/business_emailer/email_processors/continuation_in_notification.py
kialj876: yes it has the system role. All our backend services that don't involve user requests generally use system tokens for talking to other services internally (AccountService.get_bearer_token uses system token creds)

queue_services/business-emailer/src/business_emailer/email_processors/continuation_in_notification.py
kialj876: yes this is intended

mruff-aeq (approved):

meawong (approved):
Nice!

PR #4558, opened 2026-07-08
34021 - Use Legislation Datetime for Special Resolution Date Comparisons

meawong:
*Issue #:* /bcgov/entity#34021

*Description of changes:*
Special Resolution validation was comparing the resolution date against the business incorporation date using the raw UTC timestamp. Updated the Special Resolution validation logic to convert timestamps into the legislation timezone before comparing legal dates:
- Aligned with other implementations, see: https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/filings/validations/common_validations.py#L143


kialj876 (approved):
Nice!

PR #4570, opened 2026-07-14
33940 api completing party incorporation

mruff-aeq:
lets *Issue #:* /bcgov/entity#33940

*Description of changes:*
Allows authorized API users to specify the completing party name on corp incorporation applications via header certifiedBy

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/common_validations.py
+++ b/legal-api/src/legal_api/services/filings/validations/common_validations.py
@@ -33,7 +33,7 @@ from business_common.utils.legislation_datetime import LegislationDatetime
 from business_model.models import Address, Business, PartyRole
 from legal_api.core.filing import Filing as CoreFiling
 from legal_api.errors import Error
-from legal_api.services import STAFF_ROLE, SYSTEM_ROLE, MinioService, colin, flags, namex
+from legal_api.services import STAFF_ROLE, MinioService, colin, flags, namex
 from legal_api.services.permissions import ListActionsPermissionsAllowed, PermissionService
 from legal_api.services.request_context import get_request_context
 from legal_api.services.utils import get_str
@@ -1299,14 +1299,14 @@ def validate_certified_by(filing_json: dict, filing_type: str, legal_type: str) 
     certified_by = filing_json["filing"]["header"].get("certifiedBy")
 
     if legal_type in Business.CORPS:
-        # completing party statement takes its name from certifiedBy for staff and API users
-        enabled_features: list[str] = flags.value("enable-new-feature", []) or []
+        # the completing party statement takes its name from certifiedBy for staff and API users,
+        # so certifiedBy is required for them on a corp incorporation application
+        api_login_source = "API_GW"  # jwt loginSource of an API gateway user
         current_user = getattr(request_ctx, "current_user", None) if has_request_context() else None
         if (filing_type == CoreFiling.FilingTypes.INCORPORATIONAPPLICATION
-                and "incorporationApplication-completingParty" in enabled_features
                 and current_user
                 and (jwt.validate_roles(current_user, [STAFF_ROLE])
-                     or jwt.validate_roles(current_user, [SYSTEM_ROLE]))):
+                     or current_user.get("loginSource") == api_login_source)):
             if not certified_by:
                 msg.append({
                     "error": "Certified by field is required.",
--- a/legal-api/src/legal_api/services/filings/validations/common_validations.py
+++ b/legal-api/src/legal_api/services/filings/validations/common_validations.py
@@ -21,7 +21,8 @@ from http import HTTPStatus
 from typing import Final
 
 import pycountry
-from flask import current_app, g, request
+from flask import current_app, g, has_request_context, request
+from flask.globals import request_ctx
 from flask_babel import _
 from pypdf import PdfReader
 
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/services/filings/validations/incorporation_application.py
```diff
@@ -329,16 +329,20 @@ def _validate_incorporation_permission(
     """Validate the permission and completing party of the incorporation filing."""
     if not flags.is_on("enabled-deeper-permission-action"):
         return None
-    
+
+    check_name = False
     if _incorp_completing_party_not_required() and legal_type in Business.CORPS:
-        return None
+        if not has_completing_party(incorporation_json, filing_type):
```
kialj876: I don't think this is right - the api users should be allowed to specify a completing party name the same as staff NOT add a party record

kialj876 (changes requested):
I think there may be some misunderstanding here. This ticket is to allow api users to specify a completing party name (the same way as staff do already), not to add a party record which we just took out of this filing entirely. You will need to check the jwt or User record for the login source to know if they are an api user or not, and then I believe you can just take the certifiedBy in the header of the filing (validation and report generation should be looked at)

legal-api/src/legal_api/services/filings/validations/incorporation_application.py
mruff-aeq: thanks for clarifying, what you are describing is different than whats written on the ticket, revised ticket

legal-api/src/legal_api/services/filings/validations/common_validations.py
```diff
@@ -1298,7 +1299,25 @@ def validate_certified_by(filing_json: dict, filing_type: str, legal_type: str)
     certified_by = filing_json["filing"]["header"].get("certifiedBy")
 
     if legal_type in Business.CORPS:
-        return msg  # certifiedBy is not required for corporations
+        # completing party statement takes its name from certifiedBy for staff and API users
+        enabled_features: list[str] = flags.value("enable-new-feature", []) or []
+        current_user = getattr(request_ctx, "current_user", None) if has_request_context() else None
+        if (filing_type == CoreFiling.FilingTypes.INCORPORATIONAPPLICATION
+                and "incorporationApplication-completingParty" in enabled_features
```
kialj876: you can remove this check as its permanently enabled now

legal-api/src/legal_api/services/filings/validations/common_validations.py
```diff
@@ -1298,7 +1299,25 @@ def validate_certified_by(filing_json: dict, filing_type: str, legal_type: str)
...
+                and current_user
+                and (jwt.validate_roles(current_user, [STAFF_ROLE])
+                     or jwt.validate_roles(current_user, [SYSTEM_ROLE]))):
+            if not certified_by:
+                msg.append({
+                    "error": "Certified by field is required.",
+                    "path": "/filing/header/certifiedBy"
```
kialj876, line 1313: are you sure this is not set by the UI for regular users? It may be set for regular users and ignored in the api report unless its staff

legal-api/src/legal_api/services/filings/validations/common_validations.py
```diff
@@ -1298,7 +1299,25 @@ def validate_certified_by(filing_json: dict, filing_type: str, legal_type: str)
...
     if legal_type in Business.CORPS:
-        return msg  # certifiedBy is not required for corporations
+        # completing party statement takes its name from certifiedBy for staff and API users
+        enabled_features: list[str] = flags.value("enable-new-feature", []) or []
+        current_user = getattr(request_ctx, "current_user", None) if has_request_context() else None
+        if (filing_type == CoreFiling.FilingTypes.INCORPORATIONAPPLICATION
+                and "incorporationApplication-completingParty" in enabled_features
+                and current_user
+                and (jwt.validate_roles(current_user, [STAFF_ROLE])
+                     or jwt.validate_roles(current_user, [SYSTEM_ROLE]))):
```
kialj876: this is not right - the system role is for internal BE services. You need to check the login source for an api user

kialj876 (changes requested):
On the right track here but some changes needed below

legal-api/src/legal_api/services/filings/validations/common_validations.py
mruff-aeq: On the API user check: I couldn't find an existing API-user concept in legal-api. authz.py only defines role constants (system, staff, sbc_staff, contact_centre_staff, maximus_staff, basic, colin, public_user, account_identity), and there's no loginSource check anywhere in legal-api/src. The only precedent I found is in search-api (search_api/services/authz.py), which compares the token's loginSource claim to API_GW (config JWT_OIDC_LOGIN_SOURCE / JWT_OIDC_API_GW). Is that the check you want here, and should I add the same config + a helper in legal_api/services/authz.py, or is there something existing I've missed?

kialj876: yeah its not a role its a different field in the jwt. It is saved to the User table so you can look at the column 'login_source' there. I'm not sure that logic in search will be useful here, I would just check the User column. You can ask SRE or probably Doug about what the value will be for an api user (its set by the jwt loginSource)

legal-api/src/legal_api/services/filings/validations/common_validations.py
mruff-aeq: Done — removed the `enable-new-feature` flag gate. The check now runs unconditionally for corp incorporation applications since the feature is permanently enabled.

legal-api/src/legal_api/services/filings/validations/common_validations.py
mruff-aeq: Good catch — the gate now only matches staff and API users (API users via the jwt loginSource), so certifiedBy is never required for regular users even if the UI sets it. Regular users continue to have it sourced from the submitter name in the report.

legal-api/src/legal_api/services/filings/validations/common_validations.py
mruff-aeq: Talked to Doug about the API-user completing party. His guidance:

for api requests if you use the token to obtain completing party information it will always be the same - the auth account name, not a user name. You can use the loginSource to determine if it is an API request.

doug-lovett (approved):
Looks okay to me.

kialj876 (approved):
Looks good!

PR #4610, opened 2026-07-24
34200 consent continuation out output

mruff-aeq:
*Issue #:* /bcgov/entity#34200

*Description of changes:*
Generate a Continue Out Application output for staff consent to continuation out filings and show it in the filing history.


legal-api/src/legal_api/core/filing.py
kialj876: should it be removed here too?

mruff-aeq:
image

legal-api/src/legal_api/reports/report.py
```diff
@@ -1964,6 +1964,11 @@ class ReportMeta:  # pylint: disable=too-few-public-methods
             "fileName": "letterOfConsentAmalgamationOut",
             "reportType": ReportTypes.FILING_2.value  # change to filing-3 to omit stamp
         },
+        "consentContinuationOut": {
+            "filingDescription": "Continue Out Application",
+            "fileName": "continueOutApplication",
```
kialj876: can this be "consentContinuationOut"? The convention everywhere else in here is also keeping the 'fileName' the same as the key

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -582,6 +585,11 @@ def _update_additional_filing(self, doc_list: dict, drs_params: str) -> dict:
                                 not str(output).startswith("certificate") and
                                 doc_list.get(output)
                             ):
+                                # A DRS record only maps to an output configured with the same
+                                # report type (e.g. skip legacy FILING-2 letters of consent).
+                                output_report_type = ReportMeta.reports.get(output, {}).get("reportType")
+                                if report_type and output_report_type and output_report_type != report_type:
```
kialj876, line 591: Is this needed for this output or are you adding a new guard? Is this due to the "fileName" not matching the report key?

kialj876 (commented):
Looking good, I just have a few questions below

legal-api/src/legal_api/core/filing.py
mruff-aeq: Good catch, removed it from this list too so the output also shows while the filing is paid.

legal-api/src/legal_api/reports/report.py
mruff-aeq: Done, renamed the fileName and template file to consentContinuationOut.

legal-api/src/legal_api/reports/document_service.py
mruff-aeq: It's needed — legacy consent filings still hold the old letter as FILING-2 in the DRS (from before it moved to FILING-3), and without this check those records also append their params onto the letterOfConsent url.

kialj876 (approved):
Nice work!

PR #4618, opened 2026-07-27
34180 - Update Filer from Minio to Drs

meawong:
*Issue #:* /bcgov/entity#34180

*Description of changes:*

### DIFF
--- a/queue_services/business-filer/src/business_filer/filing_processors/filing_components/document_records.py
+++ b/queue_services/business-filer/src/business_filer/filing_processors/filing_components/document_records.py
@@ -23,6 +23,7 @@ from business_filer.services import Flags, document_service
 
 DRS_FEATURE_FLAG = "enable-new-feature"
 DRS_FEATURE_VARIATION = "drs-upload"
+_DRS_KEY_PATTERN = re.compile(r"^[A-Z]+-DS\d+$")
 
 
 def _is_drs_upload_enabled() -> bool:
@@ -83,9 +84,6 @@ def _build_update_info(business: Business, filing: Filing) -> dict:
     if business:
         update_info["businessIdentifier"] = business.identifier
     return {k: v for k, v in update_info.items() if v is not None}
-
-
-_DRS_KEY_PATTERN = re.compile(r"^[A-Z]+-DS\d+$")
 
 
 def _is_drs_document(file_key: str) -> bool:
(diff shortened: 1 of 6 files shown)


queue_services/business-filer/src/business_filer/filing_processors/filing_components/document_records.py
```diff
@@ -0,0 +1,95 @@
...
+
+    documents = _find_documents_for_filing(filing.id)
+
+    if not documents:
+        return
+
+    update_info = _build_update_info(business, filing)
+
+    for document in documents:
+        if not _is_drs_document(document.file_key):
```
kialj876, line 42: Hmm since we have this check I am wondering if we should remove the feature flag check in this part? Sorry I know we refined it this way but I think if:
- someone filed with the flag on, 
- then we turned it off, 
- then it was processed <- we'd want it to update DRS still since the filing is using it

kialj876 (commented):
Looks good to me, but one comment for you to consider below. Maybe we don't need the feature flag check in the filer? Potentially the API should be requiring a DRS key when the flag is on?

queue_services/business-filer/src/business_filer/filing_processors/filing_components/document_records.py
meawong: good point! I'll update this

doug-lovett (approved):
Looks great.

queue_services/business-filer/src/business_filer/services/filer.py
```diff
@@ -326,6 +326,10 @@ def process_filing(filing_message: FilingMessage): # noqa: PLR0915, PLR0912
         if not Flags.is_on("enable-sandbox"):
             PublishEvent.publish_email_message(current_app, business, filing_submission, filing_submission.status)
 
+        # Update the document record(s) for any client-submitted documents on this filing with
+        # the filing id, filing date, and business identifier
+        document_records.update_document_records(business, filing_submission)
```
vysakh-menon-aot, line 331: Why are we doing this in business-filer instead of pushing message into drs and process in the drs queue? similar to below code (publish_drs_create_message). update_document_records is called after completing transaction, if this failed due to any reason there is no retry (similar to namex consume and business profile update_entity, supposed to be moved into a separate queue). @kialj876 @meawong @doug-lovett

vysakh-menon-aot: If this is not feasible then we need to introduce all actions after transaction as a sub tasks of business-filer and implement retry

doug-lovett: Nice catch, using pub/sub for this DRS record update is a better solution. I will set up a new topic named doc-api-app-update-record. The filer can then switch to publishing on this topic in a separate. PR

mruff-aeq: i made this into a ticket: https://app.zenhub.com/workspaces/entities-team-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/34401

PR #4627, opened 2026-07-28
34300 - Serve uploaded documents from DRS and stamp affidavit on download

mruff-aeq:
Issue #: bcgov/entity#34300

## Description

- `Report._get_static_report` now serves uploaded client documents from the DRS when `documents.file_key` is a DRS key (`{documentClass}-{documentServiceId}`, e.g. `COOP-DS0000101951`). Legacy Minio UUID keys and pre-existing bare `DS...` keys are unchanged and continue to be served from Minio.
- The registrar's certification stamp is applied to the coop dissolution affidavit at serve time (both DRS and Minio backed), so the emailed "Certified Affidavit.pdf" attachment and document downloads match their name. Stored originals remain unstamped, consistent with the serve-time certification of generated reports. The emailer and ledger downloads inherit this via the documents endpoint with no changes.

## Testing

- New unit tests: DRS/Minio dispatch by key shape (3 cases) and affidavit certification for both storage backends (asserts "Filed on", identifier, and original body text on page 1).
- Verified end to end locally: affidavit document row keyed to a DRS dev document, served through `GET /api/v2/businesses/{identifier}/filings/{filing_id}/documents/affidavit` — response fetched from DRS (no Minio) with the certification stamp applied.

### DIFF
--- a/legal-api/src/legal_api/reports/report.py
+++ b/legal-api/src/legal_api/reports/report.py
@@ -13,6 +13,7 @@ import base64
 import copy
 import json
 import os
+import re
 from contextlib import suppress
 from http import HTTPStatus
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


mruff-aeq:
Certified Copy Stamp appearing on affidavit documents from dev database

Screenshot 2026-07-28 at 4 40 33 AM
image

legal-api/tests/unit/reports/test_report.py
```diff
@@ -1189,3 +1189,93 @@ def test_special_resolution_drs_report_type(session, filing_type, expected_repor
...
+    minio_response = MagicMock(data=pdf_bytes, status=HTTPStatus.OK)
+    with patch.object(doc_service, 'get_document', return_value=drs_response), \
+            patch.object(MinioService, 'get_file', return_value=minio_response):
+        response = Report(filing).get_pdf(report_type='affidavit')
+
+    stamped = PdfReader(_io.BytesIO(response.get_data()))
+    text = stamped.get_page(0).extract_text()
+    assert 'Affidavit body' in text
+    assert 'Filed on' in text
+    assert identifier in text
```
kialj876, line 1284: this isn't asserting the certified by stamp? Shouldn't it be checking for 'CERTIFIED COPY' or the registrar name?

kialj876 (approved):
Nice work, looks good to me, but please update the test if you agree

legal-api/src/legal_api/reports/report.py
```diff
@@ -76,14 +77,40 @@ def get_pdf(self, report_type=None, regenerate: bool=False):
...
+            document_data = self._certify_uploaded_document(document_data)
         return current_app.response_class(
-            response=response.data,
-            status=response.status,
+            response=document_data,
+            status=status,
             mimetype="application/pdf"
         )
 
+    def _certify_uploaded_document(self, document_bytes: bytes) -> bytes:
```
doug-lovett, line 99: There may be an inconsistency here: the DRS is already configured in PROD to add the certified copy stamp (approved by UX) for COOP client submitted documents. The DRS certified copy version may not match this version.

kialj876: it is configured this way in prod but not in dev or test?

mruff-aeq: @doug-lovett just wanted to follow up on this, on the follow up action item(s) , so prod will need to disable the existing stamp and have a unified stamp workflow across dev/test/prod , also for coop we need to render a different stamp?

doug-lovett: I do not think the stamps are different across different legal/entity types or document types. In the DRS API whether the stamp is added is determined by the configuration in an environment variable. So for a client submitted COOP rules document, for example, it is added. My understanding is that not all client documents get a certified stamp. Currently COOP-COOP_RULES, COOP-COOP_MEMORANDUM, and COOP-COSD can have a stamp added in all environments.

PR #4628, opened 2026-07-28
34400 - Serve and delete uploaded documents from DRS by file key shape

mruff-aeq:
Issue #: bcgov/entity#34400

The static document download route and draft filing deletion now dispatch to the DRS for DRS-backed file keys (`{documentClass}-{documentServiceId}`), with legacy Minio keys unchanged.

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
@@ -16,6 +16,7 @@
 Provides all the search and retrieval from the business entity datastore.
 """
 import copy
+import re
 from contextlib import suppress
 from datetime import UTC
 from datetime import datetime as _datetime
@@ -39,6 +40,7 @@ from business_common.utils.legislation_datetime import LegislationDatetime
 from business_model.models import (
     Address,
     Business,
+    Document,
     Filing,
     OfficeType,
     RegistrationBootstrap,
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
```diff
@@ -1109,9 +1112,21 @@ def is_future_effective_filing(filing_json: dict) -> bool:
...
         return is_future_effective
 
+    @staticmethod
+    def delete_uploaded_file(file_key: str):
+        """Delete an uploaded file from the DRS or Minio based on the file key shape.
+
+        DRS-backed keys are "{documentClass}-{documentServiceId}", e.g. "COOP-DS0000101951";
+        legacy Minio keys (UUIDs) do not match.
+        """
+        if re.match(r"^([A-Z]+)-(DS\d+)$", file_key):
```
kialj876, line 1122: (optional) we've got this re.match in a few places now - I think its worth making a util? The re.match or even just the regex as a constant

kialj876 (approved):
Nice work

legal-api/src/legal_api/resources/v2/business/business_filings/business_filings.py
mruff-aeq: i'll make into a ticket

PR #4629, opened 2026-07-28
34389 add court order metadata

vysakh-menon-aot:
*Issue #:* /bcgov/entity#34389

*Description of changes:*
  - Adding court order details into metadata to make sure the corrected data is not showing in the ledger (we could use court_order version table for this instead but data migration might not be able to populate this in the tombstone pipeline)
  - Adding court order files into metadata for the same reason + we need to support multiple court order documents in court order filing

### DIFF
--- a/python/common/business-registry-model/src/business_model_migrations/versions/20260727_162109_d7fc1a767d69_.py
+++ b/python/common/business-registry-model/src/business_model_migrations/versions/20260727_162109_d7fc1a767d69_.py
@@ -5,9 +5,8 @@ Revises: 1cdc6fdbf4cf
 Create Date: 2026-07-27 16:21:09.583470
 
 """
-from alembic import op
 import sqlalchemy as sa
-
+from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = 'd7fc1a767d69'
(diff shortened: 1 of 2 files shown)


kialj876 (approved):

python/common/business-registry-model/src/business_model_migrations/versions/20260727_162109_d7fc1a767d69_.py
```diff
@@ -0,0 +1,67 @@
...
+def upgrade():
+    # Add court order data to metadata of filings
+    op.execute(
+        """
+        UPDATE filings f
+        SET meta_data = JSONB_SET(meta_data, '{courtOrder}',
+                            jsonb_strip_nulls(jsonb_build_object(
+                                'fileNumber', co.file_number,
+                                'orderDetails', CASE WHEN co.order_details IS NOT NULL AND co.order_details <> '' THEN co.order_details ELSE NULL END,
+                                'effectOfOrder', CASE WHEN co.effect_of_order IS NOT NULL AND co.effect_of_order <> '' THEN co.effect_of_order ELSE NULL END
```
vysakh-menon-aot, line 27: @ketaki-deodhar @Rajandeep98 please include this in the data migration

ketaki-deodhar (approved):

PR #4641, opened 2026-07-30
32906 contact email validation

benjamin-bc-gov:
*Issue #:* /bcgov/entity#32906

*Description of changes:*
Contact email validation 

`
  - a@b.c → REJECTED ✓                                                                                                                                    
  - test@example.com → ACCEPTED ✓                                                                                                                         
  - @invalid.com → REJECTED ✓                                                                                                                             
  - test@domain (no TLD) → REJECTED ✓                                                                                                                     
  -  test@example.com (leading space) → REJECTED ✓                                                                                                        
  - test@example.com  (trailing space) → REJECTED ✓                                                                                                       
  - user@bc.gov.ca → ACCEPTED ✓                                                                                                                           
                                                                                                                                                          
  All cases match the frontend behaviour exactly. 

  - Alteration                                                                                                                                  
  - Amalgamation Application                                                                                                                    
  - Change of Registration                                                                                                                      
  - Continuation In                                                                                                                             
  - Correction                                                                                                                                  
  - Registration                                                                                                                                
  - Restoration
`


kialj876 (commented):
This should also effect incorporation application I think?

meawong:
> This should also effect incorporation application I think?

yes, good catch! I've updated the ticket requirements

kialj876:
Also, did you verify this is not already handled by the schema? I think this is already being validated so we don't need to revalidate it in the API
https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/contact_point.json#L13

benjamin-bc-gov:
> This should also effect incorporation application I think?

checking

kialj876:
It looks like this is all already being validated by the schema so can you please close this and update it there if there's any change to be made? You can verify by adding to this test case: https://github.com/bcgov/business-schemas/blob/main/tests/unit/schemas/test_required_non_empty_strings.py#L395

benjamin-bc-gov:
> It looks like this is all already being validated by the schema so can you please close this and update it there if there's any change to be made? You can verify by adding to this test case: https://github.com/bcgov/business-schemas/blob/main/tests/unit/schemas/test_required_non_empty_strings.py#L395

you are right they are both duplicate validation, but below are some of the patterns that will scale through both 

 test@-example.com          # hyphen at start of domain label           
 test@example-.com          # hyphen at end of domain label             
 test@[999.999.999.999]     # IP octet out of range (0-255)             
 "hello world"@example.com  # quoted local part allows anything inside  quotes                                                                 
 tëst@example.com           # unicode characters in local part          
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa@example.com  # no length limit    


should we update the schema to accommodate that ? or i should change my code to accommodate it.

benjamin-bc-gov:
this PR for business schema need to go in first before my test can pass 

https://github.com/bcgov/business-schemas/pull/223/changes

benjamin-bc-gov:
this PR will be closed , the changes required and test are in business schema.

PR #4648, opened 2026-08-01
API ledger client document url updates

doug-lovett:
*Issue #:* /bcgov/entity#34436

*Description of changes:*
Update the filing ledger download URL set up for client documents after switching from Minio to DRS.
- Use the DRS file key to build the download URL if available.
- Add missing document types to the configuration used to build the download URL.
- Review/update DRS documents types in existing configuration used to build the download URL.

### DIFF
--- a/legal-api/src/legal_api/reports/document_service.py
+++ b/legal-api/src/legal_api/reports/document_service.py
@@ -50,18 +50,34 @@ FILING_DOCUMENTS = {
     }, {
         "documentType": "CORP_AFFIDAVIT",
         "documentClass": "CORP"
+    }],
+    "continuationIn": [{
+        "documentType": "CNTA",
+        "documentClass": "CORP"
+    }],
+    "continuationOut": [{
+        "documentType": "CNTO",
+        "documentClass": "CORP"
     }]
 }
+# Map to DRS document based on the default static document name set when the client does not submit a filename.
+# Included for migrated minio docs where submissions with no filename were allowed.
 STATIC_DOCUMENTS = {
     "Unlimited Liability Corporation Information": {
         "documentType": "DIRECTOR_AFFIDAVIT",
         "documentClass": "CORP"
+    },
+    "Court Order": {
+        "documentType": "CRTO",
+        "documentClass": "*"
     }
 }
 APP_JSON = "application/json"
 APP_PDF = "application/pdf"
 DOC_PATH = "/documents"
 BEARER = "Bearer "
+DS_ID_PREFIX = "DS"
+DS_KEY_TOKENS = 2
 
 
 class ReportTypes(BaseEnum):
@@ -500,10 +516,10 @@ class DocumentService:
                 return response2
         return report_response
 
-
     def _update_static_document(self, doc: dict, doc_list: list) -> list:
         """
         Update static document urls in the document_list if a DRS match is found.
+        For static documents try obtaining download URL params from the file key first (DRS key).
 
         docs: The DRS document information to match on.
         doc_list: The business list of reports/documents for the filing.
@@ -516,12 +532,7 @@ class DocumentService:
         for key in doc_list:
             if key == "staticDocuments":
                 for static_doc in doc_list.get("staticDocuments"):
-                    name: str = static_doc.get("name")
-                    if (name and name == doc.get("name")) or (
-                            name and STATIC_DOCUMENTS.get(name) and
-                            doc.get("documentClass") == STATIC_DOCUMENTS[name].get("documentClass") and
-                            doc.get("documentType") == STATIC_DOCUMENTS[name].get("documentType")
-                        ):
+                    if self.is_static_doc_match(doc, static_doc):
                         static_doc["url"] = static_doc["url"] + query_params
                         break
             elif any(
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/reports/document_service.py
```diff
@@ -533,6 +544,60 @@ def _update_static_document(self, doc: dict, doc_list: list) -> list:
...
+        """
+        if str(static_doc.get("url")).find("documentClass=") > 0:
+            return False
+        drs_id = self.get_url_drs_id(static_doc.get("url", ""))
+        if drs_id:
+            return drs_id == doc.get("identifier")
+        name: str = static_doc.get("name")
+        doc_name: str = doc.get("name", "")
+        if (doc_name):
+            if name.removesuffix(".pdf") == doc_name.removesuffix(".pdf"):
```
kialj876, line 582: this will bail if 'name' is None. Add 'name' to the if?

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -533,6 +544,60 @@ def _update_static_document(self, doc: dict, doc_list: list) -> list:
...
+                return True
+            if (
+                STATIC_DOCUMENTS.get(name) == name and
+                doc.get("documentType") == STATIC_DOCUMENTS[name].get("documentType")
+            ):
+                return True
+        # Try partial matching static doc name with default name.
+        for key, value in STATIC_DOCUMENTS.items():
+            if (
+                name.startswith(str(key)) and
```
kialj876, line 592: this also looks like it will bail when 'name' is None

legal-api/tests/unit/reports/test_document_service.py
```diff
@@ -309,7 +325,25 @@
...
+   (True, STATIC_URL2, "Test court order", DRS_TEST_DOC, "CORP", "CRTO", "DS0000101951", ""),
+   (True, STATIC_URL1, "Test court order", DRS_TEST_DOC, "CORP", "CRTO", "DS0000101951", "Test court order.pdf"),
+   (True, STATIC_URL1, "Test court order", DRS_TEST_DOC, "FIRM", "CRTO", "DS0000101951", "Test court order"),
+   (True, STATIC_URL1, "Test court order", DRS_TEST_DOC, "COOP", "CRTO", "DS0000101951", "Test court order.pdf"),
+   (True, STATIC_URL1, "Court Order 1234", DRS_TEST_DOC, "CORP", "CRTO", "DS0000101951", "Court Order 1234"),
+   (True, STATIC_URL1, "Court Order 1234", DRS_TEST_DOC, "CORP", "CRTO", "DS0000101951", ""),
+   (False, STATIC_URL3, "Test court order", DRS_TEST_DOC, "CORP", "CRTO", "DS0000101951", "Test court order.pdf"),
+   (False, STATIC_URL2, "Test court order", DRS_TEST_DOC, "CORP", "CRTO", "DS0000101952", ""),
+   (False, STATIC_URL1, "Court Order 1234", DRS_TEST_DOC, "CORP", "COSD", "DS0000101951", ""),
+   (False, STATIC_URL1, "Court Order 1234", DRS_TEST_DOC, "CORP", "COSD", "DS0000101951", "Court Order.pdf"),
```
kialj876, line 344: can you add a case here for the 'drs_filename' as None? Is 'None' possible for migrated minio documents without a name or would they evaluate to empty strings?

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -533,6 +544,60 @@ def _update_static_document(self, doc: dict, doc_list: list) -> list:
...
+        drs_id = self.get_url_drs_id(static_doc.get("url", ""))
+        if drs_id:
+            return drs_id == doc.get("identifier")
+        name: str = static_doc.get("name")
+        doc_name: str = doc.get("name", "")
+        if (doc_name):
+            if name.removesuffix(".pdf") == doc_name.removesuffix(".pdf"):
+                return True
+            if (
+                STATIC_DOCUMENTS.get(name) == name and
```
kialj876: I think this is a dict to str comparison here? It will always be False

legal-api/src/legal_api/reports/document_service.py
```diff
@@ -516,12 +532,7 @@ def _update_static_document(self, doc: dict, doc_list: list) -> list:
         for key in doc_list:
             if key == "staticDocuments":
                 for static_doc in doc_list.get("staticDocuments"):
-                    name: str = static_doc.get("name")
-                    if (name and name == doc.get("name")) or (
-                            name and STATIC_DOCUMENTS.get(name) and
-                            doc.get("documentClass") == STATIC_DOCUMENTS[name].get("documentClass") and
```
kialj876, line 522: We're no longer checking against this at all inside 'is_static_doc_match' is that on purpose? I saw you introduce a wild card for this for court order which means this would need to change a bit if it was kept in. Or if we're not going to use 'documentClass' in STATIC_DOCUMENTS for anything can it be removed from the dict?

kialj876 (commented):
Looking pretty good, but I think there's a few minor issues below

legal-api/src/legal_api/reports/document_service.py
doug-lovett: Class removed as an unnecessary check.

legal-api/src/legal_api/reports/document_service.py
doug-lovett: Check added above even though the name is the ledger download name and a defect somewhere else if it does not exist.

legal-api/src/legal_api/reports/document_service.py
doug-lovett: See my previous comment.

legal-api/tests/unit/reports/test_document_service.py
doug-lovett: Added: the DRS API returns an empty string instead of None if no file name exists.

legal-api/src/legal_api/reports/document_service.py
doug-lovett: Thanks for catching, fixed.

meawong (approved):

vysakh-menon-aot (approved):

kialj876 (approved):

PR #4592, opened 2026-07-20
34299-special-resolution-drs-reporttype

mruff-aeq:
*Issue #:* /bcgov/entity#34299

*Description of changes:*
Fix coop dissolution Special Resolution serving the Dissolution Application due to a shared DRS FILING report type.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #4689, opened 2026-08-10
34457 - Distinct DRS report types for coop annual report bundled outputs

mruff-aeq:
Issue #bcgov/entity#34457 

Description:
a coop AR's bundled changeOfDirectors/changeOfAddress outputs shared the FILING DRS report type with the AR's own report, so every output served whichever document was stored first; they now use FILING-2/FILING-4 (same pattern as https://github.com/bcgov/lear/pull/4592).


doug-lovett (approved):
Increment api version?

PR #4703, opened 2026-08-17
32831 nr duplicate

benjamin-bc-gov:
*Issue #:* /bcgov/entity#32831

*Description of changes:*

 Problem: At submission time, the API validated that an NR was consumable (via NameX) but didn't check if the same NR was already referenced in another filing in PENDING or PAID status. This allowed two filings using the same NR to be  
  submitted simultaneously, potentially creating two businesses with the same name.                                                                                                                                                          
                                                                                                                                                                                                                                             
  Fix:                                                                                                                                                                                                                                       
  - Added FILING_TYPES_WITH_NR and NR_BLOCKING_STATUSES constants to common_validations.py
  
  - Added _nr_in_pending_filing(nr_number, exclude_filing_id) helper that queries the DB via JSONB path for any filing in PENDING, PAID, AWAITING_REVIEW, CHANGE_REQUESTED, or APPROVED status referencing the same NR    
                     
  - Called the check inside validate_name_request() — returns error "Name Request is already part of a pending filing."     
                                                                                                                   
  - Added 5 unit tests covering: NR in PENDING blocks, NR in PAID blocks, NR in DRAFT passes, NR not in any filing passes, own filing excluded from check

### DIFF
--- a/legal-api/tests/unit/services/filings/validations/test_common_validations.py
+++ b/legal-api/tests/unit/services/filings/validations/test_common_validations.py
@@ -2438,13 +2438,15 @@ def test_nr_in_draft_filing_does_not_block(session):
 
 
 def test_nr_excludes_own_filing(session):
-    """Assert a filing's own NR does not block itself."""
+    """Assert a filing in PENDING status cannot be re-submitted (PUT is blocked by is_allowed),
+    so a filing cannot block itself — this test confirms PENDING status is correctly blocked."""
     filing_json = {
         'filing': {
             'header': {'name': 'incorporationApplication'},
             'incorporationApplication': {'nameRequest': {'nrNumber': 'NR 0000005'}}
         }
     }
-    filing = factory_pending_filing(None, filing_json)
+    factory_pending_filing(None, filing_json)
 
-    assert _nr_in_pending_filing('NR 0000005', exclude_filing_id=filing.id) is False
+    # A second filing using the same NR should be blocked
+    assert _nr_in_pending_filing('NR 0000005') is True
--- a/legal-api/tests/unit/services/filings/validations/test_common_validations.py
+++ b/legal-api/tests/unit/services/filings/validations/test_common_validations.py
@@ -19,7 +19,7 @@ from unittest.mock import patch
 
 from legal_api.core.filing import Filing as CoreFiling
 from legal_api.errors import Error
-from business_model.models import Business, Party, PartyRole, ShareClass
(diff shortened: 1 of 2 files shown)


legal-api/tests/unit/services/filings/validations/test_common_validations.py
```diff
@@ -2375,3 +2388,65 @@ def test_get_file_data_drs_failure(session, monkeypatch):
...
+        }
+    }
+    filing = factory_pending_filing(None, filing_json)
+    filing.payment_completion_date = filing.filing_date
+    filing.save()
+
+    assert _nr_in_pending_filing('NR 0000003') is True
+
+
+def test_nr_in_draft_filing_does_not_block(session):
```
kialj876: can you please condense 'test_nr_in_pending_filing_blocks', 'test_nr_in_paid_filing_blocks', and 'test_nr_in_draft_filing_does_not_block' into one test with 'filing_type', 'state', and 'expected' params and add in all the valid filing types

legal-api/tests/unit/services/filings/validations/test_common_validations.py
```diff
@@ -2375,3 +2388,65 @@ def test_get_file_data_drs_failure(session, monkeypatch):
...
+            'header': {'name': 'incorporationApplication'},
+            'incorporationApplication': {'nameRequest': {'nrNumber': 'NR 0000004'}}
+        }
+    }
+    filing.save()
+
+    assert _nr_in_pending_filing('NR 0000004') is False
+
+
+def test_nr_excludes_own_filing(session):
```
kialj876: you can remove this as it isn't testing what is described here. It is just returning True because there is a pending filing with the NR which is already tested above

legal-api/tests/unit/services/filings/validations/test_common_validations.py
kialj876: And can you also please add a case where its an invalid nr filing type and a case or separate test where its a valid nr filing type without nr information

kialj876 (commented):
Looking good, just a couple minor comments on the tests

legal-api/tests/unit/services/filings/validations/test_common_validations.py
benjamin-bc-gov: on it

kialj876 (approved):

meawong (approved):
nice

PR #4713, opened 2026-08-20
34460 error downloading

benjamin-bc-gov:
*Issue #:* /bcgov/entity#34460

*Description of changes:*

updated how pay API is called in receipt download

### DIFF
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
@@ -119,7 +119,7 @@ def get_documents(identifier: str, # noqa: PLR0911, PLR0912
 
         if legal_filing_name:
             if legal_filing_name.lower().startswith("receipt"):
-                return _get_receipt(business, filing, jwt.get_token_auth_header())
+                return _get_receipt(business, filing)
 
             return get_pdf(filing.storage, legal_filing_name)
         elif file_key and (document := Document.find_by_file_key(file_key)):
@@ -207,7 +207,7 @@ def _get_document_list(business: Business, filing: Filing):
     return jsonify(document_list), HTTPStatus.OK
 
 
-def _get_receipt(business: Business, filing: Filing, token):
+def _get_receipt(business: Business, filing: Filing):
     """Get the receipt for the filing."""
     if filing.status not in (
             Filing.Status.COMPLETED,
--- a/legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
+++ b/legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
@@ -26,10 +26,10 @@ from flask_cors import cross_origin
 from flask_pydantic import validate as pydantic_validate
 from pydantic import BaseModel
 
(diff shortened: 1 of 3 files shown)


legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
kialj876: 'token' is unused now I think? Please remove it if so

legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
```diff
@@ -223,8 +224,8 @@ def _get_receipt(business: Business, filing: Filing, token):
         filing.filing_type == "noticeOfWithdrawal"
     ):
         effective_date = LegislationDatetime.format_as_report_string(filing.storage.effective_date)
-
-    headers = {"Authorization": "Bearer " + token}
+    service_token = AccountService.get_bearer_token()
+    headers = {"Authorization": "Bearer " + service_token}
```
kialj876, line 228: Did you validate that this works? 
I had a quick look at pay-api and they do not skip permission checks on system service accounts for this call. It looks like the users this was bailing on need a transaction viewing role on their token. I'm not sure if system accounts will get it as its a user transaction.

legal-api/tests/unit/resources/v2/test_business_filings/test_filing_documents.py
```diff
@@ -2193,6 +2198,46 @@ def test_temp_document_list_for_now(app, mocker, session, client, jwt, monkeypat
     assert rv_data == expected
 
 
+def test_receipt_download_uses_service_token(session, client, jwt, mocker):
```
kialj876: I don't think we need another test here, you can just validate the token used in the 'test_get_receipt'  and 'test_get_receipt_request_mock' test

legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
kialj876: This may be a question for the business because a user who is not allowed to view transactions on their account probably shouldn't be allowed to see the receipt in entities. Pay api may also be incorrectly evaluating the user permission to view transactions (they changed how this worked ~5 months ago and they are depending on a token role which may not be getting added as their code expects)

kialj876 (commented):
Couple comments for you below

legal-api/src/legal_api/resources/v2/business/business_filings/business_documents.py
benjamin-bc-gov: do you have someone I can reach out to ?

kialj876: Okay after our standup discussion I think this is probably okay. We can come back to it in a different ticket if there's a change needed

benjamin-bc-gov: okay , I will address other comments .

Thanks .

kialj876 (approved):

PR #4718, opened 2026-08-21
34088 legal api colin snap proxy and validation

kialj876:
*Issue #:* /bcgov/entity#34088

*Description of changes:*
- allows colin businesses into amalgs
- colin snapshot proxy


legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
```diff
@@ -78,19 +86,13 @@ def validate(amalgamation_json: dict, account_id) -> Error | None:
...
-            msg.extend(err)
-
-        err = validate_parties_delivery_address(amalgamation_json, legal_type, filing_type)
-        if err:
-            msg.extend(err)
+        msg.extend(structural_errors)
+    else:
+        # short-form: the filing must mirror the primary/holding business's current data
+        msg.extend(validate_primary_or_holding_match(amalgamation_json, filing_type, amalgamation_type))
+        msg.extend(_augment_with_source_update_hint(structural_errors, amalgamation_type))
```
kialj876, line 95: since this is now validated here as both matching existing data and structurally sound, we can trust the amalgamation filing json in the filer / outputs / colin sync (it'll work the same way for horizontal/vertical as it does for regular)

legal-api/src/legal_api/services/colin.py
```diff
@@ -13,25 +13,80 @@
...
+    return response is not None and response.status_code < HTTPStatus.INTERNAL_SERVER_ERROR
 
 
 class ColinService:
     """Provides services to use the colin-api."""
 
     @staticmethod
-    def query_business(identifier: str):
+    @cache.cached(make_cache_key=get_colin_cache_key, unless=skip_colin_cache, response_filter=is_cacheable_response)
+    def call_colin_api(path: str, token: str | None = None, use_cache = True) -> Response | None:
```
mruff-aeq: `response.json()` is called unguarded in a few places downstream of this (`business.py` line 161, `amalgamation_application.py` lines 498 and 588, `continuation_in.py` line 265). If COLIN or the gateway ever returns a non-JSON body with a 200/502, `JSONDecodeError` escapes as a 500. A small helper treated the same as a fetch failure would cover all four:

```python
def _safe_json(response) -> dict | None:
    try:
        return response.json()
    except ValueError:
        return None
```

legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
```diff
@@ -344,65 +417,285 @@ def _check_aml_permission_or_default_error(msg: list, message: str, default_erro
...
-            msg.append({
-                "error": f"{identifier} has a draft, pending or future effective filing.",
-                "path": amalgamating_business_path
-            })
-        elif Business.is_pending_amalgamating_business(identifier):
-            msg.append({
-                "error": f"{identifier} is part of a future effective amalgamation filing.",
-                "path": amalgamating_business_path
-            })
+    if business_info["state"] == Business.State.HISTORICAL.name:
```
mruff-aeq: COLIN dict passed straight through here; LEAR path always sets these keys, COLIN path not guaranteed. `business_info["state"]`, `["hasFutureEffectiveFiling"]`, `["adminFreeze"]`, `["goodStanding"]` → missing key = 500 mid-validation. Swap all four to `.get()`

legal-api/src/legal_api/services/colin.py
kialj876: okay I added this in to get used in the call function response and updated the integrations with the ColinService to consume the dict and the status as a tuple instead of the direct response

legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
kialj876: sure thing - updated. The current response does ensure these but you're right we probably shouldn't rely on it even though its our own internal api

mruff-aeq (approved):

legal-api/src/legal_api/services/filings/validations/amalgamation_application.py
vysakh-menon-aot: does ui provide these details already? if not we need to update existing amalgamation filing_json in prod

kialj876: yes it already does. I verified filing new ones in dev and I also verified the existing json filings for non regular in prod (not including placeholders ported over from colin which don't have any data)

meawong (approved):

vysakh-menon-aot (approved):

PR #4719, opened 2026-08-21
34553-firms-completing-party-client-skip

mruff-aeq:
*Issue #:* /bcgov/entity34553

*Description of changes:*
Adds validation on firm registrations filed by clients that the completing party name matches the submitting user's name from their login (Keycloak).

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/registration.py
+++ b/legal-api/src/legal_api/services/filings/validations/registration.py
@@ -18,15 +18,17 @@ from typing import Final
 
 import pycountry
 from dateutil.relativedelta import relativedelta
+from flask import g, has_request_context
 from flask.globals import request_ctx
 from flask_babel import _ as babel
 
 from business_common.utils.legislation_datetime import LegislationDatetime
-from business_model.models import Business, PartyRole
+from business_model.models import Business, PartyRole, User
 from legal_api.core.filing import Filing
 from legal_api.errors import Error
 from legal_api.services import STAFF_ROLE, NaicsService, flags
 from legal_api.services.filings.validations.common_validations import (
+    is_same_str,
     validate_court_order,
     validate_name_request,
     validate_offices_addresses,
@@ -73,6 +75,7 @@ def validate(registration_json: dict) -> Error | None:
     msg.extend(validate_naics(registration_json))
     msg.extend(validate_business_type(registration_json, legal_type))
     msg.extend(validate_party(registration_json, legal_type))
+    msg.extend(validate_completing_party_name(registration_json))
     msg.extend(validate_parties_addresses(registration_json, filing_type))
     msg.extend(validate_start_date(registration_json))
     msg.extend(validate_offices(registration_json))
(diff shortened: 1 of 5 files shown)


legal-api/src/legal_api/services/filings/validations/registration.py
```diff
@@ -163,6 +166,52 @@ def validate_party(filing: dict, legal_type: str, filing_type="registration") ->
...
+    for party in filing["filing"][filing_type].get("parties", []):
+        roles = party.get("roles", [])
+        if any(role.get("roleType", "").lower().replace(" ", "_") == PartyRole.RoleTypes.COMPLETING_PARTY.value
+               for role in roles):
+            officer = party.get("officer", {})
+            break
+    if not officer or officer.get("partyType") != "person":
+        return msg
+
+    if not (user := User.find_by_jwt_token(g.jwt_oidc_token_info)):
```
kialj876: Pretty sure you need to use get_or_create_user_by_jwt. Someone who is filing for the first time won't be in there yet. I think they are added in after validation passes

legal-api/src/legal_api/services/filings/validations/registration.py
```diff
@@ -163,6 +166,52 @@ def validate_party(filing: dict, legal_type: str, filing_type="registration") ->
...
+        return msg
+
+    user_name = " ".join(" ".join(
+        filter(None, [user.firstname, user.middlename, user.lastname])
+    ).split())
+    if not user_name:
+        return msg  # user record has no name to verify against
```
kialj876: We should return an error here I think?

legal-api/src/legal_api/services/filings/validations/registration.py
```diff
@@ -73,6 +75,7 @@ def validate(registration_json: dict) -> Error | None:
     msg.extend(validate_naics(registration_json))
     msg.extend(validate_business_type(registration_json, legal_type))
     msg.extend(validate_party(registration_json, legal_type))
+    msg.extend(validate_completing_party_name(registration_json))
```
kialj876: I think this needs to be added to change_of_registration too? 
Also did you consider adding it into validate_party? Might be nice having it there since it already checks for completing party

kialj876 (commented):
Looking good, just a couple minor comments

legal-api/src/legal_api/services/filings/validations/registration.py
mruff-aeq: Good catch, switched to get_or_create_user_by_jwt so first-time filers are covered.

legal-api/src/legal_api/services/filings/validations/registration.py
mruff-aeq: Agreed, a nameless user record now falls through to the mismatch error instead of skipping.

legal-api/src/legal_api/services/filings/validations/registration.py
mruff-aeq: Moved the check into validate_party, which picks up changeOfRegistration for free via its existing call.

legal-api/src/legal_api/services/filings/validations/registration.py
kialj876, line 137: nit: might be nice to move it here and just pass in the party so that we don't need to duplicate the logic for finding the completing party in the json again

kialj876 (approved):

PR #4724, opened 2026-08-24
34086 Filer - amalgamation colin business flow

kialj876:
*Issue #:* /bcgov/entity#34086

*Description of changes:*
- updates for colin businesses in amalgamation flow
- short forms now rely on data in the filing json (corresponding legal-api changes were already made to make this possible)
- ~in Draft because it requires poetry lock business model update which is still in PR: https://github.com/bcgov/lear/pull/4723~ done


queue_services/business-filer/src/business_filer/filing_processors/amalgamation_application.py
```diff
@@ -46,6 +45,14 @@ def create_amalgamating_businesses(amalgamation_filing: dict, amalgamation: Amal
         elif business := Business.find_by_identifier(identifier):
             amalgamating_business.business_id = business.id
             dissolve_amalgamating_business(business, filing_rec)
+        elif identifier:
+            # a business in COLIN that is not loaded in LEAR - validated against the COLIN
+            # snapshot at submission; the colin sync write-back moves the corp to HAM
+            amalgamating_business.colin_identifier = identifier
```
vysakh-menon-aot, line 51: Can you create a new ticket (if we don't have one) to move expro business into this field rather than doing it in foreign_identifier

vysakh-menon-aot (approved):

queue_services/business-filer/src/business_filer/filing_processors/amalgamation_application.py
kialj876: I was thinking it made sense to keep it as foreign due to the business rules? (i.e. staff only, can't be the holding/primary, etc.). It is a case where its both foreign and in colin, but I was thinking it fits more into the foreign bucket

vysakh-menon-aot: That was not the plan. The reason to keep expro in foreign is its in colin. Once its migrated to lear we need to remap to the right business id. Since the current work is track colin identifier in a separate filed it would be better to keep here. This will make remapping easier at the time of migration

Also verified in PROD: We dont have any amalgamation with expro

mruff-aeq (approved):

queue_services/business-filer/src/business_filer/filing_processors/amalgamation_application.py
kialj876: Okay I will dm you on teams and we can iron this out. Setting expro's to this has a larger blast radius because they are a foreign company (business rules wise) so it effects the logic at each point

PR #4726, opened 2026-08-24
34485 court orders validation in correction

vysakh-menon-aot:
*Issue #:* /bcgov/entity#34485

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/common_validations.py
+++ b/legal-api/src/legal_api/services/filings/validations/common_validations.py
@@ -533,48 +533,41 @@ def validate_share_currency(filing_json, filing_type, business=None):
     return msg
 
 
-def validate_court_order(court_order_path, court_order):
+def validate_court_order(
+    court_order_path: str,
+    court_order: dict,
+    is_file_or_details_required: bool = False
+) -> list[dict]:
     """Validate the courtOrder data of the filing."""
     msg = []
 
-    # TODO remove it when the issue with schema validation is fixed
-    min_file_number_length: Final = 5
-    max_file_number_length: Final = 20
-    if "fileNumber" not in court_order:
-        err_path = court_order_path + "/fileNumber"
-        msg.append({"error": "Court order file number is required.", "path": err_path})
-    elif (
-        len(court_order["fileNumber"]) < min_file_number_length or
-        len(court_order["fileNumber"]) > max_file_number_length
-    ):
-        err_path = court_order_path + "/fileNumber"
-        msg.append({"error": "Length of court order file number must be from 5 to 20 characters.",
-                    "path": err_path})
-
-    if (effect_of_order := court_order.get("effectOfOrder", None)) and effect_of_order != "planOfArrangement":
+    if (effect_of_order := court_order.get("effectOfOrder")) and effect_of_order != "planOfArrangement":
         msg.append({"error": "Invalid effectOfOrder.", "path": f"{court_order_path}/effectOfOrder"})
 
-    court_order_date_path = court_order_path + "/orderDate"
-    if "orderDate" in court_order:
-        try:
-            court_order_date = dt.fromisoformat(court_order["orderDate"])
-            if court_order_date.timestamp() > datetime.now(UTC).timestamp():
-                err_path = court_order_date_path
-                msg.append({"error": "Court order date cannot be in the future.", "path": err_path})
-        except ValueError:
-            err_path = court_order_date_path
-            msg.append({"error": "Invalid court order date format.", "path": err_path})
-            
+    if order_date := court_order.get("orderDate"):
+        court_order_date = dt.fromisoformat(order_date)
+        if court_order_date.timestamp() > datetime.now(UTC).timestamp():
+            msg.append({"error": "Court order date cannot be in the future.", "path": f"{court_order_path}/orderDate"})
+
+    if is_file_or_details_required:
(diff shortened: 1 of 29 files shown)


legal-api/src/legal_api/services/filings/validations/common_validations.py
```diff
@@ -533,48 +533,41 @@ def validate_share_currency(filing_json, filing_type, business=None):
...
-    if "fileNumber" not in court_order:
-        err_path = court_order_path + "/fileNumber"
-        msg.append({"error": "Court order file number is required.", "path": err_path})
-    elif (
-        len(court_order["fileNumber"]) < min_file_number_length or
-        len(court_order["fileNumber"]) > max_file_number_length
-    ):
-        err_path = court_order_path + "/fileNumber"
-        msg.append({"error": "Length of court order file number must be from 5 to 20 characters.",
-                    "path": err_path})
```
vysakh-menon-aot, line 552: schema validates this. updated unit test to verify

legal-api/src/legal_api/services/filings/validations/correction.py
```diff
@@ -170,9 +170,53 @@ def _validate_corps_correction(business: Business, filing_dict, legal_type, msg)
...
+    if filing_dict.get("filing", {}).get("correction", {}).get("courtOrder", None):
+        msg.extend(court_order_validation(filing_dict))
+
     msg.extend(_validate_continuation_in_correction(filing_dict, filing_type, legal_type))
     msg.extend(_validate_out_correction(filing_dict, filing_type))
     msg.extend(_validate_amalgamation_correction(filing_dict, filing_type, business))
+    msg.extend(_validate_court_orders_correction(filing_dict, business))
+
+
+def _validate_court_orders_correction(filing_dict, business: Business):
```
kialj876, line 182: Can you add a doc string for this one?
Am I correct in summarizing it like this:
- gets all existing court orders under the business
- ensures any update to an existing court order matches one of these and that it was part of the corrected filing
- ensures only a maximum of one new court order is added, including any new ones added as part of the corrected filing
  - does this mean if the corrected filing added a court order, then the correction cannot add a new one?
- ensures regular court order validation

legal-api/src/legal_api/services/filings/validations/correction.py
```diff
@@ -170,9 +170,53 @@ def _validate_corps_correction(business: Business, filing_dict, legal_type, msg)
...
+                is_file_or_details_required = (court_order_db["filingType"] == "courtOrder")
+        elif order.get("filingId") != filing_dict["filing"]["correction"]["correctedFilingId"]:
+            msg.append({"error": _("Filing Id does not match corrected filing Id."), "path": path})
+        else:
+            if (
+                (
+                    court_order_db := next(
+                        (o for o in court_orders_db if o["filingType"] == corrected_filing_type), None
+                    )
+                ) and next((o for o in orders if o.get("id") == court_order_db["id"]), None)
```
kialj876: Is this checking if the corrected filing already added a new court order?

I think if there were two filings of the same type under the business you may get the wrong evaluation here because you are using 'next' and comparing that order id, but that may be an unrelated filing with the same filing type (i.e. different correction on a different filing or a different alteration or something) and so the court order you select with 'next' would pass on the 'next((o for o in orders if o.get("id") == court_order_db["id"]), None)' even there was a new court order on the original filing

legal-api/src/legal_api/services/filings/validations/correction.py
vysakh-menon-aot: If the correction is filed based on a court order then courtOrder prop has those details (line 173)

legal-api/src/legal_api/services/filings/validations/correction.py
vysakh-menon-aot: Fixed

kialj876 (commented):
Looks good to me but you've got some failing tests to look at

legal-api/src/legal_api/services/filings/validations/correction.py
```diff
@@ -240,17 +287,17 @@ def _validate_special_resolution_correction(filing_dict, legal_type, msg):
...
         msg.extend(validate_signatory_name(filing_dict, filing_type))
-    if filing_dict.get("filing", {}).get(filing_type, {}).get("correction", {}).get("courtOrder", None):
+    if filing_dict.get("filing", {}).get(filing_type, {}).get("courtOrder", None):
         msg.extend(court_order_validation(filing_dict))
-    if filing_dict.get("filing", {}).get(filing_type, {}).get("correction", {}).get("rulesFileKey", None):
+    if filing_dict.get("filing", {}).get(filing_type, {}).get("rulesFileKey", None):
         msg.extend(rules_change_validation(filing_dict))
-    if filing_dict.get("filing", {}).get(filing_type, {}).get("correction", {}).get("memorandumFileKey", None):
+    if filing_dict.get("filing", {}).get(filing_type, {}).get("memorandumFileKey", None):
         msg.extend(memorandum_change_validation(filing_dict))
```
vysakh-menon-aot, line 301: Noticed this wrong validation in COOP correction. Assuming this fix has no impact in UI

kialj876: I think you are right but maybe call this out to your QA so they can do a quick verify

kialj876 (approved):

PR #4695, opened 2026-08-13
34541_noa-ccc-bca-italic

mruff-aeq:
*Issue #:* /bcgov/entity#34541

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #4729, opened 2026-08-25
34541 - Italicize Business Corporations Act in Notice of Articles CCC statement

mruff-aeq:
*Issue:*bcgov/entity#34541

Description of changes:
Apply the same italics fix from #4695 to the Notice of Articles CCC statement (common/statement.html), which renders the same text but was missed.


mruff-aeq:
image
business corporations act now italics

meawong (approved):

PR #4732, opened 2026-08-25
continuation in 2 step nr check

benjamin-bc-gov:
*Issue #:* /bcgov/entity#32831

*Description of changes:*

Bug:
When a Continuation In filing goes through its 2-step approval process, the NR duplicate check introduced blocks step 2. Because the same filing that used the NR in step 1 is being flagged as a Duplicate when staff tries to approve it in step 2. ( Double NR Check on same filing) .

Fix:

Pass the current filing's ID into the NR duplicate check so it can skip itself — allowing the same filing to proceed from step 1 to step 2 with the same NR it originally submitted.

### DIFF
--- a/legal-api/src/legal_api/services/filings/validations/common_validations.py
+++ b/legal-api/src/legal_api/services/filings/validations/common_validations.py
@@ -129,11 +129,10 @@ NR_BLOCKING_STATUSES = [
 def _nr_in_pending_filing(nr_number: str, exclude_filing_id: int | None = None) -> bool:
     """Return True if the NR is already referenced in a non-draft/non-completed filing.
 
-    exclude_filing_id: the ID of the filing currently being validated. Some filings (e.g.
-    continuationIn) go through a 2-step staff review — the filing sits in AWAITING_REVIEW
-    while staff approves it. Without this exclusion, the 2-step approval would find the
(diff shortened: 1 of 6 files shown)


legal-api/src/legal_api/services/filings/validations/common_validations.py
```diff
@@ -126,13 +126,23 @@
...
-    """Return True if the NR is already referenced in a non-draft/non-completed filing."""
+def _nr_in_pending_filing(nr_number: str, exclude_filing_id: int | None = None) -> bool:
+    """Return True if the NR is already referenced in a non-draft/non-completed filing.
+
+    exclude_filing_id: the ID of the filing currently being validated. Some filings (e.g.
+    continuationIn) go through a 2-step staff review — the filing sits in AWAITING_REVIEW
+    while staff approves it. Without this exclusion, the 2-step approval would find the
+    same filing in AWAITING_REVIEW and incorrectly block itself with a duplicate-NR error.
+    Excluding the current filing ensures only *other* filings are checked.
+    """
```
meawong, line 136: The filing sits in `APPROVED` at the point when a user is submitting the continuation application filing in step 2

benjamin-bc-gov: updated

meawong (approved):

PR #4778, opened 2026-09-11
34494 court order document list in correction fix

vysakh-menon-aot:
*Issue #:* /bcgov/entity#34494

*Description of changes:*

### DIFF
--- a/legal-api/src/legal_api/core/filing.py
+++ b/legal-api/src/legal_api/core/filing.py
@@ -588,6 +588,7 @@ class Filing:  # pylint: disable=too-many-public-methods
             filing.storage.transaction_id and
             (business_rev := VersionedBusinessDetailsService.get_business_revision_obj(filing.storage, business.id))
         ):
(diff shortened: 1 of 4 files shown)


legal-api/src/legal_api/core/filing.py
```diff
@@ -588,6 +588,7 @@ def _populate_completed_filing_documents(  # noqa: PLR0913
             filing.storage.transaction_id and
             (business_rev := VersionedBusinessDetailsService.get_business_revision_obj(filing.storage, business.id))
         ):
+            business_rev.identifier = business.identifier
```
vysakh-menon-aot, line 591: version does not have work with hybrid prop https://github.com/bcgov/lear/blob/main/python/common/business-registry-model/src/business_model/models/business.py#L322

meawong (approved):
LGTM might be worth a quick test for the "existing `fileKey` gets skipped" behavior, will leave it up to you

PR #4774, opened 2026-09-10
34510 update notice of withdrawal email template

mruff-aeq:
*Issue:* bcgov/entity#34510

Description of changes:
Move the notice of withdrawal email to the markdown template kit and fold it into the general filing notification processor, showing the filing number instead of incorporation and business numbers when the withdrawn record is a new business application , same as the AGM location change fold in 4772.

Note: `meta/filing.py` and `services/versioned_business_details.py` no longer have callers after this change; their removal is left for a separate ticket.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

PR #4781, opened 2026-09-14
34907 emailer - remove unused meta filing and versioned business details modules

mruff-aeq:
*Issue:*bcgov/entity#34907

Description of changes:
Removes the business-emailer `meta/filing.py` and `services/versioned_business_details.py` modules, their tests, and the related ruff per-file ignore, since nothing imports them after the markdown template migrations (#4772, #4774).


meawong (approved):
Nice clean-up

PR #4791, opened 2026-09-17
34788 - updated legal api to support tramways

JazzarKarim:
*Issue #:* /bcgov/entity#34788

*Description of changes:*
Updates temporary business number checks in legal-api so real tramway businesses (TMY + 7 digits, for example TMY0000008) are not treated as temp bootstrap IDs.

Temp registrations also start with T, and some random temps can even start with TMY (although super low probability), so a plain startswith("T") / startswith("TMY") check is too broad.

The endpoints now treat an identifier as temporary only when it starts with T and does not match `^TMY\d{7}$`

Example:
image


meawong (approved):

kialj876 (approved):
Worth having a test too I think

PR #4796, opened 2026-09-18
34926 Legal API, Colin API, Colin Updater - colin sync updates

kialj876:
*Issue #:* /bcgov/entity#34926

*Description of changes:*
- see comment in ticket for details


colin-api/src/colin_api/models/address.py
```diff
@@ -168,6 +168,14 @@ def create_new_address(cls, cursor, address_info: dict = None, corp_num: str = N
                 province = province_state_name
                 province_state_name = ''
 
+            # ADDR_LINE_1/2 are VARCHAR2(50) in COLIN; LEAR allows longer values
+            addr_line_1 = (address_info.get('streetAddress') or '').upper()
+            addr_line_2 = (address_info.get('streetAddressAdditional') or '').upper()
+            if len(addr_line_1) > 50 or len(addr_line_2) > 50:
+                current_app.logger.warning(f'Truncating address line(s) over 50 chars for {corp_num}')
+                addr_line_1 = addr_line_1[:50]
+                addr_line_2 = addr_line_2[:50]
```
kialj876, line 177: This allows old filings through in dev/test. Current filings through LEAR are validated for this and there are no prod cases. If we were to get a prod case like this I think its better to allow the sync through rather than have the business fall completely out of sync, but open to changing this and returning the 400

gcp-jobs/update-colin-filings/devops/vaults.gcp.env
```diff
@@ -8,3 +8,4 @@ COLIN_API_URL="op://API/$APP_ENV/colin-api-entity/COLIN_API_URL"
 COLIN_API_VERSION="op://API/$APP_ENV/colin-api-entity/COLIN_API_VERSION"
 VPC_CONNECTOR="op://CD/$APP_ENV/base/VPC_CONNECTOR"
 JOB_TOTAL_LIMIT="op://entity/$APP_ENV/filings-jobs/JOB_TOTAL_LIMIT"
+SKIPPED_IDENTIFIERS="op://entity/$APP_ENV/filings-jobs/SKIPPED_IDENTIFIERS"
```
kialj876, line 11: I updated dev/test 1password for this with all the current identifiers failing on unique identifier errors and identifier not found errors (caused by prod imports)

legal-api/src/legal_api/resources/v2/business/colin_sync.py
```diff
@@ -120,12 +127,27 @@ def _get_initial_filing_json(filing: Filing, business: Business):
...
+        if not filing_json["filing"]["business"].get("legalType"):
+            filing_json["filing"]["business"]["legalType"] = business.legal_type
+
     return filing_json
 
 
+def _has_colin_relevant_correction(filing: Filing, filing_json: dict) -> bool:
+    """Return whether the correction changes anything the colin-api can create a filing for."""
+    colin_correction_flags = ("nameChanged", "nameTranslationsChanged", "officeChanged",
+                              "partyChanged", "resolutionChanged", "shareChanged", "commentOnly")
```
vysakh-menon-aot, line 144: how does it sync the subsequent filings? let say a director name is changed in correction and a subsequent change of director filing is updating that director then colin-api error out saying could not find director

kialj876: It would only get skipped if there was nothing in it to update Colin with (according to our rules). A director name change would have the "partyChanged" flag so it would be sent for the sync. 
Otherwise it'll err forever on this when there's no changes to make in Colin and never sync subsequent filings

vysakh-menon-aot (approved):

mruff-aeq (approved):

